PR #2, opened 2021-06-17
Initial commit with actions

bobstrecansky:
Creating a very basic initial repo with necessary github action things.


SergeyKanzhelev (approved):
Approving so actions can be tested as part of repro set up. Haven't actually reviewed them deeply

PR #8, opened 2021-07-22
Add ECS Detector and Unit tests

ohamuy:
This PR adds the ECS detector:

The ECS detector looks for one of the two environment variables and populates and returns a resource with the container ID and hostName if available/succeeds. The ECS detector is split up into two classes since mock testing cannot be achieved otherwise.

A PR in the core repo will be created to add the CONTAINER_ID as a resource constant as an ID is something that exists quite often. A comment is in the code left as a TODO.

Currently the PHP Library has no Logging so comments were left in as TODOs for future development if logging is added.

Change to the directories in the phpunit.xml.dist file have been done to reflect the current repository structure

Change to GH Actions to remedy incorrectly cached files. Adds the fix to properly find newly added classes

@alolita @anuraaga @bobstrecansky

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -43,6 +43,9 @@ jobs:
       if: steps.composer-cache.outputs.cache-hit != 'true'
       run: composer install --prefer-dist --no-progress --no-suggest
 
+    - name: Update Composer
+      run: composer update
+
     - name: Check Style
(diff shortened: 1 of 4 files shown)


ohamuy:
The reason the other Psalm checks are failing are because the core dependency is not added in this PR. The PR #7 should be merged before this one.

bobstrecansky (approved):

alolita (approved):

PR #9, opened 2021-07-22
Add EKS Detector and Unit Tests

ohamuy:
This PR adds the EKS detector:

The EKS detector looks at certification files and creates requests to verify the application is running on EKS. If it is running on EKS it will look for the containerId and cluster name. Otherwise it will return an empty resource.

Currently the PHP Library has no Logging so comments were left in as TODOs for future development if logging is added.

@alolita @bobstrecansky @anuraaga

### DIFF
--- a/detectors/aws/EksDetector.php
+++ b/detectors/aws/EksDetector.php
@@ -1,0 +1,200 @@
+<?php
+
+declare(strict_types=1);
+/*
+ * Copyright The OpenTelemetry Authors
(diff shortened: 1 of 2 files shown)


ohamuy:
The reason the Psalm checks are failing are because the core dependency is not added in this PR. The PR #7 should be merged before this one.

bobstrecansky (approved):

PR #15, opened 2021-08-10
Add two AWS Sample Apps

ohamuy:
In this PR:
- Added examples folder to autoloader in `composer.json`
- Moved 2 detectors into the correct folder
- Renamed Propagator to AwsXrayPropagator.php
- Added a sample app that uses OTLPgrpc exporter and creates a child span to export traces
- Added a more complex app that creates a request to two microservices and then exports the trace

@alolita @bobstrecansky


bobstrecansky (approved):
LGTM, thanks @ohamuy !

PR #22, opened 2021-10-29
Symfony Bundle (Part1 SdkBundle)

tidal:
**Description**

This PR adds a Symfony bundle to the `opentelemetry-php-contrib` package as discussed in https://github.com/open-telemetry/opentelemetry-php/issues/220.
The bundle (SdkBundle) addresses configuration of the SDK and populating services accordingly in Symfony DI Container.
The bundle as for now covers resource and trace configuration with metrics being added in a future PR.
(Autoinstrumentation of Symfony projects will be addresses in a future InstrumentationBundle, which will use the SdkBundle)

**********

**Type of change**

New (sub) package

**********

**Testing**

As far as possible all PHP code is covered by unit tests. 
However there are three important exceptions. 
- The base configuration (sdk.php) the bundle ships with (which is written in PHP)
- The Configuration class 
- The OtelSdkExtension class

All three classes are marked to be ignored by code coverage.
The config file (sdk.php) cannot be unit tested and the Configuration and OtelSdkExtension class are almost impossible to be tested, or only with a hard to maintain "Mock Fest", since they mainly operate on and configure other objects.
The Configuration class is responsible for validating and normalizing the user provided configuration of the bundle, but it does so by operating with and returning a "TreeNodeBuilder" object. The OtelSdkExtension class is responsible for configuring the DI container according to the user provided configuration.

However the expected behavior of all above classes/files is extensively covered by integration tests.

The integration test can be run by calling `vendor/bin/phpunit tests/integration`, but they have not (yet) been added as a 
as a test suite in the `phpunit.xml.dist`. At the moment they would be run by CI in conjunction with the Unit tests, which would result in misleading code coverage reports. They should be run as a separate task in the future.

**********

**Documentation**
- A first documentation has been added in the directory of the bundle, while there is room for improvement.
- A basic documentation has been added to the README of the repository.

**********

**Further Information**

The Configuration and OtelSdkExtension classes are rather big in size and not really easy to follow. This is a general problem with these classes in Symfony Bundles, which I don't really like. I will factor out as much code as possible in a future PR, to make them easier for other developers to get into and maintain. However non of the code in this bundle will ever be called in a production environment, as the resulting output is cached by Symfony as executable PHP code.

I will create new issues for todos and  improvements regarding the bundle as separate issues in the contrib respository.


bobstrecansky:
There was a style error with this PR; you can vet this locally with `make style`.

Otherwise, I very much like the direction this PR is going in.

tidal:
@bobstrecansky 
Ok, The CS issues should be fixed, don't know why PHPStorm did not show them before.

Btw, might be idea to use [GrumPHP](https://github.com/phpro/grumphp), so the CS can be automatically fixed before a commit?

bobstrecansky:
We might be able to use GrumPHP!  I'm not opposed to it.

tidal:
Ok, should be hopefully all good now. The last CS error was introduced by me by making psalm happy.

One question, I forgot to ask though. You may have seen, that I mentioned "alternative" runtimes for PHP in the readme. Is this something the PHP library aims to support? (They are finding more and more adoption, as they are usually way faster than  the classic webserver setup.) If so, I would be happy to create some tests for them (There will possibly be some issues with global states, etc.)

bobstrecansky:
We aren't planning on supporting alternative PHP runtimes at this point, so I wouldn't put a lot of stock in trying to support them.

bobstrecansky:
I also forgot to mention - we can most certainly add the unit tests to the CI in a separate Github action.  Happy to do that whenever!

bobstrecansky:
One more small style lint I believe!

tidal:
> One more small style lint I believe!

Oh my bad, forgot to commit the fixes in the test directory.

> we can most certainly add the unit tests to the CI in a separate Github action.

The unit tests should be already running in CI , no? But I guess the integration tests would need to be added.

tidal:
Hmm, psalm seems to have a lot of problems with the tests (which it doesn't report locally for me neither with PHP7.4 nor 8.0). Especially Mocks seem to be a problem, since they are basically union types. I will try to fix this, but in the worst case, I will have to tell psalm to ignore these tests completely.

tidal:
@bobstrecansky
PHPStan should good now, too.

I found out the issue with my local psalm runs. I will  fix the psalm errors more granulary in another PR, when I don't have to bother you with running the CI.

bobstrecansky (approved):

PR #30, opened 2021-11-10
Reorganize directory structure

tidal:
This PR does the following:
- Reorganizes the directory structure to be PSR4 compatible
- Organizes contrib types in dedicated directories, so they can be published as separate packages in the future
- (applies some more granular fixes for phpstan issues in the SdkBundle tests)


bobstrecansky (approved):
Than you for this reorganization!  Much improved.

PR #34, opened 2021-12-19
Fix compatibility with main package

tidal:
Fixes compatibility with main package.
- Fixes AWS code + tests
- Adjusts composer.json + README
- Adds usage of new php version container images
- Adds a few tests to make code-coverage happy
- Activates Integration Tests to be run in CI

(Does not fix AWS examples)


tests/Unit/Aws/Eks/DetectorTest.php
```diff
@@ -86,25 +80,24 @@ public function TestValidClusterNameInvalidContainerId()
 
         $detector = new Detector($mockData, $client);
 
+        echo PHP_EOL . '-------------------------';
```
brettmc, line 83: ? output from tests

brettmc (approved):
one minor question, otherwise lgtm

tests/Unit/Aws/Eks/DetectorTest.php
tidal: Oh, good catch!

PR #43, opened 2022-02-10
Dont forcefully set service_name option on any exporter

cedricziel:
Some exporters do not define this option (OTLP f.e.) and setting this option
breaks them.

### DIFF
--- a/src/Symfony/OtelSdkBundle/DependencyInjection/OtelSdkExtension.php
+++ b/src/Symfony/OtelSdkBundle/DependencyInjection/OtelSdkExtension.php
@@ -318,7 +318,7 @@ class OtelSdkExtension extends Extension implements LoggerAwareInterface
         if (isset($config[Conf::URL_NODE])) {
             $options[Conf::URL_NODE] = $config[Conf::URL_NODE];
         }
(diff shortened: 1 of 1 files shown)


tidal:
@cedricziel 
Thanks for your contribution and sorry again for the late response.
As I mentioned on slack, we are busy setting up the infrastructure to have better 2e2 test at the moment.

There is one integration test failing due to your change.
I know the integration tests are not an easy read at the moment, so here is how you can fix the test.
I believe removing the `'service_name' => 'foo',` entry should do the trick.
https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/tests/Integration/Symfony/OtelSdkBundle/DependencyInjection/OtelSdkExtensionTest.php#L338

The other way would be to add a `serviceName` to the MockExporter -> https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/tests/Integration/Symfony/OtelSdkBundle/Mock/SpanExporter.php
However I'm not really sure atm, if that would break any other tests.

tidal (approved):

tidal:
Merging this, since the tests should be fixed with: https://github.com/open-telemetry/opentelemetry-php-contrib/pull/47

PR #46, opened 2022-02-20
Fix AWS &%$§$, Fix SDKBundle config, Update SDK dependency

tidal:
fixes: https://github.com/open-telemetry/opentelemetry-php/issues/564

Uses temporarily `dev-main` as the SDK dependency, to not have to fix the AWS stuff again once 
a new SDK release is published. This should be reverted to the appropriate version once the new release is published.

### DIFF
--- a/src/Symfony/OtelSdkBundle/Resources/config/sdk.php
+++ b/src/Symfony/OtelSdkBundle/Resources/config/sdk.php
@@ -4,7 +4,10 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Symfony\OtelSdkBundle\Resources;
 
+use OpenTelemetry\SDK\AttributeLimits;
+use OpenTelemetry\SDK\Attributes;
 use OpenTelemetry\SDK\Resource;
+use OpenTelemetry\SDK\SystemClock;
 use OpenTelemetry\SDK\Trace;
 use OpenTelemetry\SDK\Trace\Sampler;
 use OpenTelemetry\SDK\Trace\SpanProcessor;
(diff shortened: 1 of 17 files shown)


src/Symfony/OtelSdkBundle/Resources/config/sdk.php
```diff
@@ -35,24 +38,24 @@
 
     // UTIL
 
-    $helper->setService(Trace\SystemClock::class);
+    $helper->setService(SystemClock::class);
```
brettmc, line 41: Maybe a side-issue for another time, but `setService` without a key implies that there is one, and repeated calls would replace it. `addService` instead?

brettmc (approved):

src/Symfony/OtelSdkBundle/Resources/config/sdk.php
tidal: The key is generated from the FQCN, so while you could argue, that the class is indeed added, it cannot only be added once. It's just configuration. But feel free to change it, Brett. :)
```php
/**
     * @param string $class
     * @psalm-param class-string $class
     * @param bool $alias
     * @return ServiceConfigurator
     */
    public function setService(string $class, bool $alias = true): ServiceConfigurator
    {
        $id = ServiceHelper::classToId($class);
        $serviceConfigurator = $this->getConfigurator()->set($id, $class);
        if ($alias === true) {
            $this->setAlias($class, $id);
        }

        return $serviceConfigurator;
    }

```

tidal:
Thanks @brettmc

PR #50, opened 2022-03-09
Use PSR Http Client instead of Guzzle

mtk3d:
My proposition to achieve the issue https://github.com/open-telemetry/opentelemetry-php/issues/565

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,8 @@
         "php": "^7.4 || ^8.0",
         "ext-json": "*",
         "open-telemetry/opentelemetry": "~0.0.6",
-        "php-http/message": "^1.12",
-        "php-http/discovery": "^1.14"
+        "php-http/discovery": "^1.14",
+        "php-http/message": "^1.12"
     },
     "replace": {
         "open-telemetry/otel-sdk-bundle": "self.version"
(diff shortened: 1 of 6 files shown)


brettmc (approved):

brettmc:
Thankyou, this looks good. Could you please sign the CLA so that your contribution can be accepted?
Can you please check the failing action for 8.0?

mtk3d:
Sure, it's signed now.
I will fix the failing action tomorrow.
I already investigated that there is a problem with Psalm on 8.0 which fails on `\UnitEnum` type in the `getParameter` method on the newest Symfony DependencyInjection component version.

mtk3d:
I've tried to find the solution for this issue, but unfortunately, that is the psalm issue or more precisely the Symfony Plugin issue.
I found this issue https://github.com/vimeo/psalm/issues/7738 and there are no fixes in Psalm or Symfony Plugin yet, so at this moment the only way to fix the issue is to suppress it. I've specified the smallest part to ignore as possible so this is suppressed only in one file, only for a single class.

tidal:
Thank you @mtk3d !

It's looking all fine now.

tidal (approved):

PR #52, opened 2022-03-14
WIP: Fix first AWS example app

mtk3d:
This is an old example app 1 rewritten to the new opentelemetry-php API.
The example is working but I'm not sure about some parts of the code if they are correct.


examples/aws/AwsClientApp/src/main.php
```diff
@@ -0,0 +1,135 @@
...
+
+    printTraceId($span);
+
+    $span->end();
+}
+
+if ($line === 'aws-sdk-call') {
+    $span = $tracer->spanBuilder('session.generate.aws.sdk.span')->setSpanKind(SpanKind::KIND_CLIENT)->startSpan();
+
+    $propagator->inject($carrier);
```
mtk3d, line 91: I'm wondering if there should be some integration between `S3Client` and `$carrier` because it looks like this variable is unuseful in this example.
I need to learn more about AWS XRay.

bobstrecansky (approved):

PR #54, opened 2022-03-17
Specify AWS attributes to show exact services in XRay

mtk3d:
I managed to get the real AWS services instead of the generic segments in XRay
Zrzut ekranu 2022-03-17 o 15 35 57


bobstrecansky:
The codecov github action got stuck, but that's not really relevant here.  Merging.

PR #58, opened 2022-05-02
[WIP] Add OtelInstrumentationBundle

cedricziel:
**This is work in progress**

## What
This introduces a instrumentation-bundle, so we enable users to capture signals from Symfony apps out of the box.

While this PR is a draft, I'll continously force-push. The work is currently based on the OtelSDKBundle, but I found the approach kinda cumbersome, so i'll probably opt to using a more Symfony-esque approach, leveraging components instead of custom code.

## Scope

* enable easy setup
* enable capturing http interactions
* provide a few instrumentations by means of decorating known interfaces


bobstrecansky:
@cedricziel - do you have everything you need to continue on this draft?

cedricziel:
@bobstrecansky I have everything except time ;-)

bobstrecansky:
I know the feeling!  Will be ready to view when needed.

anthonyptetlow:
Hi, Is active work still going on with this bundle? 
I was hoping to try and use this on my Symfony project but when trying to install I couldn't get it to resolve to a set of packages:
image
(tried installing by cloning the repo locally and defining a relative path repository for the Symfony in composer then composer requiring it)

brettmc:
Hi @anthonyptetlow I would say that it's not being actively worked on, and is likely suffering some bit-rot.
In terms of getting it installed, it looks like you'll need to modify composer.json and relax the version constraints.

wadjei:
apologies all, I plan to refine this but have other commitments for most of this week

I kind of kludged it together to make it work for my specific case but there's a lot more that needs to be done before it would work generically (bear in mind, I'm only really familiar with my specific case though...)

I also haven't tried this with the recent commits to main so the issues I was having may well be resolved, but `composer.json` in symfony is still pinned to ^0.0.15 so I suspect not.

The main issue, as I came to understand it, was that the extension's configuration is no longer compatible with the library's Exporter configurations which now require a `TransportInterface` so I manipulated the incoming configuration options to make it so. I think a proper fix would involve fixing or redefining the bundle's configuration so that it supports the different Transports that are available.

cedricziel:
Hi folks - sorry I'm going to close this one. I just dont find the time for it.

PR #63, opened 2022-05-24
Add make target for package split

tidal:


Makefile
```diff
@@ -25,4 +26,8 @@ bash:
 	$(DC_RUN_PHP) bash
 style:
 	$(DC_RUN_PHP) env XDEBUG_MODE=off vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.php --using-cache=no -vvv
+script:
+	ls $(SCRIPT_DIR)
```
bobstrecansky, line 30: I'm not sure what the purpose of this is?  I didn't see `SCRIPT_DIR` defined anywhere else either.

bobstrecansky (approved):
One question about `SCRIPT_DIR` but not a blocker.

Makefile
bobstrecansky: @tidal mentioned in our SIG meeting this was superfluous and he was going to remove it.

PR #67, opened 2022-06-20
Add Deptrac dependency check to CI run

tidal:
- Adds deptrac a dev dependency
- Adds deptrac dependency check to CI run
- Adds deptrac make target


tidal:
Since he package validation commands in [devtools](https://github.com/opentelemetry-php/dev-tools) rely on the deptrac file for resolving the packages, there needs to be a change in these commands as well to not break CI.

brettmc (approved):

PR #72, opened 2022-07-15
Instrumentation directory structure and AWS instrumentation class

carolabadeer:
Hello!

This PR contains the initial setup and barebones of the AWS SDK Instrumentation class as well as the directory structure for adding library instrumentations. The organization of the instrumentation directory is similar to the style used in the opentelemetry-php-contrib repo to aim for a consistent experience for our customers. 

These changes and others that will follow are related to implementing instrumentation for AWS SDK mentioned in [issue #688](https://github.com/open-telemetry/opentelemetry-php/issues/688).

PR #75, opened 2022-08-11
Support for AWS SDK PHP Instrumentation

carolabadeer:
Hi! 

This PR builds on the work completed in #72 and contains the implementation that allows for instrumenting the AWS SDK in PHP. The [runkit7 library](https://www.php.net/manual/en/book.runkit7.php) is used to patch methods from the [AWS Client class](https://github.com/aws/aws-sdk-php/blob/master/src/AwsClient.php) and create spans. 

Depends on #797 in opentelemetry-php to get access to `TracerProvider`.


tests/Unit/Instrumentation/AwsSdk/AwsSdkInstrumentationTest.php
```diff
@@ -37,6 +43,68 @@ public function testInstrumentationInit()
...
+
+        $this->assertSame(
+            $awsSdkInstrumentation->getPropagator(),
+            $propagator
+        );
+    }
+
+    public function testSetTracerProvider()
+    {
+        $spanProcessor = new SimpleSpanProcessor(new ConsoleSpanExporter());
```
brettmc, line 68: The start of this and the next couple of tests is duplicated, so you could move it into a `setUp()` method, and make the variables you need to refer to into class variables instead of local.

brettmc:
Looking good. Because this is dependent on https://github.com/open-telemetry/opentelemetry-php/pull/797 let's make sure we wait for that to be merged first (and a new release tagged). I think you should also bump the version constraint on the main otel package to >= 0.0.15 (the next version).

src/Instrumentation/AwsSdk/AwsSdkInstrumentation.php
```diff
@@ -37,8 +46,90 @@ public function init(): bool
...
+        $this->root->end();
+    }
+
     public function activate(): bool
     {
+        AwsGlobal::setInstrumentation($this);
+
+        try {
+            $this->root = $this->getTracer()->spanBuilder('AwsSDKInstrumentation')->setSpanKind(SpanKind::KIND_CLIENT)->startSpan();
+            $this->root->activate();
```
brettmc, line 85: very important that you store the `$scope` you get back from activating a span, and ensure you `::detach()` it when you end the root span. This goes for any span that you activate, and we recommend doing it in a try/finally block

tests/Unit/Instrumentation/AwsSdk/AwsSdkInstrumentationTest.php
```diff
@@ -4,11 +4,29 @@
...
 use PHPUnit\Framework\TestCase;
 
 class AwsSdkInstrumentationTest extends TestCase
 {
+    private TracerProvider $tracerProvider;
+    private AwsSdkInstrumentation $awsSdkInstrumentation;
+
+    protected function setUp(): void
+    {
+        $spanProcessor = new SimpleSpanProcessor(new ConsoleSpanExporter());
```
brettmc, line 23: phpunit will probably be unhappy if this exporter ever actually exported anything (tests shouldn't create output).
Because this is a unit test, and your code should depend on interfaces, you can use mock objects: `$this->tracerProvider = $this->createMock(TracerProvider::class)`. The only "real" class that you should use is the class(es) you're testing: AwsSdkInstrumentation

tests/Unit/Instrumentation/AwsSdk/AwsSdkInstrumentationTest.php
```diff
@@ -37,6 +55,55 @@ public function testInstrumentationInit()
         );
     }
 
+    public function testSetXrayPropagator()
```
brettmc, line 58: this doesn't test what it says - it should probably add a propagator to the instrumentation class, get it back, and `$this->assertSame()` the input and output

brettmc, line 58: this test doesn't do anything useful. both set and get are covered by the next test, though.

tests/Unit/Instrumentation/AwsSdk/AwsSdkInstrumentationTest.php
```diff
@@ -37,6 +55,55 @@ public function testInstrumentationInit()
...
+            $this->awsSdkInstrumentation->getPropagator(),
+            $propagator
+        );
+    }
+
+    public function testSetTracerProvider()
+    {
+        $this->awsSdkInstrumentation->setTracerProvider($this->tracerProvider);
+
+        $this->assertInstanceOf(
```
brettmc, line 81: useless assertion. you should be checking the provider you get back from instrumentation

tests/Unit/Instrumentation/AwsSdk/AwsSdkInstrumentationTest.php
```diff
@@ -37,6 +55,55 @@ public function testInstrumentationInit()
...
+            TracerProvider::class,
+            $this->tracerProvider
+        );
+    }
+
+    public function testGetTracerProvider()
```
brettmc, line 87: useless, see previous comment

tests/Unit/Instrumentation/AwsSdk/AwsSdkInstrumentationTest.php
```diff
@@ -37,6 +55,55 @@ public function testInstrumentationInit()
...
+            $this->awsSdkInstrumentation->getTracerProvider(),
+            $this->tracerProvider
+        );
+    }
+
+    public function testGetTracer()
+    {
+        $this->awsSdkInstrumentation->setTracerProvider($this->tracerProvider);
+
+        $this->assertInstanceOf(
```
brettmc, line 101: useless, see above

brettmc (approved):

brettmc:
there are a couple of PRs in the main repo (800 and 797) that need to be merged and in a tagged release prior to this, so I'll hold off merging until then.

src/Instrumentation/AwsSdk/AwsSdkInstrumentation.php
```diff
@@ -37,8 +44,81 @@ public function init(): bool
...
+                '   
+                
+                $tracer = \OpenTelemetry\Instrumentation\AwsSdk\AwsGlobal::getInstrumentation()->getTracer();
+                $carrier = [];
+                
+                $this->span = $tracer->spanBuilder($this->getApi()->getServiceName() . "." . $name)->setSpanKind(\OpenTelemetry\Instrumentation\AwsSdk\AwsSdkInstrumentation::SPAN_KIND)->startSpan();
+                $this->scope = $this->span->activate();
+                
+                $propagator = \OpenTelemetry\Instrumentation\AwsSdk\AwsGlobal::getInstrumentation()->getPropagator();
+                $propagator->inject($carrier);
```
Nevay, line 93: Should be injected into the request headers, currently does nothing (`$carrier` isn't used afterwards).

src/Instrumentation/AwsSdk/AwsSdkInstrumentation.php
```diff
@@ -37,8 +44,81 @@ public function init(): bool
...
+                
+                ',
+            );
+
+            runkit7_method_redefine(
+                'Aws\AwsClient',
+                'executeAsync',
+                '$command',
+                ' 
+                    $this->scope->detach();
```
Nevay, line 113: Should likely be moved to `__call`, after `$this->__call_copy($name, $args);`?
Doesn't detach scope for non-`...Async` methods and doesn't handle concurrent calls.

src/Instrumentation/AwsSdk/AwsSdkInstrumentation.php
```diff
@@ -37,8 +44,81 @@ public function init(): bool
...
+                $tracer = \OpenTelemetry\Instrumentation\AwsSdk\AwsGlobal::getInstrumentation()->getTracer();
+                $carrier = [];
+                
+                $this->span = $tracer->spanBuilder($this->getApi()->getServiceName() . "." . $name)->setSpanKind(\OpenTelemetry\Instrumentation\AwsSdk\AwsSdkInstrumentation::SPAN_KIND)->startSpan();
+                $this->scope = $this->span->activate();
+                
+                $propagator = \OpenTelemetry\Instrumentation\AwsSdk\AwsGlobal::getInstrumentation()->getPropagator();
+                $propagator->inject($carrier);
+                
+                $this->span->setAttributes([
```
Nevay, line 95: [Span Creation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation)
> adding attributes at span creation is preferred to calling `SetAttribute` later, as samplers can only consider information already present during span creation.

Nevay (changes requested):
I'm not familiar with the AWS SDK - what is the reason for using runkit ("function/method manipulation is recommended only for unit testing") instead of injecting a middleware?

src/Instrumentation/AwsSdk/AwsSdkInstrumentation.php
```diff
@@ -37,8 +55,85 @@ public function init(): bool
...
+            $middleware = Middleware::tap(function ($cmd, $req) {
+                $tracer = $this->getTracer();
+                $propagator = $this->getPropagator();
+
+                $carrier = [];
+
+                $this->span = $tracer->spanBuilder($this->clientName . '.' . $cmd->getName())->setSpanKind(AwsSdkInstrumentation::SPAN_KIND)->startSpan();
+                $this->scope = $this->span->activate();
+
+                $propagator->inject($carrier, null, $this->getContext());
```
brettmc, line 105: I'm not totally sure that the root context is what you want here. We generally use current context (which the propagator will use by default)

brettmc (approved):

PR #76, opened 2022-08-12
Update the TracingKernelSubscriber example

wadjei:
Since changes in https://github.com/open-telemetry/opentelemetry-php/pull/760,
the original Kernel Listener example no longer works...

Following those changes, the stored Tracer instance gets deleted before the
application terminates so mainSpan can't be ended and none of the collected
trace info gets sent.
By refactoring to use TracerProvider instead, mainSpan's Tracer remains valid
during KernelEvents::TERMINATE so the trace session can be sent successfully.

### DIFF
--- a/src/Symfony/OtelSdkBundle/README.md
+++ b/src/Symfony/OtelSdkBundle/README.md
@@ -199,7 +199,7 @@ So our Listener class could look like this:
 namespace App\EventSubscriber;
 
 use OpenTelemetry\API\Trace\SpanInterface;
-use OpenTelemetry\SDK\Trace\Tracer;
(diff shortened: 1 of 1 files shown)


Nevay:
This should be fixed in the SDK bundle instead. `TracerProvider::shutdown()` should be called in `Bundle::shutdown()` and on exit (default Symfony `HttpKernelRunner` doesn't seem to call `::shutdown()`).

Side note: There are some other problems with the example subscriber such as not detaching the created scope. I will see if I can create a PR over the weekend to fix the subscriber / add an actual implementation.

brettmc (approved):

PR #78, opened 2022-08-22
Add PSR18 http client auto instrumentation

Nevay:
Draft to showcase autoinstrumentation capabilities.


src/Instrumentation/Psr18/client_tracing.php
```diff
@@ -0,0 +1,102 @@
...
+                if ($request->hasHeader($header)) {
+                    $spanBuilder->setAttribute(sprintf('http.request.header.%s', strtr(strtolower($header), ['-' => '_'])), $request->getHeader($header));
+                }
+            }
+
+            $span = $spanBuilder->startSpan();
+            $context = $span->storeInContext($parentContext);
+            $propagator->inject($request, HeadersPropagator::Instance, $context);
+
+            $context->activate();
```
brettmc, line 66: I don't think activating the span will work with multiple async http requests, because they will end in an undefined order. IMO, safer to pass a reference to the span into the post method.

Nevay: Calls can only return out-of-order if they suspend execution which is handled by the context storage.

brettmc:
@Nevay can you please remove composer.json from the PR? (I tried via the UI but just created a different conflict). I'll create a new one in a follow-up PR.

Nevay:
@brettmc rebased / dropped composer.json changes.

brettmc (approved):

PR #81, opened 2022-10-19
Slim and psr-15 auto-instrumentation

brettmc:


src/Instrumentation/Slim/src/SlimInstrumentation.php
```diff
@@ -0,0 +1,98 @@
...
+                    ->setAttribute(TraceAttributes::HTTP_METHOD, $request->getMethod())
+                    ->setAttribute(TraceAttributes::HTTP_REQUEST_CONTENT_LENGTH, $request->getHeaderLine('Content-Length'))
+                    ->setAttribute(TraceAttributes::HTTP_SCHEME, $request->getUri()->getScheme())
+                    ->startSpan();
+                Context::storage()->attach($root->storeInContext(Context::getCurrent()));
+            },
+            static function (App $app, array $params, ?ResponseInterface $response, ?Throwable $exception) {
+                $scope = Context::storage()->scope();
+                $scope?->detach();
+                if (!$scope || $scope->context() === Context::getCurrent()) {
```
pdelewski, line 49: Why in this case `$scope->context() === Context::getCurrent()` we return?

brettmc: I actually do not know, I copied that part from the psr-18 draft PR!

src/Instrumentation/Slim/src/SlimInstrumentation.php
```diff
@@ -0,0 +1,98 @@
...
+    public static function register(): void
+    {
+        $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.slim');
+        // @var SpanInterface $root
+        $root = null; //todo probably better to store this in context
+        hook(
+            App::class,
+            'handle',
+            static function (App $app, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation, &$root) {
+                $request = $params[0]; // @var ServerRequestInterface $request
```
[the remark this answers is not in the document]
brettmc: It's enforced by the method: https://github.com/slimphp/Slim/blob/4.x/Slim/App.php#L197 - I don't see a way for an invalid argument to make it here?
Or, should we make it a standard to always be extra-cautious and not trust the types of any `$params` ?

src/Instrumentation/Slim/src/SlimInstrumentation.php
```diff
@@ -0,0 +1,98 @@
...
+
+class SlimInstrumentation
+{
+    public static function register(): void
+    {
+        $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.slim');
+        // @var SpanInterface $root
+        $root = null; //todo probably better to store this in context
+        hook(
+            App::class,
```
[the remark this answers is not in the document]
brettmc: There's one technical issue to resolve to allow that. The observer which updates the root span's name post routing needs access to the root span.
I'm not aware of a way to access the root span without holding a reference to it (context traversal?) to enable that, so perhaps I just remove that feature? IMO, the route (with placeholders) is a good way to aggregate requests, so I think it needs to be in the trace _somewhere_, and the root span seems like the most useful/practical place for it...

src/Instrumentation/Slim/src/SlimInstrumentation.php
brettmc: @Nevay if you can explain what this does, I can document it if we think it's good practice for future instrumentations...

src/Instrumentation/Slim/src/CallableFormatter.php
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * @see https://stackoverflow.com/a/68113840/2063413
+     */
+    public static function format(callable $callable): string
+    {
+        return match (true) {
+            is_string($callable) => $callable,
+            is_array($callable) && is_object($callable[0]) => get_class($callable[0]) . '->' . $callable[1],
+            is_array($callable) => $callable[0] . '::' . $callable[1],
+            $callable instanceof Closure => 'closure',
```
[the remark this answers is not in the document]
brettmc: are you sure? I wrote a unit test for that case - `::format($callable())` executes the callable and it's the result of the callable that `::format()` operates on.
```php
$closure = function(string $foo): string { return $foo; };
$this->assertInstanceOf(\Closure::class, $closure);
$this->assertSame('closure', CallableFormatter::format($closure('bar')));
```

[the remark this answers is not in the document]
brettmc: fixed and added tests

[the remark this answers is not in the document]
Nevay: I was referring to first class callables (same applies to closures that were created using `Closure::fromCallable()`).

src/Instrumentation/Slim/src/SlimInstrumentation.php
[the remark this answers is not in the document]
Nevay: Hooks should not trust `$params`:
- the pre-hook is called before typehints are checked
- subclasses could widen the typehint

src/Instrumentation/Slim/src/SlimInstrumentation.php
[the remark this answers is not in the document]
Nevay: The span should be stored as request attribute.

src/Instrumentation/Slim/src/SlimInstrumentation.php
Nevay: Not needed here. The PSR18 draft doesn't create a span if an invalid argument is provided and attaches an unmodified context instead*, this check prevents modifying the parent span.
(* Always creating a span and only skipping the usage of argument/request related values might be preferable).

src/Instrumentation/Psr15/README.md
```diff
@@ -0,0 +1,26 @@
...
+## Overview
+Auto-instrumentation hooks are registered via composer, and spans will automatically be created for each psr-15 middleware that is executed.
+
+To export spans, you will need to create and register a `TracerProvider` early in your application's lifecycle:
+
+```php
+<?php
+require_once 'vendor/autoload.php';
+
+$tracerProvider = <create tracer provider>;
```
pdelewski, line 17: Do we plan to simplify that in production ready release?. Important aspect of auto-instrumentation is easy of use. Using this approach user needs to know some details of opentelemetry, at least how to create trace provider and export spans, then he has to add opentelemetry dependencies to application. This might be good for experienced user that want to know what's going on under the hood, but 99% users IMHO will be not interested in such details.

brettmc: If all inputs are provided via environment variables, we could automatically create and register everything. Combining https://github.com/open-telemetry/opentelemetry-php/blob/main/examples/traces/features/configuration_from_environment.php with the composer autoload-files technique is one way I think we could provide zero-code instrumentation.

src/Instrumentation/Slim/src/SlimInstrumentation.php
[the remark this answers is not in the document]
brettmc: done. there is a dependency now between slim and psr-15, so I'll either submit a follow-up addition to composer.json (when psr-15 auto-instrumentation is installable from packagist), or I could loosen the dependency (it's only really dependent on a const which defines a key for the root span for getting/setting the request attribute)

src/Instrumentation/Psr15/README.md
pdelewski: That's what I was thinking about

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,104 @@
...
+            },
+            static function (RequestHandlerInterface $handler, array $params, ?ResponseInterface $response, ?Throwable $exception) {
+                $scope = Context::storage()->scope();
+                $scope?->detach();
+                if (!$scope) {
+                    return;
+                }
+                $span = Span::fromContext($scope->context());
+                if ($exception) {
+                    $span->recordException($exception);
```
Nevay, line 96: Should set `exception.escaped`, [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md#attributes):
> [exception.escaped] SHOULD be set to true if the exception event is recorded at a point where it is known that the exception is escaping the scope of the span.

src/Instrumentation/Slim/tests/Unit81/CallableFormatterTest.php
```diff
@@ -0,0 +1,26 @@
...
+     */
+    public function test_format_first_class_callable(): void
+    {
+        if (PHP_VERSION < 8.1) {
+            $this->markTestSkipped();
+        }
+        $function = function (): stdClass {
+            return new stdClass();
+        };
+        $this->assertSame('stdClass', CallableFormatter::format($function(...)));
```
Nevay, line 24:
```suggestion
        $function = 'var_dump';
        $this->assertSame(CallableFormatter::format($function), CallableFormatter::format($function(...)));
```

src/Instrumentation/Slim/src/SlimInstrumentation.php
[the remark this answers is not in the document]
brettmc: thanks for the improvement, using `SpanInterface::class` is much better.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,104 @@
...
+                    ->setAttribute('code.namespace', $class)
+                    ->setAttribute('code.filepath', $filename)
+                    ->setAttribute('code.lineno', $lineno)
+                    ->setAttribute(TraceAttributes::HTTP_URL, (string) $request?->getUri())
+                    ->setAttribute(TraceAttributes::HTTP_METHOD, $request?->getMethod())
+                    ->setAttribute(TraceAttributes::HTTP_REQUEST_CONTENT_LENGTH, $request?->getHeaderLine('Content-Length'))
+                    ->setAttribute(TraceAttributes::HTTP_SCHEME, $request?->getUri()?->getScheme())
+                    ->startSpan();
+
+                Context::storage()->attach($root->storeInContext(Context::getCurrent()));
```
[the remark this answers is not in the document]
pdelewski: I would prefer second option, it may produce a lot of spans, but will be closer to reality. Later during testing we can revisit this decision

src/Instrumentation/Psr15/README.md
Nevay: With https://github.com/open-telemetry/opentelemetry-php/pull/849 we could define lazy initializers in the SDK that are invoked iff the user does not configure the instances manually.
```php
Globals::registerInitializer(static function(Configurator $configurator): Configurator {
    $tracerProvider = (new TracerProviderFactory('should be fetched from resource instead'))->create();
    ShutdownHandler::register($tracerProvider->shutdown(...));

    return $configurator->withTracerProvider($tracerProvider);
});
```

src/Instrumentation/Slim/src/CallableFormatter.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * @see https://stackoverflow.com/a/68113840/2063413
+     * @phan-suppress PhanUndeclaredMethod
+     * @psalm-suppress UndefinedMethod
+     */
+    public static function format($callable): string
+    {
+        if ($callable instanceof Closure || (is_string($callable) && function_exists($callable))) {
+            return (new ReflectionFunction($callable))->getName();
```
Nevay, line 20: Has to handle `Clazz::method(...)` and `$object->method(...)` too.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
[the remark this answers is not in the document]
brettmc: I think I've understood this requirement. If a span is already provided as a request attribute, parent will be current context. Else, try to extract trace context from request.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,110 @@
...
+         */
+        hook(
+            RequestHandlerInterface::class,
+            'handle',
+            static function (RequestHandlerInterface $handler, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                $request = ($params[0] instanceof ServerRequestInterface) ? $params[0] : null;
+                $parent = Context::getCurrent();
+                if ($request instanceof ServerRequestInterface && !$request->getAttribute(SpanInterface::class)) {
+                    //extract from request headers
+                    $parent = (new TraceContextPropagator())->extract($request->getHeaders());
```
Nevay, line 73:
```suggestion
                    $parent = Instrumentation\Globals::propagator()->extract($request->getHeaders());
```

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,110 @@
...
+                    ->setAttribute('code.namespace', $class)
+                    ->setAttribute('code.filepath', $filename)
+                    ->setAttribute('code.lineno', $lineno)
+                    ->setAttribute(TraceAttributes::HTTP_URL, (string) $request?->getUri())
+                    ->setAttribute(TraceAttributes::HTTP_METHOD, $request?->getMethod())
+                    ->setAttribute(TraceAttributes::HTTP_REQUEST_CONTENT_LENGTH, $request?->getHeaderLine('Content-Length'))
+                    ->setAttribute(TraceAttributes::HTTP_SCHEME, $request?->getUri()?->getScheme())
+                    ->startSpan();
+
+                Context::storage()->attach($span->storeInContext(Context::getCurrent()));
```
Nevay, line 87:
```suggestion
                Context::storage()->attach($span->storeInContext($parent));
```
Otherwise we lose extracted baggage etc.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,110 @@
...
+            RequestHandlerInterface::class,
+            'handle',
+            static function (RequestHandlerInterface $handler, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                $request = ($params[0] instanceof ServerRequestInterface) ? $params[0] : null;
+                $parent = Context::getCurrent();
+                if ($request instanceof ServerRequestInterface && !$request->getAttribute(SpanInterface::class)) {
+                    //extract from request headers
+                    $parent = (new TraceContextPropagator())->extract($request->getHeaders());
+                }
+                $span = $instrumentation->tracer()->spanBuilder(sprintf('HTTP %s', $request?->getMethod() ?? 'unknown'))
```
Nevay, line 75: Has to set `SpanKind::KIND_SERVER`.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,110 @@
...
+                    return [$request];
+                }
+            },
+            static function (RequestHandlerInterface $handler, array $params, ?ResponseInterface $response, ?Throwable $exception) {
+                $scope = Context::storage()->scope();
+                $scope?->detach();
+                if (!$scope) {
+                    return;
+                }
+                $span = Span::fromContext($scope->context());
```
Nevay, line 100: Has to set `StatusCode::STATUS_ERROR` for status code 5xx, [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#status)
> For HTTP status codes in the 5xx range, as well as any other code the client failed to interpret, span status MUST be set to Error.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,110 @@
...
+                    //extract from request headers
+                    $parent = (new TraceContextPropagator())->extract($request->getHeaders());
+                }
+                $span = $instrumentation->tracer()->spanBuilder(sprintf('HTTP %s', $request?->getMethod() ?? 'unknown'))
+                    ->setParent($parent)
+                    ->setAttribute('code.function', $function)
+                    ->setAttribute('code.namespace', $class)
+                    ->setAttribute('code.filepath', $filename)
+                    ->setAttribute('code.lineno', $lineno)
+                    ->setAttribute(TraceAttributes::HTTP_URL, (string) $request?->getUri())
```
Nevay, line 81:
```suggestion
                    ->setAttribute(TraceAttributes::HTTP_URL, $request?->getUri()->__toString())
```
Prevents adding an empty attribute.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
brettmc: done

src/Instrumentation/Slim/src/CallableFormatter.php
brettmc: I _think_ this is implemented now for both static and non-static class methods

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,115 @@
...
+                    ->setAttribute('code.namespace', $class)
+                    ->setAttribute('code.filepath', $filename)
+                    ->setAttribute('code.lineno', $lineno)
+                    ->startSpan();
+
+                Context::storage()->attach($span->storeInContext(Context::getCurrent()));
+            },
+            static function (MiddlewareInterface $middleware, array $params, ?ResponseInterface $response, ?Throwable $exception) {
+                $scope = Context::storage()->scope();
+                $scope?->detach();
```
tcarrio, line 49: Instead of using a nullish coalescer here, you could move it after the `if (!$scope) {...}` block entirely, since if it is null it won't have any `detach` method to call and you're going to return it anyway.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,115 @@
...
+                Context::storage()->attach($span->storeInContext($parent));
+                if ($request instanceof ServerRequestInterface) {
+                    $request = $request->withAttribute(SpanInterface::class, $span);
+
+                    return [$request];
+                }
+            },
+            static function (RequestHandlerInterface $handler, array $params, ?ResponseInterface $response, ?Throwable $exception) {
+                $scope = Context::storage()->scope();
+                $scope?->detach();
```
tcarrio, line 98: Same as my other comment. The ordering could be switched here to avoid an unnecessary nullish coalescer.

> Instead of using a nullish coalescer here, you could move it after the if (!$scope) {...} block entirely, since if it is null it won't have any detach method to call and you're going to return it anyway.

src/Instrumentation/Psr15/README.md
```diff
@@ -0,0 +1,26 @@
+# OpenTelemetry PSR-15 auto-instrumentation
+
+## Requirements
+
+* OpenTelemetry extension
+* OpenTelemetry SDK and exporters (required to actually export traces)
+
+## Overview
+Auto-instrumentation hooks are registered via composer, and spans will automatically be created for each psr-15 middleware that is executed.
```
tcarrio, line 9: This is a really small thing, but `PSR` is typically capitalized in usage.

src/Instrumentation/Psr15/README.md
```diff
@@ -0,0 +1,26 @@
+# OpenTelemetry PSR-15 auto-instrumentation
```
tcarrio, line 1: *Just wanna say I'm super excited for this* :fire:

tcarrio (commented):
Mostly a vested citizen in the CNCF channel excited for the OTel PHP SDK. A couple things I noticed here that I thought I would comment on slightly_smiling_face

src/Instrumentation/Psr15/README.md
brettmc: Thanks very much for the review feedback. I agree with all and have addressed them.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,115 @@
...
+                    $parent = Globals::propagator()->extract($request->getHeaders());
+                }
+                $span = $instrumentation->tracer()->spanBuilder(sprintf('HTTP %s', $request?->getMethod() ?? 'unknown'))
+                    ->setParent($parent)
+                    ->setSpanKind(SpanKind::KIND_SERVER)
+                    ->setAttribute('code.function', $function)
+                    ->setAttribute('code.namespace', $class)
+                    ->setAttribute('code.filepath', $filename)
+                    ->setAttribute('code.lineno', $lineno)
+                    ->setAttribute(TraceAttributes::HTTP_URL, $request?->getUri()->__toString())
```
pdelewski, line 83: Does it make sense to set these attributes in case request is null?

src/Instrumentation/Slim/src/CallableFormatter.php
brettmc: Done. It isn't pretty, but output now matches suggestions above.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,125 @@
...
+                if (!$request instanceof ServerRequestInterface) {
+                    //`pre` hook runs before params are checked, so this is theoretically possible. activate a non-recording
+                    //span so that the post handler has something to close
+                    $root = new NonRecordingSpan();
+                    $root->activate();
+                    Context::storage()->attach($root->storeInContext(Context::getCurrent()));
+                    return;
+                }
+                $root = $request->getAttribute(SpanInterface::class);
+                $builder = $instrumentation->tracer()->spanBuilder($root ? $function : sprintf('HTTP %s', $request?->getMethod() ?? 'unknown'))
```
pdelewski, line 81: Why $class is not a part of span name `($class . '::' . $function)`?

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,129 @@
...
+                $request = ($params[0] instanceof ServerRequestInterface) ? $params[0] : null;
+                if (!$request instanceof ServerRequestInterface) {
+                    //`pre` hook runs before params are checked, so this is theoretically possible. activate a non-recording
+                    //span so that the post handler has something to close
+                    $root = new NonRecordingSpan();
+                    $root->activate();
+                    Context::storage()->attach($root->storeInContext(Context::getCurrent()));
+                    return;
+                }
+                $root = $request->getAttribute(SpanInterface::class);
```
Nevay, line 80: Missing `SpanContextInterface $context` argument; should use `Span::wrap()` instead.

Do we want to replace the current span with a non recording span? I think we should still create the span (has to update `return` too): 
```suggestion
                $root = $request
                    ? $request->getAttribute(SpanInterface::class)
                    : Span::getCurrent();
```

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,129 @@
...
+                if (!$scope) {
+                    return;
+                }
+                $scope->detach();
+                $span = Span::fromContext($scope->context());
+                if ($exception) {
+                    $span->recordException($exception, [TraceAttributes::EXCEPTION_ESCAPED => true]);
+                    $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
+                }
+                if ($response?->getStatusCode() >= 500) {
```
Nevay, line 121: Should capture some additional [http attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes) such as `http.status_code`.

src/Instrumentation/Psr15/README.md
brettmc: I think `Globals` already takes care of that - eg if you try to get a tracer provider, it will first check in context:
```php
    public static function tracerProvider(): TracerProviderInterface
    {
        return Context::getCurrent()->get(ContextKeys::tracerProvider()) ?? self::globals()->tracerProvider;
    }
```

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
brettmc: I think it does...otherwise it would be more difficult to identify where the span came from, if you were looking at a mostly-empty span in a trace.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,115 @@
...
+            RequestHandlerInterface::class,
+            'handle',
+            static function (RequestHandlerInterface $handler, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                $request = ($params[0] instanceof ServerRequestInterface) ? $params[0] : null;
+                $parent = Context::getCurrent();
+                if ($request instanceof ServerRequestInterface && !$request->getAttribute(SpanInterface::class)) {
+                    //extract from request headers
+                    $parent = Globals::propagator()->extract($request->getHeaders());
+                }
+                $span = $instrumentation->tracer()->spanBuilder(sprintf('HTTP %s', $request?->getMethod() ?? 'unknown'))
```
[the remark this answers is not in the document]
brettmc: I have made some updates to address this.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
brettmc: done, added a few more attributes.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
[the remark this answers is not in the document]
pdelewski: I was thinking about current implementation for some time and I don't think that relying on external library parameters (I'm taking about `request` and its usage to store span) is good thing. I believe that instrumentation should be independent and rely only on itself implementation, therefore why not use `Context` (I got the same effect with it). Below pre hook that uses additional `$rootspan` variable

`pre: static function (RequestHandlerInterface $handler, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation, $rootspan) {
                $request = ($params[0] instanceof ServerRequestInterface) ? $params[0] : null;
                $root = Context::getCurrent()->get($rootspan);
                $builder = $instrumentation->tracer()->spanBuilder(
                    $root
                    ? sprintf('%s::%s', $class, $function)
                    : sprintf('PSR HTTP %s %s', $request?->getMethod(), $request->getUri() ?? 'unknown')
                )
                    ->setSpanKind(SpanKind::KIND_SERVER)
                    ->setAttribute('code.function', $function)
                    ->setAttribute('code.namespace', $class)
                    ->setAttribute('code.filepath', $filename)
                    ->setAttribute('code.lineno', $lineno);
                if (!$root && $request) {
                    //create http root span
                    $parent = Globals::propagator()->extract($request->getHeaders());
                    
                    $parent = $parent->with($rootspan, true);
                    $span = $builder
                        ->setParent($parent)
                        ->setAttribute(TraceAttributes::HTTP_URL, $request->getUri()->__toString())
                        ->setAttribute(TraceAttributes::HTTP_METHOD, $request->getMethod())
                        ->setAttribute(TraceAttributes::HTTP_REQUEST_CONTENT_LENGTH, $request->getHeaderLine('Content-Length'))
                        ->setAttribute(TraceAttributes::HTTP_SCHEME, $request->getUri()->getScheme())
                        ->startSpan();
                } else {
                }
             
                Context::storage()->attach($span->storeInContext($parent));
                return [$request];
            },`

src/Instrumentation/Slim/src/SlimInstrumentation.php
```diff
@@ -0,0 +1,148 @@
...
+         * and type SpanInterface which represents the root span, having been previously set (eg
+         * by Psr15 auto-instrumentation)
+         *
+         * @psalm-suppress ArgumentTypeCoercion
+         */
+        hook(
+            RoutingMiddleware::class,
+            'performRouting',
+            pre: null,
+            post: static function (RoutingMiddleware $middleware, array $params, ServerRequestInterface $request, ?Throwable $exception) {
```
pdelewski, line 98: This hook is used only to update span name created by other hook. I think that hooks should be self contained and orthogonal. I would remove this hook and use `request->getUri()` in `handle` hook(s)

src/Instrumentation/Slim/src/SlimInstrumentation.php
```diff
@@ -0,0 +1,148 @@
...
+{
+    public static function register(): void
+    {
+        $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.slim');
+
+        /**
+         * Create a span representing the http transaction for Slim\App::handle
+         * @psalm-suppress ArgumentTypeCoercion
+         */
+        hook(
```
pdelewski, line 36: Actually, I don't understand need for this hook as we already hooking into `RequestHandlerInterface::handle`

src/Instrumentation/Slim/src/SlimInstrumentation.php
brettmc: The reason I've done it as <set> then later <update> is:
- the spec says that span names by default should not be the uri: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#name (too-high cardinality)
- slim doesn't have a better route name until after the routing middleware has run (which it might not do, if an error occurs earlier)

src/Instrumentation/Slim/src/SlimInstrumentation.php
brettmc: You're right, it's redundant now. I left it there because I wasn't sure which way I'd end up going, but if having Psr15 instrumentation work out when to create a root span in `RequestHandlerInterface::handle`, then this hook can be removed.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -0,0 +1,127 @@
...
+        );
+
+        /**
+         * Create a span to wrap RequestHandlerInterface::handle. The first execution is assumed to be the root span, which
+         * is stored as a request attribute which may be accessed by later hooks.
+         */
+        hook(
+            RequestHandlerInterface::class,
+            'handle',
+            pre: static function (RequestHandlerInterface $handler, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
```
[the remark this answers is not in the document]
brettmc: In your example, is $rootSpan something like a `ContextKey(SpanInterface::class)`? It seems like a good idea to me.

src/Instrumentation/Slim/src/SlimInstrumentation.php
pdelewski: Maybe, we can extract part of uri, it would simplify code.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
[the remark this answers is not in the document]
pdelewski: Actually, in my example I stored just `bool`, but I think it does not matter what will be there, main point is to get rid of using `request attribute`

pdelewski:
@brettmc I'm thinking about merging this PR as it is now. We can make improvements later with additional PRs

brettmc:
@pdelewski great idea - I haven't even looked at this PR in a week. I'll give it a quick tidy up and submit, then others can collaborate

bobstrecansky (approved):

pdelewski (approved):

PR #84, opened 2022-11-15
store root span info directly in context instead of request attribute

pdelewski:
Stop relying on library request attribute and use context instead to store information about root span

### DIFF
--- a/src/Instrumentation/Psr15/src/Psr15Instrumentation.php
+++ b/src/Instrumentation/Psr15/src/Psr15Instrumentation.php
@@ -23,10 +23,17 @@ use Throwable;
  */
 class Psr15Instrumentation
 {
-    public static $rootSpan;
+    private static $rootSpan;
+
+    public static function getRootSpan()
+    {
+        return Psr15Instrumentation::$rootSpan;
+    }
 
     public static function register(): void
     {
+        Psr15Instrumentation::$rootSpan ??= Context::createKey('rootSpan');
+
         $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.psr15');
         /**
          * Create a span for each psr-15 middleware that is executed.
@@ -68,7 +75,7 @@ class Psr15Instrumentation
             'handle',
             pre: static function (RequestHandlerInterface $handler, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
                 $request = ($params[0] instanceof ServerRequestInterface) ? $params[0] : null;
-                $root = Context::getCurrent()->get(Psr15Instrumentation::$rootSpan);
+                $root = Context::getCurrent()->get(Psr15Instrumentation::getRootSpan());
                 $builder = $instrumentation->tracer()->spanBuilder(
                     $root
                     ? sprintf('%s::%s', $class, $function)
@@ -90,7 +97,7 @@ class Psr15Instrumentation
                         ->setAttribute(TraceAttributes::HTTP_REQUEST_CONTENT_LENGTH, $request->getHeaderLine('Content-Length'))
                         ->setAttribute(TraceAttributes::HTTP_SCHEME, $request->getUri()->getScheme())
                         ->startSpan();
-                    $parent = $parent->with(Psr15Instrumentation::$rootSpan, $span);
+                    $parent = $parent->with(Psr15Instrumentation::getRootSpan(), $span);
                 } else {
                     $span = $builder->startSpan();
                 }
@@ -123,5 +130,3 @@ class Psr15Instrumentation
         );
     }
 }
-
-Psr15Instrumentation::$rootSpan ??= Context::createKey('rootSpan');
--- a/src/Instrumentation/Psr15/src/Psr15Instrumentation.php
+++ b/src/Instrumentation/Psr15/src/Psr15Instrumentation.php
@@ -10,7 +10,6 @@ use OpenTelemetry\API\Trace\Span;
 use OpenTelemetry\API\Trace\SpanKind;
 use OpenTelemetry\API\Trace\StatusCode;
 use OpenTelemetry\Context\Context;
-use OpenTelemetry\Context\ContextKey;
 use function OpenTelemetry\Instrumentation\hook;
 use OpenTelemetry\SemConv\TraceAttributes;
 use Psr\Http\Message\ResponseInterface;
@@ -24,12 +23,11 @@ use Throwable;
  */
 class Psr15Instrumentation
 {
-    public static ContextKey $rootSpan;
+    public static $rootSpan;
 
     public static function register(): void
     {
         $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.psr15');
-
         /**
          * Create a span for each psr-15 middleware that is executed.
          */
@@ -126,4 +124,4 @@ class Psr15Instrumentation
     }
 }
 
-Psr15Instrumentation::$rootSpan = Context::createKey('rootSpan');
+Psr15Instrumentation::$rootSpan ??= Context::createKey('rootSpan');
--- a/src/Instrumentation/Psr15/src/Psr15Instrumentation.php
+++ b/src/Instrumentation/Psr15/src/Psr15Instrumentation.php
@@ -7,10 +7,10 @@ namespace OpenTelemetry\Contrib\Instrumentation\Psr15;
 use OpenTelemetry\API\Common\Instrumentation\CachedInstrumentation;
 use OpenTelemetry\API\Common\Instrumentation\Globals;
 use OpenTelemetry\API\Trace\Span;
-use OpenTelemetry\API\Trace\SpanInterface;
 use OpenTelemetry\API\Trace\SpanKind;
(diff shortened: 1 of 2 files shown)


src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -24,6 +24,8 @@
  */
 class Psr15Instrumentation
 {
+    public static ContextKey $rootSpanExists;
```
brettmc, line 27: I think this should just be called `$rootSpan`

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -85,14 +85,14 @@ public static function register(): void
                 if (!$root && $request) {
                     //create http root span
                     $parent = Globals::propagator()->extract($request->getHeaders());
+                    $parent = $parent->with(Psr15Instrumentation::$rootSpanExists, true);
```
brettmc, line 88: this looks like it should be storing the root span itself, and doing it after the span is created?

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -68,9 +70,7 @@ public static function register(): void
             'handle',
             pre: static function (RequestHandlerInterface $handler, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
                 $request = ($params[0] instanceof ServerRequestInterface) ? $params[0] : null;
-                $root = $request
-                    ? $request->getAttribute(SpanInterface::class)
-                    : Span::getCurrent();
+                $root = Context::getCurrent()->get(Psr15Instrumentation::$rootSpanExists);
```
brettmc, line 73: this is currently getting a boolean, and storing it in a variable that looks like it should hold a span

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
pdelewski: ok, It was `$rootSpan` at the beginning, but I changed that due to fact that I was storing bool

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
pdelewski: ok

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
pdelewski: ok, changed that

brettmc (approved):

brettmc:
I will look into why AWS build is failing. It looks like there is at least a linting issue with psr15 though: `1) Psr15/src/Psr15Instrumentation.php (braces)`. Can you fix that (and run `make all` to verify), then I'll merge it.

brettmc:
This probably also breaks the Slim auto-instrumentation, did you want to look at that in this PR, or in a new one?

pdelewski:
> This probably also breaks the Slim auto-instrumentation, did you want to look at that in this PR, or in a new one?

You mean this coverage report?  (Seems that now all checks are passing)

brettmc:
> Don't know exactly what do you mean, this coverage report?

No, this: https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/Slim/src/SlimInstrumentation.php#L44 It is trying to get the root span from a request attribute, which psr-15 doesn't do now.

pdelewski:
> > It is trying to get the root span from a request attribute, which psr-15 doesn't do now.
> 
> This is also the main reason why I dislike this change; it re-adds the direct dependency on the PSR15 package for instrumentation that wants to modify the request span.
> 
> This will be problematic if the request is also handled by non-psr implementations. A while ago I wrote a PoC runtime that runs a symfony kernel inside a `amphp/http-server` request handler; the runtime converts the amphp request into psr request into symfony request and maintains all request attributes (common concept for server requests), including the associated root span. Storing the span in an amphp/symfony/etc. specific context key would result in losing the root span association unless every instrumentation checks every context key -> If we want to store the span in the context we should use a context key that is not related to any specific request implementation.

Don't understand your last sentence, why we need check every context key. Is context key I introduced somehow related to specific request implementation?. Anyway, still I think that storing span in context is better than relying on something that we are instrumenting (request in this case) and is self-contained.

pdelewski:
> > Don't know exactly what do you mean, this coverage report?
> 
> No, this: https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/Slim/src/SlimInstrumentation.php#L44 It is trying to get the root span from a request attribute, which psr-15 doesn't do now.

I would remove hooking into `performRouting` and update root span based on information I have on PSR-15 instrumentation.

src/Instrumentation/Psr15/src/Psr15Instrumentation.php
```diff
@@ -24,10 +23,18 @@
  */
 class Psr15Instrumentation
 {
+    private static $rootSpan;
```
pdelewski, line 26: This can be moved out from `Psr15Instrumentation` then we will not depend on it but on something generic

pdelewski:
> This is also the main reason why I dislike this change; it re-adds the direct dependency on the PSR15 package for instrumentation that wants to modify the request span.

One of the solution to this problem would be base class that will be part of SDK and will contain functionality of setting/getting `$rootSpan`

pdelewski:
> > Don't know exactly what do you mean, this coverage report?
> 
> No, this: https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/Slim/src/SlimInstrumentation.php#L44 It is trying to get the root span from a request attribute, which psr-15 doesn't do now.

Just for we can solve that in another PR as we need to introduce dependency to `opentelemetry-php-contrib`, but as I written this dependency can be removed with base class for Instrumentation that will have functionality for setting/getting `$rootSpan` and maybe other things in future

brettmc:
> One of the solution to this problem would be base class that will be part of SDK and will contain functionality of setting/getting `$rootSpan`

I think being able to retrieve the root span is very useful, for precisely this reason. This is actually the very first question I had when trying to use otel for the first time - we don't always have enough information when creating the root span to give it a good  name as advised in spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md?plain=1#L32
Being able to get the root span seems like a really clean way to decouple slim/psr15, and would simplify auto-instrumentation for any framework where routing isn't the very first thing to happen, IMO.

Nevay:
> Is context key I introduced somehow related to specific request implementation?

My comment becomes obsolete if we introduce a global http server span context key; keeping the context key in the PSR15 instrumentation package would likely lead to the introduction of another context key for symfony, laravel etc. -> no longer interoperable. 

How do we handle sub-requests if we store the request span in the context?
For example calling  `$app->handle(new Request('GET', '/example'))` from within a slim route callable should create a new span,  the routing middleware should update the name of this span instead of the initial request span.

brettmc:
> How do we handle sub-requests if we store the request span in the context?

If we activate a context entry using the same key to represent a http server request and any sub-requests, then I think it can work. I'm assuming here that you'd only update a "http server" span's name when processing that request (ie, we're not trying to update the main request from a sub-request)

Here's a draft PR which adds a context key, and an example of it being used: https://github.com/open-telemetry/opentelemetry-php/pull/864 - thoughts?

Nevay:
> If we activate a context entry using the same key to represent a http server request and any sub-requests

How do we detect that we are handling a sub-request?
We cannot rely on object identity because requests are immutable. IMO the only solution is to store some kind of marker in the request attributes (and if we do this: what is the advantage of doing this over just storing the span as request attribute as we do currently?).

pdelewski:
> > If we activate a context entry using the same key to represent a http server request and any sub-requests
> 
> How do we detect that we are handling a sub-request? We cannot rely on object identity because requests are immutable. IMO the only solution is to store some kind of marker in the request attributes (and if we do this: what is the advantage of doing this over just storing the span as request attribute as we do currently?).

Storing additional marker is not a solution because as you mentioned, it's not better from what we already have. What about storing a counter, root span would have always 0 value and sub-request > 0

pdelewski:
Forget about my last comment, it's wrong

pdelewski:
@Nevay @brettmc Starting this PR I had in mind a case where we will need update root span, but not scenario mentioned by @Nevay. As of now, I don't see solution to distinguish sub-request that would be better from what we have now. Having said that, this bothers me as I can imagine library or framework where storing state in external object (the one provided by library) will not be possible

brettmc:
I agree, @pdelewski - I think that using a context key defined in API/SDK can work but it's not better than storing an attribute on the request (plus, it's an extra scope to remember to detach). The other benefit of storing it in the request itself is that there is no doubt about which "root span" belongs to which request when there are multiple.
I think I'm in favour of stopping this PR and continuing to use a request attribute...?

pdelewski:
Yes, we can close it and continue with current approach

PR #87, opened 2022-11-25
migrate context/swoole from main repo

brettmc:
The original PR implementing swoole (https://github.com/open-telemetry/opentelemetry-php/pull/675) recommended that it be moved into contrib "in the future". Today is the future.


brettmc:
I've set it up so that it will be git-split into the destination, which I'll probably need to blow away and recreate. I will also submit a PR to remove swoole from the main repo.

bobstrecansky (approved):

omigafu:
hello! @brettmc 
I don't know how to use this pr in swoole, can you provide any example？

PR #89, opened 2022-12-05
httplug async client auto-instrumentation

brettmc:


src/Instrumentation/HttpAsyncClient/src/HttpAsyncClientInstrumentation.php
```diff
@@ -0,0 +1,120 @@
...
+{
+    public static function register(): void
+    {
+        $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.http-async-client', schemaUrl: TraceAttributes::SCHEMA_URL);
+
+        hook(
+            HttpAsyncClient::class,
+            'sendAsyncRequest',
+            pre: static function (HttpAsyncClient $client, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation): ?array {
+                $request = $params[0] ?? null;
```
pdelewski, line 31: It seems that, as we are handling the same input parameter type, implementation of pre function is exactly the same as for PSR-18. I was thinking if it would not make sense to extract this as common function, however maybe it's tool early for that.

brettmc: I agree, creating/updating spans from psr-7 requests could end up being common to a few different auto-instrumentation implementations. I initially thought to just merge this+psr-18, but splitting it out might be more useful in future.

PR #93, opened 2022-12-14
relax otel requirements, dont gitsplit tags

brettmc:
now that we have a beta, the versioning has changed so update projects that can work with beta.
Symfony doesn't, and needs some attention to make it work with >= 0.0.15, so for now peg it to that version
improve Dockerfile, as the use of semi-colons between commands in a build step causes some failures to be ignored, which is never what we want.

### DIFF
--- a/.gitsplit.yml
+++ b/.gitsplit.yml
@@ -23,5 +23,3 @@ splits:
 origins:
   - ^main$
   - ^split$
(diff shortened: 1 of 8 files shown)


brettmc:
ugh, using beta broke Symfony - I'll see if I can unbreak it

bobstrecansky (approved):

PR #95, opened 2022-12-16
handle slim routing failure

brettmc:
if routing fails, there will be no request parameter to the post hook callback, so make it optional.

### DIFF
--- a/src/Instrumentation/Slim/src/SlimInstrumentation.php
+++ b/src/Instrumentation/Slim/src/SlimInstrumentation.php
@@ -30,6 +30,7 @@ class SlimInstrumentation
          * This relies upon the existence of a request attribute with key SpanInterface::class
          * and type SpanInterface which represents the root span, having been previously set (eg
          * by PSR-15 auto-instrumentation)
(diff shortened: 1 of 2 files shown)


src/Instrumentation/Slim/src/SlimInstrumentation.php
```diff
@@ -37,8 +37,8 @@ public static function register(): void
             RoutingMiddleware::class,
             'performRouting',
             pre: null,
-            post: static function (RoutingMiddleware $middleware, array $params, ServerRequestInterface $request, ?Throwable $exception) {
-                if ($exception) {
+            post: static function (RoutingMiddleware $middleware, array $params, ?ServerRequestInterface $request, ?Throwable $exception) {
```
pdelewski, line 40: IMO, would be good to add comment about routing fail consequences

pdelewski (approved):

src/Instrumentation/Slim/src/SlimInstrumentation.php
brettmc: done!

PR #99, opened 2023-01-10
create root span from slim auto-instrumentation

brettmc:
To allow avoiding all of the spans created by psr-15 auto-instrumentation, have slim auto-instrumentation create its own root span (which is a slightly simplified version of what psr-15 does). Now, slim auto-instrumentation can be used with or without psr-15, depending on whether you want to create spans for all psr-15 middlewares or not.

### DIFF
--- a/src/Instrumentation/Slim/src/SlimInstrumentation.php
+++ b/src/Instrumentation/Slim/src/SlimInstrumentation.php
@@ -5,14 +5,17 @@ declare(strict_types=1);
 namespace OpenTelemetry\Contrib\Instrumentation\Slim;
 
 use OpenTelemetry\API\Common\Instrumentation\CachedInstrumentation;
+use OpenTelemetry\API\Common\Instrumentation\Globals;
 use OpenTelemetry\API\Trace\Span;
 use OpenTelemetry\API\Trace\SpanInterface;
+use OpenTelemetry\API\Trace\SpanKind;
 use OpenTelemetry\API\Trace\StatusCode;
 use OpenTelemetry\Context\Context;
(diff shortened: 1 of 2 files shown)


brettmc:
Testing against the demo quoteservice gives the following traces:

Slim auto-instrumentation:
Screenshot from 2023-01-10 16-15-38

Slim + psr-15 auto-instrumentation:
Screenshot from 2023-01-10 16-15-21

pdelewski:
> To allow avoiding all of the spans created by psr-15 auto-instrumentation, have slim auto-instrumentation create its own root span (which is a slightly simplified version of what psr-15 does). Now, slim auto-instrumentation can be used with or without psr-15, depending on whether you want to create spans for all psr-15 middlewares or not.

@brettmc Is it now possible to disable one or other instrumentation dynamically at runtime (by using API) without manipulating with installed packages?

brettmc:
> Is it now possible to disable one or other instrumentation dynamically at runtime (by using API) without manipulating with installed packages?

No, it's not. This can only be achieved by either installing or not some of the packages.

pdelewski:
> > Is it now possible to disable one or other instrumentation dynamically at runtime (by using API) without manipulating with installed packages?
> 
> No, it's not. This can only be achieved by either installing or not some of the packages.

I'm wondering if that would not be a useful feature

pdelewski (approved):

brettmc:
> I'm wondering if that would not be a useful feature

I think it's a good idea, I'll create an issue for it. I don't think we can un-register hooks once they exist, but perhaps checking a configuration - something like `OTEL_PHP_DISABLED_INSTRUMENTATIONS` which contains a list?

PR #108, opened 2023-02-09
moving WordPress root span into a hook function

brettmc:
starting the root span too eagerly (during composer autoload functions) can lead to a race condition where we try to use the SDK before it has been completely initialized (by other autoload functions). Instead, create the root span in a pre hook on wp_initial_constants, which is the earliest function call I could find that is executed exactly once.

Fixes https://github.com/open-telemetry/opentelemetry-php/issues/920

### DIFF
--- a/src/Instrumentation/Wordpress/README.md
+++ b/src/Instrumentation/Wordpress/README.md
@@ -10,7 +10,7 @@
 * OpenTelemetry [SDK Autoloading](https://github.com/open-telemetry/opentelemetry-php/blob/main/examples/autoload_sdk.php) configured
 
 ## Overview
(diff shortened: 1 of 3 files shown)


pdelewski:
> starting the root span too eagerly (during composer autoload functions) can lead to a race condition where we try to use the SDK before it has been completely initialized (by other autoload functions). Instead, create the root span in a pre hook on wp_initial_constants, which is the earliest function call I could find that is executed exactly once.
> 
> Fixes [open-telemetry/opentelemetry-php#920](https://github.com/open-telemetry/opentelemetry-php/issues/920)

Tested it and it solves problem. Thx

pdelewski (approved):

PR #113, opened 2023-02-11
adding pdo instrumentation to test matrix

pdelewski:


src/Instrumentation/PDO/composer.json
```diff
@@ -8,7 +8,7 @@
   "license": "Apache-2.0",
   "minimum-stability": "dev",
   "require": {
-    "php": "^8.0",
+    "php": "^8.2",
```
brettmc, line 11: this is surprising - what it is about PDO that means we can't instrument it in 8.0 and 8.1 ?

pdelewski: PDO APIs are not user defined functions. It's possible to hook into them since 8.2 https://github.com/php/php-src/commit/625f1649639c2b9a9d76e4d42f88c264ddb8447d. I missed that before as I'm doing testing on 8.2

brettmc (approved):

PR #116, opened 2023-02-21
initial laravel auto instrumentation, root span, kernel::handle hook

pdelewski:


bobstrecansky (approved):

bobstrecansky:
One small CI failure, but everything else looks good!

pdelewski:
> One small CI failure, but everything else looks good!

Yes, the problem is that it's inside laravel itself. Don't know if we can suppress that

PR #118, opened 2023-02-23
adding curl_init and io params

brettmc:
trace curl_init, and some interesting parameters for io calls


brettmc:
Example of `fopen` with params specified:

Screenshot from 2023-02-23 13-01-24

pdelewski (approved):

PR #120, opened 2023-02-25
Symfony auto instrumentation improvements

piotrantosik:
Hi, I add more Symfony instrumentation improvements:
- more tests
- make classes final
- calculate content length if header doesn't exist

Big clap for your work, all works perfect 🎉

### DIFF
--- a/src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
@@ -81,11 +81,14 @@ final class SymfonyInstrumentation
                     }
                     $span->setAttribute(TraceAttributes::HTTP_STATUS_CODE, $response->getStatusCode());
                     $span->setAttribute(TraceAttributes::HTTP_FLAVOR, $response->getProtocolVersion());
+                    $contentLength = $response->headers->get('Content-Length');
                     /** @psalm-suppress PossiblyFalseArgument */
-                    $span->setAttribute(
-                        TraceAttributes::HTTP_RESPONSE_CONTENT_LENGTH,
-                        $response->headers->get('Content-Length', false !== $response->getContent() ? (string) \strlen($response->getContent()) : '0')
-                    );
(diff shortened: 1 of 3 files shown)


src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
```diff
@@ -76,12 +76,19 @@ public static function register(): void
...
+                    if ($response->getStatusCode() >= Response::HTTP_BAD_REQUEST) {
                         $span->setStatus(StatusCode::STATUS_ERROR);
                     }
                     $span->setAttribute(TraceAttributes::HTTP_STATUS_CODE, $response->getStatusCode());
                     $span->setAttribute(TraceAttributes::HTTP_FLAVOR, $response->getProtocolVersion());
-                    $span->setAttribute(TraceAttributes::HTTP_RESPONSE_CONTENT_LENGTH, $response->headers->get('Content-Length'));
+                    $contentLength = $response->headers->get('Content-Length');
+                    /** @psalm-suppress PossiblyFalseArgument */
+                    //BinaryFileResponse and StreamedResponse return boolean as response
+                    if (null === $contentLength && is_string($response->getContent())) {
```
pdelewski, line 87: Don't know if that may happen, but what in the case when `Content-length` header is not provided and `$response->getContent()` returns false ? 
Otherwise LGTM.

piotrantosik: it's set null as before - default value for `$response->headers->get()`, I added tests that cover this case and also removed some redundant parameters.

brettmc (approved):

src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
pdelewski: I missed test. I was just thinking if it makes sense to set this attribute in this case

pdelewski (approved):

PR #121, opened 2023-03-02
moving auto instrumentation installer scripts to contrib

pdelewski:
I'm moving opentelemetry-php-instrumentation/bin to contrib as discussed in https://github.com/open-telemetry/opentelemetry-php-instrumentation/pull/39


.gitsplit.yml
```diff
@@ -28,6 +28,8 @@ splits:
     target: "https://${GH_TOKEN}@github.com/opentelemetry-php/contrib-auto-wordpress.git"
   - prefix: "src/Context/Swoole"
     target: "https://${GH_TOKEN}@github.com/opentelemetry-php/context-swoole.git"
+  - prefix: "src/Instrumentation/bin"
```
brettmc, line 31: I think the directory should be `src/Instrumentation/Installer`

pdelewski: ok, I was thinking about that, but was not sure if this will not be misleading (as someone can think of it as another library that we instrumenting).

brettmc: mmm, good point. What about moving it up to the top level, as `AutoInstrumentationInstaller` or something like that?

pdelewski: Yes, that might be right place. I was also thinking about that, however could not choose meaningful name, `AutoInstrumentationInstaller` is ok for me.

src/AutoInstrumentationInstaller/install-otel-instrumentation
```diff
@@ -0,0 +1,462 @@
...
+);
+
+// core packages are always installed
+// it's possible to get them via packagist API
+// however there is no way at the moment
+// to distinguish core packages from others
+$opentelemetry_core_packages = array(
+  "open-telemetry/sdk", 
+  "open-telemetry/api",
+  "open-telemetry/sdk-contrib",
```
brettmc, line 20: this is a package we've just deprecated. How about just choosing `open-telemetry/exporer-otlp`? (Optionally, with the grpc transport and zipkin exporter, too?)

pdelewski: Yes, that good point. So, for now I will add `open-telemetry/exporer-zipkin` by default.

src/AutoInstrumentationInstaller/install-otel-instrumentation
```diff
@@ -0,0 +1,464 @@
...
+);
+
+// core packages are always installed
+// it's possible to get them via packagist API
+// however there is no way at the moment
+// to distinguish core packages from others
+$opentelemetry_core_packages = array(
+  "open-telemetry/sdk", 
+  "open-telemetry/api",
+  "open-telemetry/exporer-otlp",
```
brettmc, line 20:
```suggestion
  "open-telemetry/exporter-otlp",
```

src/AutoInstrumentationInstaller/install-otel-instrumentation
```diff
@@ -0,0 +1,464 @@
...
+  "open-telemetry/exporer-zipkin",
```
brettmc, line 21:
```suggestion
  "open-telemetry/exporter-zipkin",
```

brettmc (changes requested):

src/AutoInstrumentationInstaller/install-otel-instrumentation
pdelewski: Thx for catching that

brettmc (approved):

PR #124, opened 2023-03-07
adding io readme.md

pdelewski:
- adding missing readme
- adding preferred installation way section

### DIFF
--- a/src/Instrumentation/IO/README.md
+++ b/src/Instrumentation/IO/README.md
@@ -1,8 +1,8 @@
 # OpenTelemetry IO auto-instrumentation
 
-**Preferred and simplest way to install auto-instrumentation is to use opentelemetry-instrumentation-installer.**
+**Preferred and simplest way to install auto-instrumentation (c extension plus instrumentation libraries) is to use [opentelemetry-instrumentation-installer](https://github.com/open-telemetry/opentelemetry-php-contrib/tree/main/src/AutoInstrumentationInstaller).**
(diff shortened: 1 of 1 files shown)


src/Instrumentation/IO/README.md
```diff
@@ -0,0 +1,67 @@
+# OpenTelemetry IO auto-instrumentation
+
```
brettmc, line 2: maybe also point to the auto-instrumentation readme here, for other installation methods. Or, move this text into auto-instrumentation readme?

pdelewski: > maybe also point to the auto-instrumentation readme here, for other installation methods. Or, move this text into auto-instrumentation readme?

@brettmc  Not sure if I understand it, when you say methods, you mean packages?
BTW. I planned to add the same section to all supported `auto` packages

brettmc: I think the documentation on how to install the extension should just be written once, in the extension's readme: https://github.com/open-telemetry/opentelemetry-php-instrumentation/blob/main/README.md#installation
Then, we link to that from all `auto` packages (which reduces some repetition). wdyt?

pdelewski: This is not only about extension, `open-telemetry/opentelemetry-instrumentation-installer` installs everything auto-instrumentation need, but yes we can add description there and link to that.

pdelewski: @brettmc Is it better now?

brettmc: Maybe call out that the easiest way to install auto-instrumentation _plus implementations_ is the installer... or manually via <link to extension's readme> ?
I just want to make sure all options are documented, eg the installer might not be the best fit for automation/dockerfile etc

pdelewski: I agree that we should document all options. Will update it

brettmc (approved):

PR #133, opened 2023-03-28
Be less restrictive when autoloading instrumentation libraries

xvilo:
Emit a warning through PHP's `trigger_error` instead of using an `assert` function call. This allows one to use the libraries in environments where the extension might not be available. However, in case of issues, it will be clear through logs why code might not be instrumented. 

The following approach was shortly discussed in the CNCF Slack, which should be fine for the use case at hand. The project can still run in environments that are not compatible or that are misconfigured. 

I would personally also add a check for the `OTEL_PHP_AUTOLOAD_ENABLED` to let these autoload/configure. Through that, one could use the code in the libraries but not autoload it, or just let it autoload when needed. This can be discussed at a later point in time, though.

As a side note, I noticed that only the WordPress auto-instrumenter had an import for the Instrumentation class, so I took the liberty to add imports in all the files. Now they are more the same.

### DIFF
--- a/src/Instrumentation/IO/_register.php
+++ b/src/Instrumentation/IO/_register.php
@@ -4,8 +4,10 @@ declare(strict_types=1);
 
 use OpenTelemetry\Contrib\Instrumentation\IO\IOInstrumentation;
 
-if (extension_loaded('otel_instrumentation') === true) {
-    IOInstrumentation::register();
-} else {
+if (extension_loaded('otel_instrumentation') === false) {
     trigger_error('The otel_instrumentation extension must be loaded in order to autoload the OpenTelemetry IO auto-instrumentation', E_USER_WARNING);
+
+    return;
 }
+
+IOInstrumentation::register();
--- a/src/Instrumentation/IO/_register.php
+++ b/src/Instrumentation/IO/_register.php
@@ -2,6 +2,10 @@
 
 declare(strict_types=1);
 
-assert(extension_loaded('otel_instrumentation'));
+use OpenTelemetry\Contrib\Instrumentation\IO\IOInstrumentation;
 
-\OpenTelemetry\Contrib\Instrumentation\IO\IOInstrumentation::register();
+if (extension_loaded('otel_instrumentation') === true) {
+    IOInstrumentation::register();
+} else {
+    trigger_error('The otel_instrumentation extension must be loaded in order to autoload the OpenTelemetry IO auto-instrumentation', E_USER_WARNING);
+}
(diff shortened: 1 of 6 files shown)


src/Instrumentation/IO/_register.php
```diff
@@ -2,6 +2,10 @@
 
 declare(strict_types=1);
 
-assert(extension_loaded('otel_instrumentation'));
+use OpenTelemetry\Contrib\Instrumentation\IO\IOInstrumentation;
 
-\OpenTelemetry\Contrib\Instrumentation\IO\IOInstrumentation::register();
+if (extension_loaded('otel_instrumentation') === true) {
```
pdelewski, line 7: We could also reverse condition to check it first and get rid of else branch
```
if (extension_loaded('otel_instrumentation') === false) {
   trigger_error('The otel_instrumentation extension must be loaded in order to autoload the OpenTelemetry IO 
   auto-instrumentation', E_USER_WARNING);
}
IOInstrumentation::register();
```

Overall LGTM

pdelewski: > This will _always_ call the `IOInstrumentation::register()` static methods, which we don't want to do when the `otel_instrumentation` extension is not available/loaded. However, we could do this approach if we add a `return;` statement just after the `trigger_error`. Is this preferred over the current code?

Yes, I forgot to add return. Don't think that there is a convention. This is just my preference.

xvilo: I have applied your feedback 👍

brettmc:
This PR reminded me that we also have the capability to turn off auto-instrumentations by config, since https://github.com/open-telemetry/opentelemetry-php/pull/909, but it has not been added to the . I think it should be an extra check right where you have added the various _register scripts. How do you feel about adding that here?

xvilo:
That would be a great idea @brettmc, I didn't know that was already there. However, I do feel personally a discussion is in place about certain things and would like to discuss this first in an issue. So for now, I'd like this PR scoped to just replacing the `assert` function calls. 

Furthermore, I have fixed the code style issues for some sub-libraries. We might want to stream-line those as well, as not all projects have php-cs-fixer in them, but that's also something for later.

pdelewski:
I forgot about this capability https://github.com/open-telemetry/opentelemetry-php/pull/909, but I agree that this is right place to discuss such things

brettmc (approved):

brettmc:
created https://github.com/open-telemetry/opentelemetry-php/issues/947 for somebody to implement disabling via config

PR #137, opened 2023-03-29
Enable or disabled auto instrumentation

xvilo:
This is done through the `OpenTelemetry\SDK\Sdk::isInstrumentationDisabled` method, which was created in https://github.com/open-telemetry/opentelemetry-php/pull/909/files/. For now, I took the same names as in the example and expanded on those. Please see the discussions in line on some questions I had implementing this. 

Closes: https://github.com/open-telemetry/opentelemetry-php/issues/947


src/Instrumentation/Slim/_register.php
```diff
@@ -3,6 +3,11 @@
 declare(strict_types=1);
 
 use OpenTelemetry\Contrib\Instrumentation\Slim\SlimInstrumentation;
+use OpenTelemetry\SDK\Sdk;
+
+if (Sdk::isInstrumentationDisabled('slim') === true) {
```
xvilo, line 8: While I only had it with Slim, is it an idea to add `-framework` to this one (and the others?) this is fine too with me, but just as a general note

pdelewski: We have to document that somewhere. I'm thinking whether we can verify values passed via `OTEL_PHP_DISABLED_INSTRUMENTATION`, it's easy to make mistake

brettmc: Agreed that it will need to be documented, we should probably list them over in our docs in opentelemetry.io
Rather than using string values such as `slim`, I'd like to see constants, eg `SlimInstrumentation::NAME`.
re: naming, I think that the major frameworks are well-known enough that "slim", "laravel", "symfony" etc will be okay

xvilo: Alright, so if I understand correctly, you want the names to be added to their respective `*Instrumentation`-classes and then used in these places. 

I at first thought that @pdelewski meant that we needed to add these to the main repo like we have for other OTEL vars. Which I wasn't totally sure about. 

I'll update the repo to add the names in public constants and use them here. I'll open a docs PR later (if thats possible and if I find the location of them :D )

pdelewski: I was also thinking about using constants instead of strings

src/Instrumentation/HttpAsyncClient/_register.php
```diff
@@ -2,4 +2,11 @@
 
 declare(strict_types=1);
 
-\OpenTelemetry\Contrib\Instrumentation\HttpAsyncClient\HttpAsyncClientInstrumentation::register();
+use OpenTelemetry\Contrib\Instrumentation\HttpAsyncClient\HttpAsyncClientInstrumentation;
+use OpenTelemetry\SDK\Sdk;
+
+if (Sdk::isInstrumentationDisabled('http-async-client') === true) {
```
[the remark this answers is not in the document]
brettmc: According to spec, libraries should not assume that an SDK is installed: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#sdk, so this might also need a further check (I thought `class_exists` would be enough)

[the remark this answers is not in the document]
xvilo: What would be the logic if the SDK is _not_ installed? 
1. Don't run any code from the `_register.php` files (early return)
2. Do run the code, but don't do this check, and assume that instrumentation is enabled
3. Run any other content in this file, but assume instrumentation is disabled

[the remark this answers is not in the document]
xvilo: I have implemented a `class_exist` check inline, and ignore the check if we don't have the SDK library

src/Instrumentation/HttpAsyncClient/_register.php
```diff
@@ -2,4 +2,11 @@
 
 declare(strict_types=1);
 
-\OpenTelemetry\Contrib\Instrumentation\HttpAsyncClient\HttpAsyncClientInstrumentation::register();
+use OpenTelemetry\Contrib\Instrumentation\HttpAsyncClient\HttpAsyncClientInstrumentation;
+use OpenTelemetry\SDK\Sdk;
+
+if (class_exists(Sdk::class) && Sdk::isInstrumentationDisabled(HttpAsyncClientInstrumentation::NAME) === true) {
```
brettmc, line 8: what happens here if no SDK and no extension? we likely want another condition for "not extension installed"?

src/Instrumentation/Psr18/composer.json
```diff
@@ -32,6 +32,13 @@
     "psalm/plugin-phpunit": "^0.16",
     "open-telemetry/sdk": "^1.0",
     "phpunit/phpunit": "^9.5",
-    "vimeo/psalm": "^4.0"
+    "vimeo/psalm": "^4.0",
+    "symfony/http-client": "^6.2",
+    "guzzlehttp/promises": "^1.5"
```
brettmc, line 37: Are these needed? I don't see any corresponding test changes that would require them...

src/Instrumentation/HttpAsyncClient/_register.php
xvilo: Good spot, the warning was not added here yet. I have added it here too

src/Instrumentation/Psr18/composer.json
xvilo: Nope, and I did not do this myself. I suspect it's the `php-http/discovery` plugin. Removed both changes for now

brettmc:
@xvilo you can ignore the Symfony failures, we know about that one and somebody is working on a fix in another PR.

xvilo:
@brettmc in that case, feel free to merge!

xvilo:
@pdelewski should I wait (for you) to get the last remaining CI issues fixed? I see you added the latest changes from main into my branch, but now we have new CI errors :D

pdelewski:
> @pdelewski should I wait (for you) to get the last remaining CI issues fixed? I see you added the latest changes from main into my branch, but now we have new CI errors :D

No, however it would be good to fix laravel 
```
vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.php -v --using-cache=no -vvv
```
and suppress Psr18 issues.

Symfony requires more time.

xvilo:
Fixed the `php-cs-fixer` and `psalm` issues, also brought the branch up to date which fixed the Symfony issues. The checks are all passing right now 👍

src/Instrumentation/HttpAsyncClient/_register.php
```diff
@@ -2,4 +2,17 @@
...
+    return;
+}
+
+if (extension_loaded('otel_instrumentation') === false) {
```
pdelewski, line 12: @brettmc opened PR that changes name of extension, after https://github.com/open-telemetry/opentelemetry-php-instrumentation/pull/49 it will be `opentelemetry`, we have to synchronize between the two.

pdelewski:
LGTM

src/Instrumentation/HttpAsyncClient/_register.php
xvilo: I would say, lets merge this first as the other PR still needs to get merged and a new version tag has to be made available. Renaming is a good choice imho, but it requires A LOT more changes through the repo which could be an easy search-replace

pdelewski:
@xvilo Could you also extend README files with information about disabling particular package?

xvilo:
Good suggestion, I have added a small section, reference to General SDK Configuration in case somebody might be confused and added the name in the README as an example. Also added that these are comma separated and took the last line from the PR that added support for this.

src/Instrumentation/Laravel/README.md
```diff
@@ -49,3 +49,14 @@ From Laravel subdirectory:
...
+## Configuration
+
+Parts of this auto-instrumentation library can be configured, more options are available throught the
+[General SDK Configuration](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration):
+
+| Name                                | Default value | Values                  | Example | Description                                                                     |
+|-------------------------------------|---------------|-------------------------|---------|---------------------------------------------------------------------------------|
+| OTEL_PHP_DISABLED_INSTRUMENTATIONS  | []            | Instrumentation name(s) | laravel | Disable one or more installed auto-instrumentations, names are comma seperated. |
+
+Configurations can be provided as environment variables, or via `php.ini` (or a file included by `php.ini`)
```
pdelewski, line 62: I'm not sure that we can set this via `php.ini` (at least I'm not aware of that)

src/Instrumentation/HttpAsyncClient/README.md
```diff
@@ -36,4 +36,15 @@ $tracerProvider->shutdown();
 
 ```bash
 $ composer require open-telemetry/opentelemetry-auto-http-async
-```
\ No newline at end of file
+```
+
+## Configuration
```
pdelewski, line 41:
```suggestion
## Disabling instrumentation
```

src/Instrumentation/HttpAsyncClient/README.md
```diff
@@ -36,4 +36,15 @@ $tracerProvider->shutdown();
...
+
+Parts of this auto-instrumentation library can be configured, more options are available throught the
```
pdelewski, line 43:
```suggestion
You can disable auto-instrumentation of this library by using following environment variable, more settings are available through the 
```

src/Instrumentation/HttpAsyncClient/README.md
pdelewski: I was thinking about being more specific

src/Instrumentation/HttpAsyncClient/README.md
xvilo: In that case the table doesn't make sense, and it doesn't allow for future expansion. Either keep it like this, or we should replace it with just a "disable" section that explains the env var name

src/Instrumentation/Laravel/README.md
xvilo: Thats what the main docs says about this feature

pdelewski: ok, you are right

xvilo:
@pdelewski I have resolved all the discussions, feel free to merge 👍

src/Instrumentation/HttpAsyncClient/_register.php
piotrantosik: maybe it's better to add extension name as constant variable to `open-telemetry/api` - it's required in all packages?

xvilo:
@brettmc Made a PR https://github.com/open-telemetry/opentelemetry-php-contrib/pull/140 feel free to add some commits later, for now I'm a bit busy

PR #143, opened 2023-04-07
Add TraceResponse propagator

cedricziel:
**Note:** This introduces a new experimental package as `traceresponse` is currently a w3c editors' draft.

This package provides a [Trace Context HTTP Response Headers Format](https://w3c.github.io/trace-context/#trace-context-http-response-headers-format)
propagator to inject the current span context into Response datastructures.

The main goal is to allow client-side technology (Real User Monitoring, HTTP Clients) to record
the server side context in order to allow referencing or assuming it.

### DIFF
--- a/src/Propagation/TraceResponse/composer.json
+++ b/src/Propagation/TraceResponse/composer.json
@@ -1,0 +1,39 @@
+{
+  "name": "open-telemetry/opentelemetry-propagation-traceresponse",
+  "description": "OpenTelemetry traceresponse propagator.",
+  "keywords": ["opentelemetry", "otel", "open-telemetry", "propagator", "traceresponse"],
+  "type": "library",
+  "homepage": "https://opentelemetry.io/docs/php",
+  "readme": "./README.md",
+  "license": "Apache-2.0",
+  "minimum-stability": "dev",
+  "prefer-stable": true,
+  "require": {
+    "php": "^7.0|^8.0",
+    "open-telemetry/context": "^1.0"
+  },
+  "autoload": {
+    "psr-4": {
+      "OpenTelemetry\\Contrib\\Propagation\\TraceResponse\\": "src/"
+    }
+  },
+  "require-dev": {
+    "friendsofphp/php-cs-fixer": "^3",
+    "phan/phan": "^5.0",
+    "phpstan/phpstan": "^1.1",
+    "phpstan/phpstan-phpunit": "^1.0",
+    "psalm/plugin-phpunit": "^0.16",
+    "open-telemetry/sdk": "^1.0",
+    "phpunit/phpunit": "^9.5",
+    "vimeo/psalm": "^4.0",
+    "symfony/http-client": "^5.4|^6.0",
+    "guzzlehttp/promises": "^1.5",
+    "php-http/message-factory": "^1.0",
+    "nyholm/psr7": "^1.5"
+  },
+  "config": {
+    "allow-plugins": {
+      "php-http/discovery": true
+    }
+  }
+}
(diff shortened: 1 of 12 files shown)


src/Propagation/TraceResponse/tests/Unit/PropagatorTest.php
```diff
@@ -0,0 +1,124 @@
...
+    private const TRACE_ID = '5759e988bd862e3fe1be46a994272793';
+    private const SPAN_ID = '53995c3f42cd8ad8';
+    private const TRACERESPONSE_HEADER_SAMPLED = '00-5759e988bd862e3fe1be46a994272793-53995c3f42cd8ad8-1';
+    private const TRACERESPONSE_HEADER_NOT_SAMPLED = '00-5759e988bd862e3fe1be46a994272793-53995c3f42cd8ad8-0';
+
+    /**
+     * @test
+     * fields() should return an array of fields with AWS X-Ray Trace ID Header
+     */
+    public function TestFields()
```
brettmc, line 28: project style is for test names to be snake_case, please

src/Propagation/TraceResponse/composer.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "open-telemetry/opentelemetry-propagation-traceresponse",
```
brettmc, line 2: I'm happy with this name. I've created https://github.com/opentelemetry-php/contrib-propagator-traceresponse which will be the git-split target for this directory. Can you add an entry to the top-level `.gitsplit.yaml` file for this dir. After this is merged, I'll add it to packagist and tag.

src/Propagation/TraceResponse/tests/Unit/PropagatorTest.php
```diff
@@ -0,0 +1,124 @@
...
+class PropagatorTest extends TestCase
+{
+    private const TRACE_ID = '5759e988bd862e3fe1be46a994272793';
+    private const SPAN_ID = '53995c3f42cd8ad8';
+    private const TRACERESPONSE_HEADER_SAMPLED = '00-5759e988bd862e3fe1be46a994272793-53995c3f42cd8ad8-1';
+    private const TRACERESPONSE_HEADER_NOT_SAMPLED = '00-5759e988bd862e3fe1be46a994272793-53995c3f42cd8ad8-0';
+
+    /**
+     * @test
+     * fields() should return an array of fields with AWS X-Ray Trace ID Header
```
brettmc, line 26: update comment to be about traceresponse

brettmc:
A couple of small things from me, but this looks good.

We might need to think about how this propagator interacts with the tracecontext one; for example we current can globally configure a propagator which is then accessible from `API\Globals::propagator()`. The spec talks about a [composite propagator](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#composite-propagator) - we don't have one yet (probably because it has not been needed before now)

cedricziel:
> We might need to think about how this propagator interacts with the tracecontext one; for example we current can globally configure a propagator which is then accessible from API\Globals::propagator(). The spec talks about a [composite propagator](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md?rgh-link-date=2023-04-07T12%3A42%3A59Z#composite-propagator) - we don't have one yet (probably because it has not been needed before now)

True - the reason i didnt opt for this, is that traceresponse is strictly limited to responses and without a specific setter implementation, the propagator wont know.

My first approach was to explore in core to have `Globals::responsePropagators()`, but that seemed to be too invasive for an experimental package.

cedricziel:
Note that I separated out Symfony (#146) and Slim (#147) auto propagation so we can discuss them separately after this is merged, gitsplit happened and the package has been created.

brettmc (approved):

PR #147, opened 2023-04-07
Add traceresponse propagation to Slim auto-instrumentation

cedricziel:
Depends on #143


src/Instrumentation/Slim/src/SlimInstrumentation.php
```diff
@@ -61,7 +62,7 @@ public static function register(): void
 
                 return [$request];
             },
-            post: static function (App $app, array $params, ?ResponseInterface $response, ?Throwable $exception) {
+            post: static function (App $app, array $params, ?ResponseInterface &$response, ?Throwable $exception) {
```
brettmc, line 65: My understanding is that you can modify return value without pass-by-reference by returning the updated value/object: https://github.com/open-telemetry/opentelemetry-php-instrumentation/blob/main/tests/010.phpt

cedricziel: Changed to return instead of reference

PR #148, opened 2023-04-12
Monolog handler

brettmc:
initial implementation of a logs bridge api for opentelemetry logging


src/Logs/Monolog/src/Handler.php
```diff
@@ -0,0 +1,58 @@
...
+namespace OpenTelemetry\Contrib\Logs\Monolog;
+
+use Monolog\Formatter\NormalizerFormatter;
+use Monolog\Handler\AbstractHandler;
+use Monolog\Level;
+use Monolog\LogRecord;
+use OpenTelemetry\API\Common\Instrumentation\Globals;
+use OpenTelemetry\API\Logs as API;
+
+class Handler extends AbstractHandler
```
Nevay, line 14: Could extend `AbstractProcessingHandler`.

src/Logs/Monolog/src/Handler.php
```diff
@@ -0,0 +1,58 @@
...
+        if (!$this->isHandling($record)) {
+            return false;
+        }
+        $normalized = $this->normalizer->format($record);
+        $logRecord = (new API\LogRecord())
+            ->setTimestamp($record->datetime->getTimestamp() * API\LogRecord::NANOS_PER_SECOND)
+            ->setSeverityNumber(API\Map\Psr3::severityNumber($record->level->toPsrLogLevel()))
+            ->setSeverityText($record->level->toPsrLogLevel())
+            ->setBody($normalized['message'])
+            ->setAttributes(array_merge($normalized['context'], $normalized['extra']))
```
Nevay, line 41:
```suggestion
            ->setAttributes($normalized['context'] + $normalized['extra'])
```
IMO `context` should take precedence over `extra`.

src/Logs/Monolog/src/Handler.php
```diff
@@ -0,0 +1,58 @@
...
+
+class Handler extends AbstractHandler
+{
+    private API\LoggerInterface $logger;
+    private NormalizerFormatter $normalizer;
+
+    /**
+     * @psalm-suppress ArgumentTypeCoercion
+     */
+    public function __construct(API\LoggerProviderInterface $loggerProvider = null, int|string|Level $level = Level::Debug, bool $bubble = true)
```
Nevay, line 22: Would prefer if `$loggerProvider` is a required argument.

src/Logs/Monolog/src/Handler.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        parent::__construct($level, $bubble);
+        $this->normalizer = new NormalizerFormatter();
+        $loggerProvider ??= Globals::loggerProvider();
+        $this->logger = $loggerProvider->getLogger('monolog');
+    }
+
+    public function handle(LogRecord $record): bool
```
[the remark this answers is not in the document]
brettmc: I've been able to make it work back to monolog v1.1 and php 7.4, which is almost everything. Prior to 1.1, composer didn't support autoloading.

src/Logs/Monolog/src/Handler.php
brettmc: done, and added a test for this

src/Logs/Monolog/src/Handler.php
```diff
@@ -0,0 +1,38 @@
...
+     * @phan-suppress PhanTypeMismatchArgument
+     * @psalm-suppress InvalidOperand
+     */
+    protected function write($record): void
+    {
+        $logRecord = (new API\LogRecord())
+            ->setTimestamp((int) ($record['datetime']->format('Uu') * 1000))
+            ->setSeverityNumber(API\Map\Psr3::severityNumber($record['level_name']))
+            ->setSeverityText($record['level_name'])
+            ->setBody($record['message'])
```
Nevay, line 33:
```suggestion
            ->setBody($record['formatted'])
```
and change `::getDefaultFormatter()` to a formatter that returns `$record['message']` (could use `new LineFormatter('%message%')` if we do not care about the unnecessary normalization) to allow users to change the body format.

Nevay (commented):
We should also include `$record['channel']` as attribute.
Would it make sense to record `context`, `extra` and `channel` as separate attributes to prevent overriding data?

brettmc:
> Would it make sense to record context, extra and channel as separate attributes to prevent overriding data?

I've put them all into their own attributes (and set the default formatter to `NormalizerFormatter` so that exceptions can be logged). I think let's get a 0.0.1 release out, and see whether the log record structure is useful when people actually use it.

src/Logs/Monolog/src/Handler.php
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public function __construct(API\LoggerProviderInterface $loggerProvider, $level, bool $bubble = true)
+    {
+        parent::__construct($level, $bubble);
+        $this->logger = $loggerProvider->getLogger('monolog');
+    }
+
+    protected function getDefaultFormatter(): FormatterInterface
+    {
+        return new NormalizerFormatter('Uu');
```
Nevay, line 27:
```suggestion
        return new NormalizerFormatter();
```
I don't think we should change the dateFormat as it affects datetimes in `context` and `extra` too; I think we should populate timestamp, severity and channel from the original record and use the formatted record for body, context and extra.

bobstrecansky (approved):

PR #161, opened 2023-05-26
[WIP] Decrease minimum PDO autoinstrumentation PHP version

piotrantosik:


bobstrecansky (approved):

bobstrecansky:
Looks like the 8.0 PDO tests are failing (as expected)?

src/Instrumentation/PDO/composer.json
```diff
@@ -8,7 +8,8 @@
   "license": "Apache-2.0",
   "minimum-stability": "dev",
   "require": {
-    "php": "^8.2",
+    "php": "^8.0",
```
pdelewski, line 11: PDO requires at least 8.2 as since this version hooking into built-in functions is possible

pdelewski:
> Looks like the 8.0 PDO tests are failing (as expected)?

They are failing as PDO instrumentation requires at least 8.2

piotrantosik:
> > Looks like the 8.0 PDO tests are failing (as expected)?
> 
> They are failing as PDO instrumentation requires at least 8.2

It's not 100% true, I use it with 8.0, and works perfectly (custom class implementation). Now I want to switch to an official library, I try to finish this PR this week

pdelewski:
> > > Looks like the 8.0 PDO tests are failing (as expected)?
> > 
> > 
> > They are failing as PDO instrumentation requires at least 8.2
> 
> It's not 100% true, I use it with 8.0, and works perfectly (custom class implementation). Now I want to switch to an official library, I try to finish this PR this week

Just to clarify, so in your case PDO is implemented as user kind classes/interfaces, not kind of builtin/extension stuff? Am I correct?

piotrantosik:
> 

I use Doctrine but behind is PDO, and I use normal hook from opentelemetry:

```php
hook(
            \PDO::class,
            '__construct',
            pre: static function (
                \PDO $pdo,
                array $params,
                string $class,
                string $function,
                ?string $filename,
                ?int $lineno,
            ) use ($instrumentation): void {
```

As I understand, from 8.2 it's possible to observe native PHP function, but for classes such as PDO should work from 8.0

piotrantosik:
Finally, I check this and I can confirm that don't works < 8.2, you were right.
In my project works, because I use Doctrine DBAL 2.x which PDO implementation extends PDO class, see: https://github.com/doctrine/dbal/blob/v2.10.1/lib/Doctrine/DBAL/Driver/PDOConnection.php

PR #166, opened 2023-06-14
laravel: improve events watcher registration

xdavidwu:
The original code hooks on ServiceProvider::{__construct,boot}. Hooking on ServiceProvider is a bad idea since it is the parent of all service providers and thus will hook on all of them. `boot` hook may be run multiple times, thus the $watcherInstalled check is needed. Even worse, only a few providers have a `boot` method, making it more unreliable.

This improves it by registering the watchers after application bootstrap, via Application::booted() Laravel-native hook. The whole application will be only run once, thus a $watcherInstalled check is not needed.

To get the Application object, Kernel::__construct is used.

$watcherInstalled in original code is also broken by the fact that it will be reset on ServiceProvider::__construct, which may happen in the middle of service providers bootstrapping process, via Laravel "deferred" provider registration, which is, when booting a provider, it requests service from another provider that is not registered yet. This may lead to the watchers being registered multiple times as seen in open-telemetry/opentelemetry-php#1032.

Fixes: open-telemetry/opentelemetry-php#1032

### DIFF
--- a/src/Instrumentation/Laravel/src/LaravelInstrumentation.php
+++ b/src/Instrumentation/Laravel/src/LaravelInstrumentation.php
@@ -7,7 +7,6 @@ namespace OpenTelemetry\Contrib\Instrumentation\Laravel;
 use Illuminate\Foundation\Application;
 use Illuminate\Foundation\Http\Kernel;
 use Illuminate\Http\Request;
-use Illuminate\Support\ServiceProvider;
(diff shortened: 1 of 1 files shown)


brettmc (approved):

brettmc:
@xdavidwu linting didn't like one file (unused import), can you fix that?

PR #168, opened 2023-06-19
Update wordpress example to work without lockfile

aabmass:
When I commited the example originally, I had the lockfile locally but it is actually gitignored causing the example to fail in a clean repo. Removed it from the dockerfile to fix.

Or should I check-in a lockfile?


brettmc (approved):

brettmc:
> Or should I check-in a lockfile?

I think without lockfile.

PR #169, opened 2023-06-19
Symfony Bundle compatibility updates for beta

wadjei:
* Change configuration to be based on SpanExporterFactory definitions instead of Exporter factories - this facilitates the injection of TransportInterfaces rather than the client etc.
* Remove references to obsolete, removed Jaeger and NewRelic exporters
* Updated unit and integration tests


src/Symfony/src/OtelSdkBundle/DependencyInjection/ConfigMappings.php
```diff
@@ -26,4 +26,8 @@ interface ConfigMappings
         Configuration::OTLP_GRPC_EXPORTER => SpanExporters::OTLP_GRPC,
         Configuration::ZIPKIN_TO_NEWRELIC_EXPORTER => SpanExporters::ZIPKIN_TO_NEWRELIC,
```
brettmc, line 27: anything related to NewRelic can be removed - we dropped their custom exporters as they now fully support otlp

src/Symfony/src/OtelSdkBundle/DependencyInjection/Configuration.php
```diff
@@ -94,13 +94,13 @@ class Configuration implements ConfigurationInterface
     public const OTLP_HTTP_EXPORTER = 'otlphttp';
     public const OTLP_GRPC_EXPORTER = 'otlpgrpc';
     public const ZIPKIN_TO_NEWRELIC_EXPORTER = 'zipkintonewrelic';
```
brettmc, line 96: remove

src/Symfony/tests/Integration/OtelSdkBundle/DependencyInjection/OtelSdkExtensionTest.php
```diff
@@ -5,6 +5,8 @@
 namespace OpenTelemetry\Tests\Symfony\Integration\OtelSdkBundle\DependencyInjection;
 
 use Exception;
+use OpenTelemetry\Contrib\Newrelic\SpanExporterFactory as NewrelicSpanExporterFactory;
```
brettmc, line 8:
```suggestion
```

src/Symfony/tests/Integration/OtelSdkBundle/DependencyInjection/config/exporters/config.yaml
```diff
@@ -7,24 +7,17 @@ trace:
   exporters:
     # dsn example
     exporter1: zipkin+http://zipkinhost:1234/path
```
brettmc, line 9: DSN-based config has just been deprecated. I'm in favour of removing it from here, unless somebody has a good argument not to...

src/Symfony/tests/Integration/OtelSdkBundle/DependencyInjection/config/exporters/config.yaml
```diff
@@ -7,24 +7,17 @@ trace:
...
-    # dsn key example
-    exporter2:
-      dsn: jaeger+https://foo:bar@jaegerhost:1234/path/to/endpoint?name=foo
     # parameter example
-    exporter3:
+    exporter2:
       type: newrelic
```
brettmc, line 12: change this to zipkin? (newrelic exporter no longer a thing)

brettmc (commented):
Looking good! Just a couple of minor comments from me - newrelic and DSN-based config. There were more newrelic mentions in the code that I didn't comment on, could you search and remove?

src/Symfony/tests/Integration/OtelSdkBundle/DependencyInjection/config/exporters/config.yaml
wadjei: I opted to switch this to otlp (seeing as there were no tests for it before)

brettmc:
> there's a lot of new failures in AWS, Logs/Monolog and Instrumentation

Actively fixing those, most are OK now if you re-base. In any case, definitely not caused by you and I don't think a blocker for this PR.

brettmc (approved):

brettmc:
@wadjei let's get this merged. There's a failing style check for symfony that should be an easy fix (`make style` locally), then if it goes green I'll merge it.

src/Symfony/phpstan.neon
```diff
@@ -0,0 +1,13 @@
+includes:
+    - vendor/phpstan/phpstan-phpunit/extension.neon
+    - vendor/phpstan/phpstan-symfony/extension.neon
+    - vendor/jangregor/phpstan-prophecy/extension.neon
```
wadjei, line 4: I wanted to avoid adding this config but without it phpstan cannot deal with prophecy mocks

brettmc (approved):

wadjei:
@brettmc thanks for reviewing and merging... A few minor things left over
* when running the test workflow, there were differences in the static analysis stages between PHP 7.4/8.0 and 8.1/8.2 which could not be resolved in one without breaking the other. I came up with a minor re-write to get around that - shall I open a PR for it?
* The namespace of the Symfony bundle now differs between the monorepo's psr-4 definition and that of the bundle itself. Should we refactor the bundle so it matches in all cases? (OpenTelemetry\\Contrib\\Symfony looks more correct to me now)
* Finally the readme of the bundle no longer offers any info for anyone trying it out. is it time to restore the old one?

PR #175, opened 2023-06-28
Initial version of MongoDB instrumentation

lstrojny:
Uses the MongoDB APM subscriber infrastructure instead of hooks, so it doesn’t even require the `opentelemetry` extension.

The sub-project setup was blatantly stolen from the PDO module so I am 100% I am missing something.

### DIFF
--- a/src/Instrumentation/MongoDB/README.md
+++ b/src/Instrumentation/MongoDB/README.md
@@ -1,14 +1,15 @@
 # OpenTelemetry MongoDB auto-instrumentation
 
 **Preferred and simplest way to install auto-instrumentation (c extension plus instrumentation libraries) is to use [opentelemetry-instrumentation-installer](https://github.com/open-telemetry/opentelemetry-php-contrib/tree/main/src/AutoInstrumentationInstaller).**
-**The same process can be done manually by installing [c extension](https://github.com/open-telemetry/opentelemetry-php-instrumentation#installation) plus all needed instrumentation libraries like [MongoDB](#Installation-via-composer)**
+**To run only this instrumentation, [install via composer](#Installation-via-composer)**
 
 ## Requirements
 
-* OpenTelemetry extension
-* OpenTelemetry SDK and exporters (required to actually export traces)
+-   OpenTelemetry extension
+-   OpenTelemetry SDK and exporters (required to actually export traces)
 
 ## Overview
+
 Auto-instrumentation hooks are registered via composer, and spans will automatically be created for all MongoDB
 operations like `find` or `aggregate`.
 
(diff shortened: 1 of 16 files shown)


src/Instrumentation/MongoDB/_register.php
```diff
@@ -0,0 +1,18 @@
...
+declare(strict_types=1);
+
+use OpenTelemetry\Contrib\Instrumentation\MongoDB\MongoDBInstrumentation;
+use OpenTelemetry\SDK\Sdk;
+
+if (class_exists(Sdk::class) && Sdk::isInstrumentationDisabled(MongoDBInstrumentation::NAME) === true) {
+    return;
+}
+
+if (extension_loaded('opentelemetry') === false) {
```
lstrojny, line 12: Note to self: this can go

src/Instrumentation/MongoDB/phpstan.neon.dist
```diff
@@ -0,0 +1,9 @@
+includes:
+    - vendor/phpstan/phpstan-phpunit/extension.neon
+
+parameters:
+    tmpDir: var/cache/phpstan
+    level: 5
```
lstrojny, line 6: Increase level

src/Instrumentation/MongoDB/psalm.xml.dist
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<psalm
+    errorLevel="3"
```
lstrojny, line 3: Make stricter

brettmc:
Welcome, @lstrojny 
I've taken a first pass at the code, it looks good to me.
I notice the tests are failing for php < 8.2, will be run in lower versions? If not, can you add exclusions to the workflow to disable tests against earlier versions? If possible, it would be good if it could work with as many versions of PHP as we support (7.4+, although most auto-instrumentations only support 8+)

lstrojny:
@brettmc fixed. Supporting 7.4 and up is trivially possible so let’s do that for now. Enums in PHP 8.1 is probably the only interesting feature, especially for the OTEL core libraries. Adopting those will probably require another major version so we can kick that can down the road.

brettmc:
@lstrojny making progress. Integration tests are failing due to no mongodb available.
We can either look to disable those tests, or add a mongo (which is probably a better test).

I had a quick play and it looks like you can conditionally install mongo, so if you try adding a step into `.github/workflows/php.yml` like so, it may just work:

```
    - name: (optional) Setup mongodb
      uses: supercharge/mongodb-github-action@1.8.0
      if: matrix.project == 'Instrumentation/MongoDB'
      with:
        mongodb-version: 6.0
```

lstrojny:
We shouldn't even need a real MongoDB instance as the attempt to query should be good enough. Can you retry?

brettmc:
@lstrojny I did a little more investigation into mongodb. It should be available on localhost:27017, however in my testing there is a race condition: it looks like the tests are running before mongo is ready to accept connections.
I raised an issue against the action, but in the meantime I found that adding a replicaset does have a readiness check, so try adding this to the `with` section: `mongodb-replica-set: test-rs`

src/Instrumentation/MongoDB/README.md
```diff
@@ -0,0 +1,63 @@
+# OpenTelemetry MongoDB auto-instrumentation
+
+**Preferred and simplest way to install auto-instrumentation (c extension plus instrumentation libraries) is to use [opentelemetry-instrumentation-installer](https://github.com/open-telemetry/opentelemetry-php-contrib/tree/main/src/AutoInstrumentationInstaller).**
```
xvilo, line 3: Do we need the C-Extension or could we leave this out?

lstrojny: We don’t need it, no. So we can leave that part out

lstrojny: Thought about it more: we could leave it out, but then people wouldn’t be pointed to the auto installer and wouldn't install the extension. So instead I added a comment about standalone integration but still mentioned the installer as the preferred way.

brettmc (approved):

brettmc:
Released to packagist as v0.0.1

PR #176, opened 2023-07-07
fix auto-instrumentation installer

brettmc:
* install from PECL rather than github
* allow choice of stability for core packages
* create .ini file in ini_dir
* use psr/http-client-implementation to match upstream change
* add a dockerfile for local dev
* only run composer update once at the end
* ignore case for questions
* minor tidy up

Fixes: https://github.com/open-telemetry/opentelemetry-php-instrumentation/issues/67


bobstrecansky (approved):

bobstrecansky:
Approved with the caveat of that one unit test being fixed on the PHP 7.4 github action run

PR #184, opened 2023-08-07
Laravel Instrumentation

ChrisLightfootWild:
Hello 👋 

I'm opening this as a draft, while I test it locally. This PR moves watcher registration from the Http Kernel into the Application contract.

This should allows for watcher registration in console Kernel contexts, such as commands that are configured to run in the Kernel scheduler.

I have tested this locally in Laravel 10 only so far.


pdelewski:
> Hello 👋
> 
> I'm opening this as a draft, while I test it locally. This PR moves watcher registration from the Http Kernel into the Application contract.
> 
> This should allows for watcher registration in console Kernel contexts, such as commands that are configured to run in the Kernel scheduler.
> 
> I have tested this locally in Laravel 10 only so far.

Are all tests passing?

src/Instrumentation/Laravel/composer.json
```diff
@@ -15,21 +15,22 @@
     "open-telemetry/sem-conv": "^1"
   },
   "require-dev": {
+    "friendsofphp/php-cs-fixer": "^3",
+    "guzzlehttp/guzzle": "*",
+    "laravel/sail": "*",
     "laravel/sanctum": "*",
+    "laravel/tinker": "*",
+    "mockery/mockery": "2.0.x-dev",
```
ChrisLightfootWild, line 23: Added mockery.

src/Instrumentation/Laravel/composer.json
```diff
@@ -46,6 +47,7 @@
     }
   },
   "config": {
+    "sort-packages": true,
```
ChrisLightfootWild, line 50: Housekeeping :broom:

src/Instrumentation/Laravel/src/LaravelInstrumentation.php
```diff
@@ -4,8 +4,8 @@
 
 namespace OpenTelemetry\Contrib\Instrumentation\Laravel;
 
-use Illuminate\Foundation\Application;
-use Illuminate\Foundation\Http\Kernel;
+use Illuminate\Contracts\Foundation\Application;
```
ChrisLightfootWild, line 7: Each Watcher is already typed against `\Illuminate\Contracts\Foundation\Application`, so this maintains that type for consistency.

ChrisLightfootWild:
> Are all tests passing?

The existing ones do. I'm looking to write some which target the console but just need to figure out what that instrumentation should look like.

src/Instrumentation/Laravel/src/LaravelInstrumentation.php
```diff
@@ -92,20 +94,54 @@ public static function register(): void
...
         );
+
+        hook(
+            ConsoleKernel::class,
+            'call',
+            pre: static function (ConsoleKernel $kernel, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                /** @psalm-suppress ArgumentTypeCoercion */
+                $builder = $instrumentation->tracer()
+                    ->spanBuilder(sprintf('Console %s', $params[0] ?? 'unknown'))
+                    ->setSpanKind(SpanKind::KIND_PRODUCER)
```
ChrisLightfootWild, line 105: I'm not sure if this is correct or not?

src/Instrumentation/Laravel/composer.json
ChrisLightfootWild: Subsequently, removed as not needed for now.

src/Instrumentation/Laravel/src/LaravelInstrumentation.php
brettmc: Good question. Nothing in https://github.com/open-telemetry/opentelemetry-specification/blob/v1.24.0/specification/trace/api.md#spankind seems to quite fit for a console application - `producer` seems like a good choice.

brettmc:
This looks pretty good to me. There's a linting error to fix to make the build green. I also think it's worth waiting on #184 to be merged, and then ensuring your changes work back to Laravel 6.x

ChrisLightfootWild:
> This looks pretty good to me. There's a linting error to fix to make the build green. I also think it's worth waiting on #184 to be merged, and then ensuring your changes work back to Laravel 6.x

Thanks @brettmc. I think I've got the linting one sorted now.

This is #184, did you mean this should go in after another that's in queue? :smile:

Is there a reference/documented test strategy for this? Or do you tend to just go and manually create Laravel projects and test instrumentation that way? Happy to investigate this either way, all part of familiarising myself with the ecosystem.

Thanks!

brettmc:
Sorry, #185 (which has just been merged) is for Laravel too. Can you incorporate those changes into your branch. It looks like we can support back to Laravel v6.x

> Is there a reference/documented test strategy for this?

Just the existing integration tests, but verifying against a real Laravel app is also encouraged. If you do come across issues, please try to replicate them in a test. I'm very open to any improvements you can imagine :)

ChrisLightfootWild:
> Sorry, #185 (which has just been merged) is for Laravel too. Can you incorporate those changes into your branch. It looks like we can support back to Laravel v6.x
> 
> > Is there a reference/documented test strategy for this?
> 
> Just the existing integration tests, but verifying against a real Laravel app is also encouraged. If you do come across issues, please try to replicate them in a test. I'm very open to any improvements you can imagine :)

Thanks, I've merged main into this branch to reflect those changes. I think I fixed the import error on the last build too; IDE and php-cs-fixer in slight disagreement over what sorting the imports should look like!

ChrisLightfootWild:
I'm keeping this in draft still as I'm not happy with how it instruments the commands currently.

Will continue testing this through the coming evenings to better refine it.

src/Instrumentation/Laravel/src/HeadersPropagator.php
```diff
@@ -24,6 +24,7 @@ public static function instance(): self
     public function keys($carrier): array
     {
         assert($carrier instanceof Request);
+
```
ChrisLightfootWild, line 27: Linter complained without this.

ChrisLightfootWild:
I'm about to push my latest changes (which currently have a failing test); I am still trying to get the artisan handle to be recorded as the first span in the trace in the tests but falling over with it currently.

--

I've used the following test route to produce a trace in Zipkin:
```php
Route::get('/test', function () {
    \Illuminate\Support\Facades\Artisan::call('optimize');
    \Illuminate\Support\Facades\Artisan::call('optimize:clear');
});
```

image

ChrisLightfootWild:
@brettmc thanks for taking the time to debug this. Apologies, I didn't realise how the spans were recorded - so that massively helped in my understanding. Fixed in a8138ad43f7b8036da564de5b4a9a124028711bd.

brettmc (approved):

PR #190, opened 2023-09-04
improve readability of traces

iosifch:
Hi! I'm back with the proposed changes! As I said in the previous pull request, the changes should aim to a better readability of traces.

[previous pull request](https://github.com/opentelemetry-php/contrib-auto-symfony/pull/3)

### DIFF
--- a/src/Instrumentation/Symfony/src/HttpClientInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/HttpClientInstrumentation.php
@@ -36,7 +36,7 @@ final class HttpClientInstrumentation
                 /** @psalm-suppress ArgumentTypeCoercion */
                 $builder = $instrumentation
                     ->tracer()
-                    ->spanBuilder(\sprintf('HTTP %s', $params[0]))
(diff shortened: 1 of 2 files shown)


brettmc:
We discussed this at SIG just now. Somebody pointed out [this part of the spec](https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/http/#name) which I couldn't find earlier.

It's clear about how http server spans should be named, and also client spans. We're not quite getting either of them right at the moment, it seems :(

server spans: `{method} {optional route name}`
client spans: `{method}` (and MUST NOT use the URI).

Based on that this PR doesn't meet the spec requirements. However, we'd love it if this PR or a new one could bring http span names into alignment with the spec.

iosifch:
From what I understand the only issue with the pull request right now is that it adds the hostname for client request spans.

brettmc:
> From what you say I understand that the only issue with the pull request right now is that it adds the hostname for client request spans.

I think also that it adds "unknown route", which the spec says to leave out if not known.

PR #192, opened 2023-09-14
fixing broken symfony build

brettmc:


brettmc:
"fixing" might be generous, but the tests pass now.

pdelewski (approved):

PR #196, opened 2023-09-25
symfony auto-instrumentation readme

brettmc:
removing some content from symfony readme, and directing users to the central docs on opentelemetry.io for most info (since most READMEs are repetitive, and not as good as the central docs)

### DIFF
--- a/src/Instrumentation/Symfony/README.md
+++ b/src/Instrumentation/Symfony/README.md
@@ -38,3 +38,11 @@ From Symfony subdirectory:
 $ composer install
 $ ./vendor/bin/phpunit tests
 ```
+
+## Configuration
+
+The extension can be disabled via [runtime configuration](https://opentelemetry.io/docs/instrumentation/php/sdk/#configuration):
+
+```shell
+OTEL_PHP_DISABLED_INSTRUMENTATIONS=symfony
+```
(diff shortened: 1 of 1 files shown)


brettmc:
I thought I'd start with one readme, then follow up with the rest if we're happy with this approach.

src/Instrumentation/Symfony/README.md
```diff
@@ -49,14 +38,3 @@ From Symfony subdirectory:
...
 ```
-
-## Configuration
-
-Parts of this auto-instrumentation library can be configured, more options are available throught the
-[General SDK Configuration](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#general-sdk-configuration):
-
-| Name                                | Default value | Values                  | Example | Description                                                                     |
-|-------------------------------------|---------------|-------------------------|---------|---------------------------------------------------------------------------------|
-| OTEL_PHP_DISABLED_INSTRUMENTATIONS  | []            | Instrumentation name(s) | symfony | Disable one or more installed auto-instrumentations, names are comma seperated. |
```
ChrisLightfootWild, line 60: Is it worth keeping a reference to the instrumentation name somewhere in this document? Or adding it to the referenced instructions?

ChrisLightfootWild (approved):

bobstrecansky (approved):

src/Instrumentation/Symfony/README.md
brettmc: good idea, updated to mention config.

PR #204, opened 2023-10-19
Update `TraceAttributes` to semconv to 1.22.0

julianocosta89:
After PR https://github.com/open-telemetry/opentelemetry-php/pull/1126 got merged, `HTTP_FLAVOR` was dropped and all the  instrumentations that used it started failing.

This PR updates `TraceAttributes::HTTP_FLAVOR` to `TraceAttributes::NETWORK_PROTOCOL_VERSION`.


brettmc (approved):

brettmc:
Thanks, @julianocosta89 - I've just noticed that I only check for removed `ResourceAttributes` but not `TraceAttributes` as part of updating SemConv...
There are a few failures still, it looks like some packages are not picking up the latest SemConv version - they might need a more explicit version constraint. I think `^1.22` should do it.

julianocosta89:
Thanks @brettmc!

It seems that there are other attributes that need to be updated, I'll take care of that

julianocosta89:
Just one attribute that I couldn't find: `HTTP_CLIENT_IP`

I've replaced with `NET_SOCK_PEER_ADDR`, but not sure if that is correct.

julianocosta89:
@brettmc and @pdelewski I've updated a bunch of other TraceAttributes after your review, would you mind double checking it?
Also, I'm not sure about this: https://github.com/open-telemetry/opentelemetry-php-contrib/pull/204#issuecomment-1771065608

Appreciate the help!

cedricziel (approved):
lgtm!

brettmc (approved):

brettmc:
I've tagged new versions of the updated packages. Thanks, @julianocosta89

PR #206, opened 2023-10-19
update metapackage to stable

brettmc:


brettmc:
We still get ~200/day installations through this metapackage, and this should move those on to stable.

bobstrecansky (approved):

PR #209, opened 2023-11-10
Update SwooleContextHandler.php

Anoxia:
If the exception in the coroutine is not caught or other unknown circumstances, the code stack will run in a non-coroutine. At this time, Coroutine::getPcid($pcid) will return false, and SwooleContextHandler::isForked(int id) will raise a TypeError. fatal error.


brettmc (approved):

brettmc:
LGTM - thanks @Anoxia !
Since you are using this successfully, we'd really love a documentation patch (eg to the README) which describes how to set it up and use it. The original author didn't supply this :(

Anoxia:
@brettmc 
Okay, I will do it as soon as possible.

PR #213, opened 2023-11-23
Add ServerTiming response propagator

cedricziel:
This adds a response propagator for server-timings headers.

Server-Timing headers are especially useful as they are accessible for client side technology like RUM libraries. This is true even for the initial page-load. This makes server-timing the perfect vehicle to send the trace-id to the client for context-assumption, even without nasty hacks like meta html elements.

This approach has been demonstrated by vendors like Splunk and Instana. As there is currently no response propagation format agreed on in OpenTelemetry, the new package adds one as well. In a later iteration, these should optimally be backed by spec and then unified. For now, the interface is duplicated to avoid yet another package to be created.

Note to reviewers: The Symfony tests are failing, because the package referenced doesn't exist yet.


src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
```diff
@@ -117,6 +117,12 @@ public static function register(): void
 
                 $span->setAttribute(TraceAttributes::HTTP_RESPONSE_BODY_SIZE, $contentLength);
 
+                // Propagate server-timing header to response, if ServerTimingPropagator is present
```
brettmc, line 120: I think that if these two propagators globally register themselves (see B3 and cloudtrace in core repo), then they will be available via `Globals::propagator()` and can do all propagation in one hit. It should also then just work with any future propagators.

src/Propagation/ServerTiming/composer.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "open-telemetry/opentelemetry-propagation-server-timing",
```
brettmc, line 2: let's call it `open-telemetry/contrib-propagator-server-timing` (which is similar to existing propagator names, and assumes that eventually we'll move it to core and replace with `extension-propagator-server-timing`

src/Propagation/ServerTiming/src/ResponsePropagator.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace OpenTelemetry\Contrib\Propagation\ServerTiming;
+
+use OpenTelemetry\Context\ContextInterface;
+use OpenTelemetry\Context\Propagation\PropagationSetterInterface;
+
+/**
+ * This propagator type is used to inject the trace-context into HTTP responses.
+ */
+interface ResponsePropagator
```
brettmc, line 13: I think this is not required, and it should use core's `TextMapPropagatorInterface` instead

cedricziel:
@brettmc the `TextMapPropagator` and the way it is used currently for requests is not suited for ServerTiming propagation.

If we registered `ServerTimingPropagator` there and used `Globals::propagator()->inject()`, not just server-timings would be propagated _into the response_, but also the other configured propagation formats, which is not the desired behavior. Same way, if server-timings was a regularly activated propagator, it would be propagated forward, which is also not the desired behavior.

What I thought about, is to amend the core to offer a dedicated method for response propagators:

```php
// register response propagator in _register.php
\OpenTelemetry\SDK\Registry::registerTextMapResponsePropagator(
   'server-timing',
    ServerTimingPropagator::getInstance()
);

// get a composite propagator with _just_ response propagators
Globals::responsePropagator()->inject($carrier);
```

brettmc:
> What I thought about, is to amend the core to offer a dedicated method for response propagators

Yes, you're absolutely right. Are we confident enough in the idea of a response propagator to attempt that in core? I feel like it's something that should be spec-driven.

src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
brettmc: I'm wrong here, since response propagators are similar but applied differently, they will likely need their own interface in future.

brettmc (approved):
LGTM - I think a future TODO to formalize in core the idea of a response propagator, possibly based on a spec enhancement.

cedricziel:
> I feel like it's something that should be spec-driven.

💯 

@jpkrohling and I want to propose a spec amendment somewhere in the next 3 months

brettmc:
@cedricziel one thing missing is an entry in .gitsplit.yaml for the new package.
I've created https://github.com/opentelemetry-php/contrib-propagator-server-timing as the target repo

cedricziel:
@brettmc pushed!

brettmc (approved):

PR #214, opened 2023-11-23
Add traces attributes

weslenteche:
After the semantic convention was stabilized in [v1.23](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-spans.md#http-server), I added some missing attributes.
I noticed that some automatic instrumentation already had it, but other instrumentation was missing.


brettmc (approved):
LGTM - thanks!

PR #215, opened 2023-11-26
Record spans for symfony messenger

cedricziel:
This instrumentation adds spans for the Symfony messenger component.

It creates:
* an internal span when a message is sent to a MessageBusInterface
* an internal span, when a message is sent to a transport

Nota bene:
* enable all testsuites in PHP unit
* configure Github Actions to only run one job in parallel per PR and cancel existing run when new commits are pushed


cedricziel:
PSR-3 testsuite is broken..

src/Instrumentation/Symfony/src/MessengerInstrumentation.php
```diff
@@ -0,0 +1,175 @@
...
+                $builder = $instrumentation
+                    ->tracer()
+                    ->spanBuilder(\sprintf('DISPATCH %s', $messageClass))
+                    ->setSpanKind(SpanKind::KIND_INTERNAL)
+                    ->setAttribute(TraceAttributes::CODE_FUNCTION, $function)
+                    ->setAttribute(TraceAttributes::CODE_NAMESPACE, $class)
+                    ->setAttribute(TraceAttributes::CODE_FILEPATH, $filename)
+                    ->setAttribute(TraceAttributes::CODE_LINENO, $lineno)
+
+                    ->setAttribute(self::ATTRIBUTE_MESSENGER_BUS, $class)
```
brettmc, line 67: There are a bunch of messaging-related attributes in SemConv: https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#messaging-attributes - some might be appropriate here, eg `messaging.system`, `messaging.message.id` ?

src/Instrumentation/Symfony/src/MessengerInstrumentation.php
```diff
@@ -0,0 +1,175 @@
...
+            ) use ($instrumentation): array {
+                /** @var object|Envelope $message */
+                $message = $params[0];
+                $messageClass = \get_class($message);
+
+                /** @psalm-suppress ArgumentTypeCoercion */
+                $builder = $instrumentation
+                    ->tracer()
+                    ->spanBuilder(\sprintf('DISPATCH %s', $messageClass))
+                    ->setSpanKind(SpanKind::KIND_INTERNAL)
```
brettmc, line 61: It looks like the span kind should be `PRODUCER` here: https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-kind

src/Instrumentation/Symfony/src/MessengerInstrumentation.php
```diff
@@ -0,0 +1,175 @@
...
+                ?int $lineno,
+            ) use ($instrumentation): array {
+                /** @var object|Envelope $message */
+                $message = $params[0];
+                $messageClass = \get_class($message);
+
+                /** @psalm-suppress ArgumentTypeCoercion */
+                $builder = $instrumentation
+                    ->tracer()
+                    ->spanBuilder(\sprintf('DISPATCH %s', $messageClass))
```
brettmc, line 60: if possible, should be `<destination name> <operation name>`: https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-name

brettmc:
@cedricziel you can leave the psr3 failures to me, those tests haven't been running previously.

src/Instrumentation/Symfony/src/MessengerInstrumentation.php
cedricziel: I'm hesitant about the semconv related comments.

In Symfony Messenger, the chain of events is:

Dispatch to Bus -> Send To Transport -> Send to messaging system. 

The actual messaging system will be rabbit, Kafka or the like and there we would add producer kind and so on. I don't want to abuse the semconv as the pieces instrumented here are really internal bits; hence internal spans. The actual instrumentation for interaction with a queue or messaging and the context propagation should be done in a different instrumentation IMHO

.github/workflows/php.yml
```diff
@@ -129,7 +133,7 @@ jobs:
 
     - name: Run PHPUnit
       working-directory: src/${{ matrix.project }}
-      run: vendor/bin/phpunit --testsuite integration --testsuite unit --coverage-text --coverage-clover=coverage.clover
+      run: vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover
```
brettmc, line 136: I think the phpt failures are due to xdebug, can you try it with:
```suggestion
      run: XDEBUG_MODE=coverage vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover
```

brettmc:
The psr-3 failures are sort of xdebug-related. I was able to replicate this locally, and it occurs when running the .phpt tests with coverage enabled, whilst actually generating coverage. eg `XDEBUG_MODE=coverage vendor/bin/phpunit --coverage-text`
With this command, the output from the phpt test is different (it looks like exporting did not happen, or at least did not make it to console).

So, #217 is a compromise: we run *all* tests without generating coverage (which includes any phpt tests), and generate coverage only off the unit & integration suites.

The reason this failed to fail previously is that `--testsuite unit --testsuite integration` is not the right way to run two suites, one clobbers the other. You need to `--testsuite unit,integration`.

brettmc:
@cedricziel merging in main should make the build green.

cedricziel:
Done @brettmc ! Thanks for the thorough review! 🙏

brettmc (approved):

PR #218, opened 2023-11-28
Add ext-amqp instrumentation

cedricziel:
This new instrumentation will create spans for RabbitMQ interactions through [`ext-amqp`](https://github.com/php-amqp/php-amqp) and inject and extract context where possible.

- [x] Instrument `AMQPExchange::publish` - create Producer span && inject context
- [ ] ~~Instrument `AMQPQueue::consume` - create Consumer span && extract context~~
- [ ] ~~Instrument `AMQPQueue::get` - create Consumer span && extract context~~
- [x] Instrument `AMQPQueue::ack` - create client span
- [x] Instrument `AMQPQueue::nack` - create client span
- [x] Instrument `AMQPQueue::reject` - create client span


brettmc (commented):
Looking good so far. To run tests in GH actions, you'll probably need to spin up a rabbitmq instance. We do something similar with mongodb, so maybe some inspiration there.

src/Instrumentation/ExtAmqp/tests/Integration/ExtAmqpInstrumentationTest.php
```diff
@@ -0,0 +1,83 @@
...
+use PHPUnit\Framework\TestCase;
+
+class ExtAmqpInstrumentationTest extends TestCase
+{
+    private ScopeInterface $scope;
+    /**
+     * @var ImmutableSpan[]
+     */
+    private ArrayObject $storage;
+    private ImmutableSpan $span;
```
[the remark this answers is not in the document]
cedricziel: No, it's c&p

cedricziel:
I think at this point, I have 2 questions:
* why is the instrumentation not active on 8.1 and 8.2?
* how can i start a new root span and replace any current active span in the current process?

To my 2nd point, my thought process is as follows:
* consumer continously `get`s new messages from the queue, which is captured as span
* consumer processes would do processing on messages as siblings to `get`ing messages, so processing will naturally not be a child
* if we wanted to capture processing of messages as child spans of consumer spans, we probably need to replace the active root span to keep semantics intact

Please check me here :) 

Another way would be to not instrument `AMQPQueue::get` directly, but to instrument any messaging framework (such as symfony/messenger) and the AMQP driver there, but there would be quite similar questions.

cedricziel:
> the observer API can only observe internal functions (including extensions) from PHP8.2+, and that's why it fails for 8.0+8.1 (PDO is the same)

d'oh! - how could I miss that? :) Thanks Brett

> edit: [cloud-native.slack.com/archives/CJFCJHG4Q/p1698833937912889](https://cloud-native.slack.com/archives/CJFCJHG4Q/p1698833937912889) touches on this subject

I think the gist of that thread is that we should provide documentation on how to do this in general in a worker/consumer scenario and then find a way to encapsulate this / do it automatically when looking at frameworks

src/Instrumentation/ExtAmqp/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Example
+
+```php
+//Create and declare channel
+$channel = new AMQPChannel($connection);
+
+$routing_key = 'task_queue';
+
+$callback_func = function(AMQPEnvelope $message, AMQPQueue $q) use (&$max_jobs) {
```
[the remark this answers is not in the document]
cedricziel: copy & paste - removed

src/Instrumentation/ExtAmqp/src/ExtAmqpInstrumentation.php
```diff
@@ -0,0 +1,225 @@
...
+                    ->tracer()
+                    ->spanBuilder($routingKey . ' publish')
+                    ->setSpanKind(SpanKind::KIND_PRODUCER)
+                    // code
+                    ->setAttribute(TraceAttributes::CODE_FUNCTION, $function)
+                    ->setAttribute(TraceAttributes::CODE_NAMESPACE, $class)
+                    ->setAttribute(TraceAttributes::CODE_FILEPATH, $filename)
+                    ->setAttribute(TraceAttributes::CODE_LINENO, $lineno)
+                    // messaging
+                    ->setAttribute(TraceAttributes::MESSAGING_SYSTEM, 'rabbitmq')
```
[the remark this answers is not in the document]
cedricziel: technically, it will work with any amqp broker. good catch. Let me see if I can use some of the available context to find out.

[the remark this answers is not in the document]
cedricziel: I think we can't know the broker type from ext-amqp - it could be anything, really. I propose we switch from `rabbitmq` to `amqp` on this required attribute

src/Instrumentation/ExtAmqp/tests/Integration/ExtAmqpInstrumentationTest.php
```diff
@@ -0,0 +1,151 @@
...
+            $this->assertEquals($routing_key . ' publish', $span->getName());
+            $this->assertEquals('amqp', $span->getAttributes()->get(TraceAttributes::MESSAGING_SYSTEM));
+            $this->assertEquals(SpanKind::KIND_PRODUCER, $span->getKind());
+
+            /**
+             * Our message should be the first one in the queue
+             */
+            $envelope = $queue->get();
+
+            self::assertFalse($envelope->hasHeader('traceparent'), 'traceparent header is present');
```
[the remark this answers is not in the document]
cedricziel: So the way I'm used to using PHPUnit messages is: describe the condition that is unexpected. So in this case, if the test fails, it spells out what is unexpected.

 Should I switch both to the other way around?

src/Instrumentation/ExtAmqp/src/ExtAmqpInstrumentation.php
```diff
@@ -0,0 +1,225 @@
...
+use OpenTelemetry\SemConv\TraceAttributes;
+use Throwable;
+
+final class ExtAmqpInstrumentation
+{
+    public const NAME = 'ext_amqp';
+
+    public static function register(): void
+    {
+        $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.ext_amqp');
```
Nevay, line 26: Should add version and schema url.

src/Instrumentation/ExtAmqp/src/ExtAmqpInstrumentation.php
```diff
@@ -0,0 +1,225 @@
...
+                string $function,
+                ?string $filename,
+                ?int $lineno,
+            ) use ($instrumentation): array {
+                $routingKey = $params[1];
+
+                /** @psalm-suppress ArgumentTypeCoercion */
+                $builder = $instrumentation
+                    ->tracer()
+                    ->spanBuilder($routingKey . ' publish')
```
Nevay, line 44: Should include exchange name.

src/Instrumentation/ExtAmqp/src/ExtAmqpInstrumentation.php
```diff
@@ -0,0 +1,225 @@
...
+                    ->setAttribute(TraceAttributes::CODE_LINENO, $lineno)
+                    // messaging
+                    ->setAttribute(TraceAttributes::MESSAGING_SYSTEM, 'amqp')
+                    ->setAttribute(TraceAttributes::MESSAGING_OPERATION, 'publish')
+
+                    ->setAttribute(TraceAttributes::MESSAGING_DESTINATION, $routingKey)
+                    ->setAttribute(TraceAttributes::MESSAGING_DESTINATION_NAME, $routingKey)
+                    ->setAttribute(TraceAttributes::MESSAGING_DESTINATION_PUBLISH_NAME, $routingKey)
+
+                    ->setAttribute(TraceAttributes::MESSAGING_DESTINATION_KIND, 'queue')
```
Nevay, line 59: Do we want to populate deprecated / removed attributes? If yes, should be `topic`.

src/Instrumentation/ExtAmqp/src/ExtAmqpInstrumentation.php
```diff
@@ -0,0 +1,225 @@
...
+                    ->setAttribute(TraceAttributes::CODE_NAMESPACE, $class)
+                    ->setAttribute(TraceAttributes::CODE_FILEPATH, $filename)
+                    ->setAttribute(TraceAttributes::CODE_LINENO, $lineno)
+                    // messaging
+                    ->setAttribute(TraceAttributes::MESSAGING_SYSTEM, 'amqp')
+                    ->setAttribute(TraceAttributes::MESSAGING_OPERATION, 'publish')
+
+                    ->setAttribute(TraceAttributes::MESSAGING_DESTINATION, $routingKey)
+                    ->setAttribute(TraceAttributes::MESSAGING_DESTINATION_NAME, $routingKey)
+                    ->setAttribute(TraceAttributes::MESSAGING_DESTINATION_PUBLISH_NAME, $routingKey)
```
Nevay, line 57: Must use exchange name, [spec](https://github.com/open-telemetry/semantic-conventions/blob/40839cbd53939126b213f2b50921b92a107719e0/docs/messaging/rabbitmq.md#rabbitmq-attributes)
> In RabbitMQ, the destination is defined by an exchange and a routing key. messaging.destination.name MUST be set to the name of the exchange. This will be an empty string if the default exchange is used.

Nevay (changes requested):

src/Instrumentation/ExtAmqp/src/ExtAmqpInstrumentation.php
cedricziel: Done, thanks

src/Instrumentation/ExtAmqp/src/ExtAmqpInstrumentation.php
cedricziel: Done.

src/Instrumentation/ExtAmqp/src/ExtAmqpInstrumentation.php
cedricziel: Done

src/Instrumentation/ExtAmqp/src/ExtAmqpInstrumentation.php
cedricziel: I've set the version by reflecting through the composer runtime API, there is no schema URL to my knowledge

cedricziel:
I think this is good to go :)

src/Instrumentation/ExtAmqp/tests/Integration/ExtAmqpInstrumentationTest.php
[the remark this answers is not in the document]
brettmc: > describe the condition that is unexpected

I think it's more of an art than a science, and from skimming the phpunit docs I didn't find any guidance. I've always thought of the message as a human-readable form of the expectation, but there are plenty of examples of the message describing the failure (which makes sense too, since it's only displayed in case of failure).
Let's leave it as-is :)

src/Instrumentation/ExtAmqp/src/ExtAmqpInstrumentation.php
brettmc: I think we don't do this anywhere, but should: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.27.0/specification/schemas/README.md

> Third-party libraries, instrumentation or applications are advised to define and publish their own telemetry schema if it is completely different from OpenTelemetry schema (or use OpenTelemetry schema) and include the schema URL in the emitted telemetry.

Linked from there is an example of how go does it, and they use the semconv schema url, so `ResourceAttribute::SCHEMA_URL` seems like the right value to use.

cedricziel: Okay, pushed

brettmc (approved):

brettmc:
https://packagist.org/packages/open-telemetry/opentelemetry-auto-ext-amqp

PR #219, opened 2023-11-29
Rename the span name of Laravel HTTP requests to include the low cardinality route name if possible.

LauJosefsen:
This renames the span name created around a request in Laravel to include the low cardinality route name if possible.

For example, the route registered as `Route::get('/user/{userId}` when requested with `GET /user/123` would previously result in a span.name of `GET`, where it will now show as `GET /user/{userId}`.

This is to follow the spec presented in the OpenTelemetry docs: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#:~:text=HTTP%20server%20span%20names%20SHOULD%20be%20%7Bmethod%7D%20%7Bhttp.route%7D%20if%20there%20is%20a%20(low%2Dcardinality)%20http.route%20available%20(see%20below%20for%20the%20exact%20definition%20of%20the%20%7Bmethod%7D%20placeholder).

### DIFF
--- a/src/Instrumentation/Laravel/src/HttpInstrumentation.php
+++ b/src/Instrumentation/Laravel/src/HttpInstrumentation.php
@@ -87,6 +87,7 @@ class HttpInstrumentation
                         $route = '/' . $route;
                     }
 
+                    /** @psalm-suppress ArgumentTypeCoercion */
                     $span->updateName(sprintf('%s %s', $request?->method() ?? 'unknown', $route));
                 }
 
--- a/src/Instrumentation/Laravel/src/HttpInstrumentation.php
+++ b/src/Instrumentation/Laravel/src/HttpInstrumentation.php
@@ -6,6 +6,7 @@ namespace OpenTelemetry\Contrib\Instrumentation\Laravel;
 
 use Illuminate\Contracts\Http\Kernel;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Route;
 use OpenTelemetry\API\Globals;
 use OpenTelemetry\API\Instrumentation\CachedInstrumentation;
 use OpenTelemetry\API\Trace\Span;
@@ -78,6 +79,10 @@ class HttpInstrumentation
                     $span->setAttribute(TraceAttributes::HTTP_RESPONSE_STATUS_CODE, $response->getStatusCode());
                     $span->setAttribute(TraceAttributes::NETWORK_PROTOCOL_VERSION, $response->getProtocolVersion());
                     $span->setAttribute(TraceAttributes::HTTP_RESPONSE_BODY_SIZE, $response->headers->get('Content-Length'));
+                }
+                if(($route = Route::getCurrentRoute()?->uri()) !== null) {
+                    $request = ($params[0] instanceof Request) ? $params[0] : null;
+                    $span->updateName(sprintf('%s %s', $request?->method() ?? 'unknown', $route));
                 }
 
                 $span->end();
--- a/src/Instrumentation/Laravel/src/HttpInstrumentation.php
+++ b/src/Instrumentation/Laravel/src/HttpInstrumentation.php
@@ -82,6 +82,11 @@ class HttpInstrumentation
                 }
                 if(($route = Route::getCurrentRoute()?->uri()) !== null) {
                     $request = ($params[0] instanceof Request) ? $params[0] : null;
+
+                    if (! str_starts_with($route, '/')) {
+                        $route = '/' . $route;
+                    }
+
                     $span->updateName(sprintf('%s %s', $request?->method() ?? 'unknown', $route));
                 }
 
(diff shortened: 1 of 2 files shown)


src/Instrumentation/Laravel/src/HttpInstrumentation.php
```diff
@@ -79,6 +80,15 @@ public static function register(CachedInstrumentation $instrumentation): void
                     $span->setAttribute(TraceAttributes::NETWORK_PROTOCOL_VERSION, $response->getProtocolVersion());
                     $span->setAttribute(TraceAttributes::HTTP_RESPONSE_BODY_SIZE, $response->headers->get('Content-Length'));
                 }
+                if(($route = Route::getCurrentRoute()?->uri()) !== null) {
+                    $request = ($params[0] instanceof Request) ? $params[0] : null;
+
+                    if (! str_starts_with($route, '/')) {
```
brettmc, line 86: what happens here if it's a [named route](https://laravel.com/docs/10.x/routing#named-routes) rather than a uri pattern?

brettmc: Sorry, what I was trying to say was "should the route name be used as the span name if one is defined for a route" ? https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name doesn't go down to that level of detail, and I don't think we're consistent between various frameworks' auto-instrumentation, so I don't have a strong opinion here.

LauJosefsen: Ah, I understand now.

That is actually a good question, which I also don't have a strong opinion of.

One argument to use URI instead of route name would be that it always exists for registered routes, and thus more consistent.

brettmc: no problem, let's leave it as is for now :)

brettmc:
You can fix the psalm error with a `@psalm-suppress ArgumentTypeCoercion`

src/Instrumentation/Laravel/src/HttpInstrumentation.php
```diff
@@ -79,6 +80,15 @@ public static function register(CachedInstrumentation $instrumentation): void
                     $span->setAttribute(TraceAttributes::NETWORK_PROTOCOL_VERSION, $response->getProtocolVersion());
                     $span->setAttribute(TraceAttributes::HTTP_RESPONSE_BODY_SIZE, $response->headers->get('Content-Length'));
                 }
+                if(($route = Route::getCurrentRoute()?->uri()) !== null) {
```
ChrisLightfootWild, line 83: Can this be achieved in the pre hook [here](https://github.com/open-telemetry/opentelemetry-php-contrib/blob/720adadba6e8081a301a857c93ef7f1dc56a81b7/src/Instrumentation/Laravel/src/HttpInstrumentation.php#L32) instead?

The `$request` object is already in scope and should be able to provide the URI there.

Or `$request->route()` for the named route?

ChrisLightfootWild: @LauJosefsen ahh, that might make sense then.

Failing that, you could pull a request object out of the params in the post-hook in the same way as the pre-hook:
```php
$request = ($params[0] instanceof Request) ? $params[0] : null;
```
(anything to avoid the facade usage! 👀 )

LauJosefsen: I can confirm the route is unavailable in the pre hook, unless we call the router manually, which I think is a bad idea.

`$request->route` can be used instead of the facade. The return type is not very nice though, being `Route|null|object|string` . I can do an implementation that sets it to the string returned if string, the route uri if Route, and ignore it if it is a non-Route-object or null?  

Something like this:

```php
if(($route = self::getRouteName($request)) !== null){
    $span->updateName(sprintf('%s %s', $request?->method() ?? 'unknown', $route));
}

private static function getRouteName(Request $request): string {
    $route = $request->route();
    if($route instanceof Route){
        return $route->uri();
    }
    if(is_string($route)){
        return $route;
    }
    return null;
}
```

ChrisLightfootWild: Agreed, let's stick to the post-hook :+1: 

You can end up with something like this and avoid facade usage:

```php
$request = ($params[0] instanceof Request) ? $params[0] : null;
if ($request && $routeName = $request->route()?->uri() ?: $request->route() ?: null) {
    $span->updateName(sprintf('%s %s', $request?->method() ?? 'unknown', '/'. ltrim($routeName, '/')));
}
```

(disclaimer: facade avoidance is a personal preference :sweat_smile: ) 

@LauJosefsen I'm looking forward to seeing this land, thanks for looking into it!

brettmc (approved):

brettmc:
released in 0.0.18

PR #220, opened 2023-12-05
Laravel: refactor the Request facade usage.

ChrisLightfootWild:
Hey @LauJosefsen :wave: 

I've built on what you added in your PR to remove the facade. I believe the outcome is the same. Please can you take a look to make sure you are happy with the proposed changes?

Thanks!

### DIFF
--- a/src/Instrumentation/Laravel/src/HttpInstrumentation.php
+++ b/src/Instrumentation/Laravel/src/HttpInstrumentation.php
@@ -6,7 +6,6 @@ namespace OpenTelemetry\Contrib\Instrumentation\Laravel;
 
 use Illuminate\Contracts\Http\Kernel;
 use Illuminate\Http\Request;
-use Illuminate\Support\Facades\Route;
 use OpenTelemetry\API\Globals;
 use OpenTelemetry\API\Instrumentation\CachedInstrumentation;
 use OpenTelemetry\API\Trace\Span;
@@ -80,15 +79,11 @@ class HttpInstrumentation
                     $span->setAttribute(TraceAttributes::NETWORK_PROTOCOL_VERSION, $response->getProtocolVersion());
                     $span->setAttribute(TraceAttributes::HTTP_RESPONSE_BODY_SIZE, $response->headers->get('Content-Length'));
                 }
-                if(($route = Route::getCurrentRoute()?->uri()) !== null) {
-                    $request = ($params[0] instanceof Request) ? $params[0] : null;
 
-                    if (! str_starts_with($route, '/')) {
-                        $route = '/' . $route;
-                    }
-
+                $request = ($params[0] instanceof Request) ? $params[0] : null;
+                if (($route = $request?->route()?->uri()) !== null) {
                     /** @psalm-suppress ArgumentTypeCoercion */
-                    $span->updateName(sprintf('%s %s', $request?->method() ?? 'unknown', $route));
+                    $span->updateName(sprintf('%s %s', $request?->method() ?? 'unknown', '/' . ltrim($route, '/')));
                 }
 
                 $span->end();
--- a/src/Instrumentation/Laravel/src/HttpInstrumentation.php
+++ b/src/Instrumentation/Laravel/src/HttpInstrumentation.php
@@ -80,13 +80,6 @@ class HttpInstrumentation
                     $span->setAttribute(TraceAttributes::HTTP_RESPONSE_BODY_SIZE, $response->headers->get('Content-Length'));
                 }
 
-                $request = ($params[0] instanceof Request) ? $params[0] : null;
-                if ($request && ($route = call_user_func($request->getRouteResolver()))) {
-                    /** @var \Illuminate\Routing\Route|null $route */
-                    /** @psalm-suppress ArgumentTypeCoercion */
-                    $span->updateName(sprintf('%s %s', $request->method() ?: 'unknown', '/' . ltrim($route->uri(), '/')));
-                }
-
                 $span->end();
             }
         );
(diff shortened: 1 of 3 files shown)


cedricziel (approved):

src/Instrumentation/Laravel/src/HttpInstrumentation.php
```diff
@@ -80,15 +79,11 @@ public static function register(CachedInstrumentation $instrumentation): void
...
                 }
-                if(($route = Route::getCurrentRoute()?->uri()) !== null) {
-                    $request = ($params[0] instanceof Request) ? $params[0] : null;
-
-                    if (! str_starts_with($route, '/')) {
-                        $route = '/' . $route;
-                    }
 
+                $request = ($params[0] instanceof Request) ? $params[0] : null;
+                if (($route = $request?->route()?->uri()) !== null) {
```
cedricziel, line 84:
```
Error: src/HttpInstrumentation.php:84:52: PossiblyInvalidMethodCall: Cannot call method on possible string variable $__tmp_nullsafe__4512 (see https://psalm.dev/113)
```

cedricziel: https://laravel.com/api/6.x/Illuminate/Http/Request.html#method_route

So `Request::route()` returns `Route|object|string|null`. The implementation here assumes it's always a route. Can you account for object & string as well?

cedricziel (changes requested):

src/Instrumentation/Laravel/src/HttpInstrumentation.php
LauJosefsen: I think a viable solution could be as i proposed here https://github.com/open-telemetry/opentelemetry-php-contrib/pull/219/files/41e7dac805c26e2ea179a6145697545f87d5acc7#r1411897050

ChrisLightfootWild: I've stepped through this now. Once the `Illuminate\Routing\Router` dispatches the request, the route resolver is set as a closure which will only ever return a `Route` object.

This feels cleaner than using `$request->route()` personally, which takes additional parameters and code-paths which the tooling flags as having multiple return types.

nizacegodk:
Besides the specifics of the different syntax.

Are there any reasons for this change? Do facades pose a problem for the integration that are not obvious for other contributors to OpenTelemetry?

A facade in Laravel resolves a key from the ServiceContainer and calls the given method. It is syntactic suger for calling a bound object in the ServiceContainer statically.

Meaning `$kernel->getApplication()->get(\Illuminate\Routing\Router::class)->getCurrentRoute()` is equivalent with `Route::getCurrentRoute()`.

If there are any problems with using facades, could that problem be documented so new contribution does not make the same mistake?

ChrisLightfootWild:
> Besides the specifics of the different syntax.
> 
> Are there any reasons for this change? Do facades pose a problem for the integration that are not obvious for other contributors to OpenTelemetry?
> 
> A facade in Laravel resolves a key from the ServiceContainer and calls the given method. It is syntactic suger for calling a bound object in the ServiceContainer statically.
> 
> Meaning `$kernel->getApplication()->get(\Illuminate\Routing\Router::class)->getCurrentRoute()` is equivalent with `Route::getCurrentRoute()`.
> 
> If there are any problems with using facades, could that problem be documented so new contribution does not make the same mistake?

In my experience, code-editors rely quite heavily on PHPDoc to make facades even semi-useful.
Perhaps it is personal preference, but I prefer the use of DI where possible over the usage of a facade.

In cases such as this, I think using the facade just masks how we can otherwise access an object we can access in the current scope.

Maybe this can be discussed a little further in a SIG to gather some collective thoughts.

ChrisLightfootWild:
Hey @nizacegodk 👋

A discussion thread was opened in Slack to discuss this issue further: https://cloud-native.slack.com/archives/C01NFPCV44V/p1701869446274709?thread_ts=1701869446.274709&cid=C01NFPCV44V

Hopefully we can get some input to clarify the above 🤞

Nevay:
I am not familiar with laravel, but the change looks good to me. It seems less fragile if laravel supports subrequests.

An alternative approach would be to use a [watcher](https://github.com/open-telemetry/opentelemetry-php-contrib/tree/cb436eca5a7be8fb350ddbd4ecfc9e5287b740cf/src/Instrumentation/Laravel/src/Watchers) that listens to the `RouteMatched` event (untested):
```php
$app['events']->listen(RouteMatched::class, static function(RouteMatched $event): void {
    /** @var SpanInterface|null $span */
    if (!$span = $event->request->attributes->get(SpanInterface::class)) {
        return;
    }

    $span->updateName(sprintf('%s %s', $event->request->getMethod(), $event->route->uri));
    $span->setAttribute(TraceAttributes::HTTP_ROUTE, $event->route->uri);
});
```

ChrisLightfootWild:
> I am not familiar with laravel, but the change looks good to me. It seems less fragile if laravel supports subrequests.
> 
> An alternative approach would be to use a [watcher](https://github.com/open-telemetry/opentelemetry-php-contrib/tree/cb436eca5a7be8fb350ddbd4ecfc9e5287b740cf/src/Instrumentation/Laravel/src/Watchers) that listens to the `RouteMatched` event (untested):
> 
> ```
> $app['events']->listen(RouteMatched::class, static function(RouteMatched $event): void {
>     /** @var SpanInterface|null $span */
>     if (!$span = $event->request->attributes->get(SpanInterface::class)) {
>         return;
>     }
> 
>     $span->updateName(sprintf('%s %s', $event->request->getMethod(), $event->route->uri));
>     $span->setAttribute(TraceAttributes::HTTP_ROUTE, $event->route->uri);
> });
> ```

Thanks for taking a look at this @Nevay - and apologies for the very delayed response.

Your suggestion looked feasible, I've updated the PR with those changes.

brettmc (approved):

brettmc:
@ChrisLightfootWild one psalm issue to fix

ChrisLightfootWild:
> @ChrisLightfootWild one psalm issue to fix

Thanks! I've fixed it now on the back of `all-checks` :sweat_smile:

PR #230, opened 2024-01-27
TraceResponse and ServerTiming response propagation for CodeIgniter, Yii, Laravel

agoallikmaa:
There are 2 response propagators which include a W3C TraceParent formatted tracing context info: TraceResponse and ServerTiming. The latter is necessary for RUM integration since frontend JS can read the value from ServerTiming, and the former is the more standardized method, but unsuitable for RUM until browsers add it to the list of headers that can be read from JS.

Previous state:
* Symfony already had both
* CodeIgniter and Yii only had `TraceResponse`
* Laravel had neither

Added the missing ones to make sure both are present for Symfony, Laravel, CodeIgniter and Yii. PSR-15 and consequently Slim have a method for setting headers that isn't trivially compatible with `PropagationSetterInterface` (as response is modified by `$response = $response->withX(...)` pattern), so it is not included here.

As integration tests are not applicable for testing that the headers actually get set, tested this manually for each of them.

The code duplication for each framework for this purpose is not ideal, so at some point we should think of ways to fix that, but it is not in the scope for this PR.

### DIFF
--- a/src/Instrumentation/CodeIgniter/src/CodeIgniterInstrumentation.php
+++ b/src/Instrumentation/CodeIgniter/src/CodeIgniterInstrumentation.php
@@ -116,6 +116,14 @@ class CodeIgniterInstrumentation
                         $span->setStatus(StatusCode::STATUS_ERROR);
                     }
 
+                    // Propagate server-timing header to response, if ServerTimingPropagator is present
+                    if (class_exists('OpenTelemetry\Contrib\Propagation\ServerTiming\ServerTimingPropagator')) {
+                        /** @phan-suppress-next-line PhanUndeclaredClassMethod */
+                        $prop = new \OpenTelemetry\Contrib\Propagation\ServerTiming\ServerTimingPropagator();
+                        /** @phan-suppress-next-line PhanUndeclaredClassMethod */
+                        $prop->inject($response, ResponsePropagationSetter::instance(), $scope->context());
+                    }
+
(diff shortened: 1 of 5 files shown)


src/Instrumentation/CodeIgniter/src/CodeIgniterInstrumentation.php
```diff
@@ -116,6 +116,14 @@ public static function register(): void
                         $span->setStatus(StatusCode::STATUS_ERROR);
                     }
 
+                    // Propagate server-timing header to response, if ServerTimingPropagator is present
+                    if (class_exists('OpenTelemetry\Contrib\Propagation\ServerTiming\ServerTimingPropagator')) {
```
brettmc, line 120: I think this and subsequent FQN strings would be safer if it was a `use` statement with `::class`. I don't think that should cause any issues (except maybe with static analysis).

agoallikmaa: Maybe that would make more sense once this has been extracted to some common module? Current approach is more copy-paste friendly, and automatic tests do not cover this if the class cannot be found due to a missing `use`. If one were to copy this to a new instrumentation and forget the `use`, the only way they find out that it doesn't work is with manual testing with the propagator included as a dependency and seeing that the header is missing.

brettmc (approved):

PR #235, opened 2024-02-11
Exclude ApiPlatform test HttpClient from instrumentation

cedricziel:
The client doesn't support the `on_progress` option and makes tests fail that use the ApiPlatform test client

Longer term we should probably maintain a blacklist or level up this instrumentation significantly so we know the allowed options

### DIFF
--- a/src/Instrumentation/Symfony/src/HttpClientInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/HttpClientInstrumentation.php
@@ -23,7 +23,7 @@ final class HttpClientInstrumentation
      */
     const NON_SUPPORTED_CLIENTS = [
         /** @psalm-suppress UndefinedClass */
-        \ApiPlatform\Symfony\Bundle\Test\Client::class,
+        '\ApiPlatform\Symfony\Bundle\Test\Client',
     ];
 
     public static function isSupportedImplementation(string $class): bool
--- a/src/Instrumentation/Symfony/src/HttpClientInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/HttpClientInstrumentation.php
@@ -18,6 +18,20 @@ use Symfony\Contracts\HttpClient\ResponseInterface;
 
 final class HttpClientInstrumentation
 {
+    /**
+     * These clients are not supported by this instrumentation.
+     *
+     * @psalm-suppress UndefinedClass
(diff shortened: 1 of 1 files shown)


agoallikmaa (approved):

brettmc:
If the pre hook doesn't start+activate a span, then the post hook could potentially deactivate some other scope, couldn't it? If so, might it be safer to create, activate and then end a span here so that the post hook works as expected? (I don't have access to a pc to actually test this, though)

src/Instrumentation/Symfony/src/HttpClientInstrumentation.php
```diff
@@ -33,6 +33,11 @@ public static function register(): void
                 ?string $filename,
                 ?int $lineno,
             ) use ($instrumentation): array {
+                /** @psalm-suppress UndefinedClass */
+                if (class_exists("\ApiPlatform\Symfony\Bundle\Test\Client") && $client instanceof \ApiPlatform\Symfony\Bundle\Test\Client) {
```
Nevay, line 37:
```suggestion
                if ($client instanceof \ApiPlatform\Symfony\Bundle\Test\Client) {
```

`class_exists()` triggers an unnecessary autoload attempt.

Condition has to be checked in `post` hook too, otherwise we might detach the parent scope / end the parent span.

brettmc (changes requested):
I think the `pre` hook should be unchanged, and the active span closed in the `post` callback if the client doesn't support on_progress. This way, a span is still created and activated in `pre` hook, and `post` hook cannot detach the wrong scope (which I'm pretty sure it will do, with this change).

bobstrecansky:
@cedricziel - are you still planning on working on this PR?

nesl247:
We also just ran into this issue, but also ran into when using prophecy to mock the `HttpClientInterface`.

cedricziel:
The problem is that Symfony will evaluate the $options array and will fail if the client doesn't support one of the provided options.

I'm adapting the instrumentation so that if will create a span, but need to close it manually in post hook as the callback method won't work.

cedricziel:
I think the AMQP error is unrelated...

cedricziel:
rebased. Unrelated style issues

brettmc (approved):

PR #241, opened 2024-02-26
Extract response headers as attributes for Guzzle, HTTPlug Async

agoallikmaa:
Extracting response headers as span attributes based on an INI option was supported on PSR-18, but not on Guzzle and HTTPlug Async. Both of these use PSR HTTP interfaces, so the code for these is identical to that of PSR-18.


bobstrecansky (commented):
Let's discuss this in the SIG meeting Wednesday - I can see the benefits and drawbacks of this

brettmc (approved):

PR #242, opened 2024-02-26
Avoid supplying integer to function expecting string if status code is not in array.

LauJosefsen:
This fixes https://github.com/open-telemetry/opentelemetry-php/issues/1238


LauJosefsen:
Also: Should type errors like this not be checked by CI?

weslenteche (approved):

brettmc:
> Also: Should type errors like this not be checked by CI? I can see PHPStan lvl 5 is running, this should catch typeerrors like this?

Great question! I just tried this out, phpstan doesn't care (even against level 9), nor does phan. However, psalm at level 1 does complain about this issue (along with a bunch of other things, which we could possibly improve).

brettmc (approved):

PR #244, opened 2024-03-05
Psr 16 auto

weslenteche:


src/Instrumentation/Psr16/tests/Integration/Psr16InstrumentationTest.php
```diff
@@ -0,0 +1,117 @@
...
+
+    public function tearDown(): void
+    {
+        $this->scope->detach();
+    }
+
+    public function test_get_key(): void
+    {
+        $this->assertCount(0, $this->storage);
+        $driver = $this->createMemoryCacheDriver();
```
brettmc, line 45: this could be moved into `setUp`

src/Instrumentation/Psr16/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+# OpenTelemetry PSR-16 auto-instrumentation
+Please read https://opentelemetry.io/docs/instrumentation/php/automatic/ for instructions on how to
+install and configure the extension and SDK.
+
+## Overview
+Auto-instrumentation hooks are registered via composer.
+
+* create spans automatically for each PSR-16 request that is sent (sync or async)
+* add a `traceparent` header to the request to facilitate distributed tracing
```
brettmc, line 18: copy/paste, not true for this package.

src/Instrumentation/Psr16/src/Psr16Instrumentation.php
```diff
@@ -0,0 +1,180 @@
...
+        CachedInstrumentation $instrumentation,
+        string $function,
+        string $class,
+        ?string $filename,
+        ?int $lineno
+    ): SpanBuilderInterface
+    {
+        return $instrumentation->tracer()
+            ->spanBuilder(sprintf('%s::%s', $class, $function))
+            ->setSpanKind(SpanKind::KIND_CLIENT)
```
cedricziel, line 156:
```suggestion
            ->setSpanKind(SpanKind::KIND_INTERNAL)
```

A cache interaction usually involves outgoing calls to redis, memcached or another implementation. If these had instrumentation active, we'd create CLIENT spans with CLIENT children, which is semantically incorrect. I propose we either detect if other instrumentation is active and make the kind contingent on that, or make these spans internal per se

cedricziel (changes requested):

src/Instrumentation/Psr16/src/Psr16Instrumentation.php
```diff
@@ -0,0 +1,180 @@
...
+    public const NAME = 'psr16';
+
+    public static function register(): void
+    {
+        $instrumentation = new CachedInstrumentation(
+            'io.opentelemetry.contrib.php.psr16',
+            InstalledVersions::getVersion('open-telemetry/opentelemetry-auto-psr16')
+        );
+
+        hook(
```
agoallikmaa, line 33: There don't seem to be any differences between these 8 hooks besides the method name. Perhaps you can extract it into a function that takes `$instrumentation` and method name as parameters?

src/Instrumentation/Psr16/src/Psr16Instrumentation.php
weslenteche: When I was making the changes discussed, I ended up having some questions:
- Would making all spans internal be violating the semantic convention of maintaining attributes (`db.operation`, `db.system`)?
- In the image below I used [this in-memory cache file](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/244/files#diff-ed82a484ae6e7f8cbae2e02d761ebe71f99ba9791404318f23b52c1bf3f0b2bbR103) to perform some tests, so I had a question that is related to your comment when continuing to create spans as a Client, in cases such as the deleteMultiple operation that can generate child spans, these must be assigned like Internal ? Could you get an example of how to detect an active Client interval?

traces

weslenteche: @cedricziel could you take a look at these questions?

src/Instrumentation/Psr16/tests/Integration/Psr16InstrumentationTest.php
```diff
@@ -0,0 +1,161 @@
...
+        $span = $this->storage->offsetGet(0);
+        $this->assertEquals('get', $span->getName());
+        $this->assertEquals('psr16', $span->getAttributes()->get(TraceAttributes::DB_SYSTEM));
+        $this->assertEquals('get', $span->getAttributes()->get(TraceAttributes::DB_OPERATION));
+    }
+
+    public function test_set_key(): void
+    {
+        $this->assertCount(0, $this->storage);
+        $driver = $this->createMemoryCacheDriver();
```
ChrisLightfootWild, line 61: Could/should this use `$this->driver` instead?

ChrisLightfootWild (commented):
There are a few places that call: `$driver = $this->createMemoryCacheDriver();` where it looks like `$this->driver` could be used. Might have been part of an earlier refactor but looks like it could be tided up a little. :+1: :broom:

ChrisLightfootWild (approved):
Nice work! :muscle:

src/Instrumentation/Psr16/src/Psr16Instrumentation.php
brettmc: I think the part of the spec Cedric is referring to is https://opentelemetry.io/docs/concepts/signals/traces/#span-kind
> "... the parent of a server span is often a remote client span, and the child of a client span is usually a server span ..."

It's not a MUST though, but let's assume that it is semantically incorrect to have client->client. So, let's make the spans internal, and not use DB attributes. How about using `cache.xxx` as the attribute names? Then, if the underlying implementation _is_ a database, it should set `db.xxx` as appropriate.

I think that's good enough for an initial release.

weslenteche: I like the idea, I was thinking about at least having some attributes like:
- In the operations `get, set, has, getMultiple, has ...` have an attribute like cache.operation
- In` get, set, getMultiple, setMultiple, deleteMultiple and has` operations have a `cache.key` or `cache.keys` attribute to identify the keys used.

These would be some attributes that I see that can make up the span. What do you think ?

brettmc: I think that's good, do it. I'm leaning towards using `cache.keys` and making it an array of strings for the second part.

bobstrecansky:
@weslenteche - are you planning on finishing up this work?

weslenteche:
> @weslenteche - are you planning on finishing up this work?

Hi @bobstrecansky, I finished what was commented [here](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/244#discussion_r1554948301) and I don't have any further updates. Is there any other point to be considered or can we continue like this?

brettmc (approved):

brettmc:
https://packagist.org/packages/open-telemetry/opentelemetry-auto-psr16

PR #250, opened 2024-04-02
Laravel: tracing Queues (+ hook refactoring)

ChrisLightfootWild:
### Tracing Laravel Queues

I have added some message tracing to Laravel.

Ref: https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/

An example trace:

image

---

The above screenshot was produced with the following Laravel test code:
```php
Route::get('/job', function (\Illuminate\Contracts\Queue\Queue $queue) {
    logger()->info('Start request');

    $queue->push(new \App\Jobs\TestJob());

    $queue->bulk([
        new \App\Jobs\TestJob(),
        new \App\Jobs\TestJob(),
        new \App\Jobs\TestJob(),
        new \App\Jobs\TestJob(),
        new \App\Jobs\TestJob(),
    ]);

    $queue->later(
        now()->addSeconds(5),
        new \App\Jobs\TestJob(),
    );

    dispatch_sync(new \App\Jobs\TestJob());

    return [
        'queue' => config('queue.default'),
        'time' => time(),
    ];
});
```
and the following test job:
```php
<?php

namespace App\Jobs;

use Illuminate\Bus\Queueable;
use Illuminate\Contracts\Queue\ShouldBeUnique;
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Queue\InteractsWithQueue;

class TestJob implements ShouldQueue
{
    use InteractsWithQueue, Queueable;

    public function handle(): void
    {
        logger()->info('Start job');

        logger()->info('Job info', [
            'job' => $this->job->getRawBody(),
            'trace' => \OpenTelemetry\API\Trace\Span::getCurrent()->getContext()->getTraceId(),
        ]);

        logger()->debug('Working...', [
            'sleep' => $sleep = rand(10, 1000),
        ]);

        usleep($sleep * 1000);
        logger()->info('End job');
    }
}
```

This is an example of a successful read from the queue:
image

Trace whereby no message received, thus the span was dropped:
image

I tried linking this span to a different parent by extracting the traceparent, but it didn't seem possible to change the existing span's parent.

## Refactoring

During the course of this PR, I decided to try and organise the numerous additional hooks in a way which reflects the layout of the Laravel framework. I hope that this might ultimately make it easier to add further hooks in future.


src/Instrumentation/Laravel/src/Watchers/QueueWatcher.php
```diff
@@ -0,0 +1,90 @@
...
+    public function __construct(
+        private CachedInstrumentation $instrumentation,
+    ) {
+    }
+
+    /** @psalm-suppress UndefinedInterfaceMethod */
+    public function register(Application $app): void
+    {
+        Queue::createPayloadUsing(function () {
+            $opentelemetry = [];
```
brettmc, line 29: maybe name this `carrier` to align with the function's argument?

src/Instrumentation/Laravel/src/Hooks/Queue.php
```diff
@@ -19,9 +21,30 @@ class Queue extends AbstractHook
...
+            $carrier = [];
+            TraceContextPropagator::getInstance()->inject($carrier);
+
+            return $carrier;
+        });
+
+        return hook(
+            AbstractQueue::class,
+            'createPayloadArray',
+            post: function (AbstractQueue $queue, array $params, array $payload, ?Throwable $exception) {
```
brettmc, line 41:
```suggestion
            post: function (AbstractQueue $queue, array $params, array $payload, ?Throwable $exception): array {
```
I think this will resolve your todo above. Modified return value is ignored without a type-hint.

src/Instrumentation/Laravel/src/Watchers/QueueWatcher.php
ChrisLightfootWild, line 1: This can be removed once hook usage is confirmed acceptable.

src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Queue/Queue.php
```diff
@@ -0,0 +1,145 @@
...
+        );
+    }
+
+    protected function hookLater(): bool
+    {
+        return hook(
+            QueueContract::class,
+            'later',
+            pre: function (QueueContract $queue, array $params, string $class, string $function, ?string $filename, ?int $lineno) {
+                $clock = ClockFactory::getDefault();
```
ChrisLightfootWild, line 77: Is this permissible usage of an SDK utility?

src/Instrumentation/Laravel/src/Hooks/Illuminate/Queue/Worker.php
```diff
@@ -0,0 +1,143 @@
...
+                    return;
+                }
+
+                // Discard empty receives.
+                if (!$job) {
+                    Context::storage()->destroy(
+                        $scope->detach(),
+                    );
+
+                    $this->endSpan($exception);
```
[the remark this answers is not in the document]
ChrisLightfootWild: Ah, good point! So if the scope is detached, I guess the spans are discarded with it too - so no additional cleanup required? 🤔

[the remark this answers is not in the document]
brettmc: I don't think that the span is discarded, although possibly would be garbage-collected if there were no remaining references. It'd be better to deliberately close and tidy up if possible.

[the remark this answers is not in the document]
ChrisLightfootWild: @brettmc I'm not entirely sure I'm using this correctly and would be happy to drop this hook if it isn't possible. I did have a look, but think I'm a little outside of my current understanding here! Any pointers appreciated!

src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Queue/Queue.php
brettmc: I think it falls foul of the "should not depend on the SDK" part of spec, ie this package should work without the SDK installed.
I created https://github.com/open-telemetry/opentelemetry-php/pull/1281 to migrate clock into API, but note the downside that using it would mean dropping 8.0 support (unless it's important enough to try to back-patch it)

src/Instrumentation/Laravel/tests/Integration/Queue/QueueTest.php
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * @test
+     */
+    public function it_can_push_a_message_with_a_delay(): void
+    {
+        /** @var ClockInterface|MockInterface $clock */
+        $clock = $this->createMock(ClockInterface::class);
+
+        /** @psalm-suppress UndefinedInterfaceMethod */
+        $clock
```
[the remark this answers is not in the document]
ChrisLightfootWild: That methods returns an int, which is only set once while setting up the mock.

I'm going to replace it with the following to try to avoid confusion:
```php
->willReturn(1234567890 * ClockInterface::NANOS_PER_SECOND);
```

src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Queue/Queue.php
brettmc: https://opentelemetry.io/docs/concepts/instrumentation/libraries/#opentelemetry-api is why the library shouldn't directly use any part of the SDK.

> Libraries should only use the OpenTelemetry API

src/Instrumentation/Laravel/src/Hooks/Illuminate/Queue/Worker.php
```diff
@@ -0,0 +1,141 @@
...
+            },
+            post: function (QueueWorker $worker, array $params, ?Job $job, ?Throwable $exception) {
+                $scope = Context::storage()->scope();
+                if (!$scope) {
+                    return;
+                }
+
+                // Discard empty receives.
+                if (!$job) {
+                    Context::storage()->destroy(
```
brettmc, line 120: not sure that destroy should be used here. I'm not 100% on how it works, but it looks like it will destroy a forked context (eg, for fiber-based processing). That's likely not doing anything here, but if there _were_ multiple forks, it could delete random other forks?
Maybe just detaching the scope will be enough. Thorough testing should uncover whether it works as expected, and DebugScope should complain if any mess was left behind.

src/Instrumentation/Laravel/src/Hooks/Illuminate/Queue/Worker.php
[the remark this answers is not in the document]
brettmc: I left a different comment on `destroy` usage.

src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Queue/Queue.php
ChrisLightfootWild: I've removed the usage of the clock component for now. I simplified the test and hope to revisit this once the clock components are available in the API! :+1:

src/Instrumentation/Laravel/src/Hooks/Illuminate/Queue/Worker.php
ChrisLightfootWild: Thanks @brettmc, it looks like detach is all I needed.

I've added coverage of this via test, `test_it_drops_empty_receives`.

src/Instrumentation/Laravel/composer.json
```diff
@@ -7,21 +7,21 @@
   "readme": "./README.md",
   "license": "Apache-2.0",
   "minimum-stability": "dev",
+  "prefer-stable": true,
   "require": {
     "php": "^8.0",
     "ext-json": "*",
     "ext-opentelemetry": "*",
-    "laravel/framework": ">=6.0",
+    "laravel/framework": "^6.0 || ^7.0 || ^8.0 || ^9.0 || ^10.0 || ^11.0",
```
ChrisLightfootWild, line 15: image

Source: https://laravel.com/docs/11.x/releases#support-policy

Listing the currently supported versions explicitly. I'd like to drop versions 6-9 after we make a 1.0 release of the package so that we can align with the API on PHP ^8.1.

agoallikmaa (approved):

PR #252, opened 2024-04-04
Laravel: InstrumentationInterface experimentation.

ChrisLightfootWild:
# :warning: :construction: Experimental :construction:

I was curious as to how the `InstrumentationInterface` may be used.

This appears to work but does *not* pass under test - I suspect because there are multiple `hook()` registered.

This is more than likely throwaway, so please feel free to tear it apart!

Removed `_register.php`, but it might be necessary to include something akin to:

```php
<?php

declare(strict_types=1);

use Nevay\SPI\ServiceLoader;
use OpenTelemetry\API\Instrumentation\InstrumentationInterface;
use OpenTelemetry\Contrib\Instrumentation\Laravel\LaravelInstrumentation;

// In case spi plugin has not been allowed in composer allow-plugins (root-level).
ServiceLoader::register(InstrumentationInterface::class, LaravelInstrumentation::class);
```

---
Dependencies:
 - https://github.com/open-telemetry/opentelemetry-php/pull/1274


src/Instrumentation/Laravel/src/LaravelInstrumentation.php
```diff
@@ -4,52 +4,81 @@
...
-                self::registerWatchers($application, new ExceptionWatcher());
-                self::registerWatchers($application, new LogWatcher());
-                self::registerWatchers($application, new QueryWatcher($instrumentation));
-                self::registerWatchers($application, new RequestWatcher());
-            },
-        );
+    public function activate(): bool
+    {
+        if (!$this->init()) {
+            return false;
```
ChrisLightfootWild, line 53: I was unsure of the expected usage of this, based on [these comments](https://github.com/open-telemetry/opentelemetry-php/blob/e44f5a97d999e88de41aed7d54723354ea8505bf/src/API/Instrumentation/InstrumentationTrait.php#L100-L105).

src/Instrumentation/Laravel/_register.php
```diff
@@ -5,14 +5,18 @@
...
-    return;
-}
+    if (extension_loaded('opentelemetry') === false) {
+        trigger_error('The opentelemetry extension must be loaded in order to autoload the OpenTelemetry Laravel auto-instrumentation', E_USER_WARNING);
 
-LaravelInstrumentation::register();
+        return;
+    }
+
+    $instrumentation->activate();
```
ChrisLightfootWild, line 21: Could there be a way to register this instrument with the API so that activation is delegated elsewhere?

ChrisLightfootWild: Toying with this idea as part of this PR now, so resolving this comment.

PR #255, opened 2024-04-15
Azure Resource Detector

matt-hensley:
This package provides OpenTelemetry `ResourceDetector`s which will detect 
resource attributes for these Azure services:
* App Service
* Container Apps
* Virtual Machines

The README has a list of resource attributes provided by the detectors.


src/Azure/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Installation via composer
+
+```bash
+$ composer require open-telemetry/contrib-azure
+```
+
+## Usage
+
+The detector will be automatically registered as part of composer autoloading.
```
ChrisLightfootWild, line 50: I think this might have been copied over from the `open-telemetry/detector-container` package, so this is not currently the case for this Azure package. Looks like you would just need to copy the `_register.php` logic across to achieve the same!

ChrisLightfootWild: @matt-hensley final bit that makes the register call work is by adding to autoloading mechanism in composer (eg. [here](https://github.com/open-telemetry/opentelemetry-php-contrib/blob/fad23386904c3a5c6e0d9b669e0b0d248f86dc5b/src/ResourceDetectors/Container/composer.json#L18-L21)).

agoallikmaa (approved):

src/Azure/composer.json
```diff
@@ -0,0 +1,54 @@
...
+    ],
+    "minimum-stability": "dev",
+    "prefer-stable": true,
+    "require": {
+        "php": "^7.4 || ^8.0",
+        "guzzlehttp/guzzle": "^7.4",
+        "guzzlehttp/psr7": "^2.6",
+        "open-telemetry/api": "^1.0",
+        "open-telemetry/sdk": "^1.0",
+        "php-http/guzzle7-adapter": "^1.0"
```
brettmc, line 21: rather than requiring guzzle, it could use psr-18 discovery to find an http client (which is what the SDK does)

src/Azure/src/Vm/Detector.php
```diff
@@ -0,0 +1,83 @@
...
+                ResourceAttributes::HOST_NAME => $metadata['compute']['name'],
+                ResourceAttributes::HOST_TYPE => $metadata['compute']['vmSize'],
+                ResourceAttributes::OS_TYPE => $metadata['compute']['osType'],
+                ResourceAttributes::OS_VERSION => $metadata['compute']['version'],
+                ResourceAttributes::SERVICE_INSTANCE_ID => $metadata['compute']['vmId'],
+            ];
+
+            return ResourceInfo::create(Attributes::create($attributes), ResourceAttributes::SCHEMA_URL);
+        } catch (Throwable $e) {
+            return ResourceInfo::emptyResource();
```
brettmc, line 71: could log a warning here? otherwise, how will people know when/why it's not working?

src/Azure/src/Vm/Detector.php
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    private function getAzureMetadata()
+    {
+        $req = $this->requestFactory->createRequest('GET', self::AZURE_METADATA_ENDPOINT_URL);
+        $req->withHeader('Metadata', 'true');
```
brettmc, line 78: psr-7 messages are [immutable](https://www.php-fig.org/psr/psr-7/#31-psrhttpmessagemessageinterface) so this probably doesn't work.
```suggestion
        $req = $req->withHeader('Metadata', 'true');
```

src/Azure/src/Vm/Detector.php
```diff
@@ -0,0 +1,83 @@
...
+use Psr\Http\Client\ClientInterface;
+use Psr\Http\Message\RequestFactoryInterface;
+use Throwable;
+
+/**
+ * AzureVmDetector
+ */
+class Detector implements ResourceDetectorInterface
+{
+    private const AZURE_METADATA_ENDPOINT_URL = 'http://169.254.169.254/metadata/instance/compute?api-version=2021-12-13&format=json';
```
brettmc, line 36: This will work for everybody, forever? There's not a well-known service name or something?

src/Azure/tests/Unit/AppService/DetectorTest.php
```diff
@@ -0,0 +1,70 @@
...
+            if ($item[1] !== null) {
+                putenv($item[1]);
+            }
+        }
+    }
+
+    /**
+     * @test
+     */
+    public function TestResourceUriGeneration()
```
brettmc, line 65: project style is to use `snake_case` for test functions, starting with `test_`, and not to use the `@test` annotation.

src/Azure/src/Vm/Detector.php
matt-hensley: Per [Azure documentation](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows) this is the official way of accessing the metadata service. Unfortunately there is not a well-known hostname published, just a fixed IP.

> IMDS is a REST API that's available at a well-known, non-routable IP address (169.254.169.254). You can only access it from within the VM. Communication between the VM and IMDS never leaves the host. Have your HTTP clients bypass web proxies within the VM when querying IMDS, and treat 169.254.169.254 the same as [168.63.129.16](https://learn.microsoft.com/en-us/azure/virtual-network/what-is-ip-address-168-63-129-16).

src/Azure/.phan
brettmc, line 1: Can this be a copy not a symlink? For local development, we docker volume mount just the subproject into a container, and symlinks don't work.

src/Azure/src/Vm/Detector.php
```diff
@@ -0,0 +1,90 @@
...
+use Throwable;
+
+/**
+ * AzureVmDetector
+ */
+class Detector implements ResourceDetectorInterface
+{
+    use LogsMessagesTrait;
+
+    private const AZURE_METADATA_ENDPOINT_URL = 'http://169.254.169.254/metadata/instance/compute?api-version=2021-12-13&format=json';
```
brettmc, line 41:
```suggestion
    /** @see https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows */
    private const AZURE_METADATA_ENDPOINT_URL = 'http://169.254.169.254/metadata/instance/compute?api-version=2021-12-13&format=json';
```

src/Azure/tests/Unit/AppService/DetectorTest.php
```diff
@@ -0,0 +1,64 @@
...
+            [ResourceAttributes::DEPLOYMENT_ENVIRONMENT, Detector::ENV_WEBSITE_SLOT_NAME_KEY, 'testing'],
+            [ResourceAttributes::HOST_ID, Detector::ENV_WEBSITE_HOSTNAME_KEY, 'example.com'],
+            [ResourceAttributes::SERVICE_INSTANCE_ID, Detector::ENV_WEBSITE_INSTANCE_ID_KEY, uniqid()],
+            [ResourceAttributes::SERVICE_NAME, Detector::ENV_WEBSITE_SITE_NAME_KEY, self::SERVICE_NAME],
+            [null, Detector::ENV_WEBSITE_RESOURCE_GROUP_KEY, self::RESOURCE_GROUP],
+            [null, Detector::ENV_WEBSITE_OWNER_NAME_KEY, self::OWNER_NAME],
+        ];
+        $expected = [];
+
+        foreach ($data as $item) {
```
brettmc, line 36: in the main otel repo, we use assertwell/phpunit-global-state to set/restore environment for tests, eg https://github.com/open-telemetry/opentelemetry-php/blob/main/tests/Unit/SDK/SdkTest.php#L26

src/Azure/composer.json
```diff
@@ -0,0 +1,55 @@
...
+        {
+            "name": "opentelemetry-php-contrib contributors",
+            "homepage": "https://github.com/open-telemetry/opentelemetry-php-contrib/graphs/contributors"
+        }
+    ],
+    "minimum-stability": "dev",
+    "prefer-stable": true,
+    "require": {
+        "php": "^7.4 || ^8.0",
+        "guzzlehttp/guzzle": "^7.8",
```
brettmc, line 17: use psr-18 discovery (but can use guzzle as a dev dependency if it makes testing easier)

src/Azure/composer.json
```diff
@@ -0,0 +1,55 @@
...
+        "open-telemetry/api": "^1.0",
+        "open-telemetry/sdk": "^1.0"
```
brettmc, line 19: ideally, and per [spec](https://opentelemetry.io/docs/concepts/instrumentation/libraries/#opentelemetry-api), instrumentation libraries should not depend on the SDK (which I think contradicts something I said in an earlier comment on this PR, sorry).

matt-hensley: Without the SDK dependency it's not possible to call `Registry::registerResourceDetector`. I modeled this after the container resource detector.

Is there an alternative approach you'd recommend?

brettmc: Yes, you are right. I was thinking that this was an "instrumenting library" but it's really an SDK extension. My bad, continue as you were :)

src/Azure/README.md
```diff
@@ -0,0 +1,59 @@
+[![Releases](https://img.shields.io/badge/releases-purple)](https://github.com/opentelemetry-php/contrib-aws/releases)
+[![Issues](https://img.shields.io/badge/issues-pink)](https://github.com/open-telemetry/opentelemetry-php/issues)
+[![Source](https://img.shields.io/badge/source-contrib-green)](https://github.com/open-telemetry/opentelemetry-php-contrib/tree/main/src/Azure)
+[![Mirror](https://img.shields.io/badge/mirror-opentelemetry--php--contrib-blue)](https://github.com/opentelemetry-php/contrib-azure)
+[![Latest Version](http://poser.pugx.org/open-telemetry/contrib-azure/v/unstable)](https://packagist.org/packages/open-telemetry/contrib-azure/)
```
brettmc, line 5: can we name the package `detector-azure`?

src/Azure/composer.json
```diff
@@ -0,0 +1,56 @@
+{
+    "name": "open-telemetry/contrib-azure",
```
brettmc, line 2:
```suggestion
    "name": "open-telemetry/detector-azure",
```

brettmc:
Just about ready to merge, I think. I created https://github.com/opentelemetry-php/contrib-detector-azure as the git-split destination, can you please add that to the top-level `.gitsplit.yaml` file, and update the links in readme accordingly.
Also, can you add this to the github actions so that tests run as part of CI?

brettmc (approved):

brettmc:
Follow-up PR #258

brettmc:
https://packagist.org/packages/open-telemetry/detector-azure

PR #257, opened 2024-04-18
fixing build failures

brettmc:
- CodeIgniter tests only pass against 4.4, so update minimum (with a case insensitivity change to test)
- MongoDB using a now-deprecated semconv, which can be removed (we already have the suggested replacement)
- Psr16 using an old version of psalm
- adding missing phan config for ServerTiming and Container detector
- adding update-lowest and all-lowest Make targets to run tests against lowest installable dependencies
- remove deprecated docker-compose.version
- update Dockerfile to use the core one, with some contrib additions (opentelemetry, mongo)

I've left Laravel's phan complaints, because I think Chris might be working on it.


ChrisLightfootWild (approved):
Nice!

I will fix any Laravel ones in my branch once this is merged 👍

bobstrecansky:
Looks like the only thing left here is @ChrisLightfootWild's Laravel tests that will get fixed subsequently.  Good to merge.

PR #259, opened 2024-04-26
Include schema_url in CachedInstrumentation

sudol:
This change is in reference to https://github.com/open-telemetry/opentelemetry-php/issues/1290 and is in line with other Auto Instrumentation implementations.

https://github.com/open-telemetry/opentelemetry-php-contrib/blob/491827d0c93921669e657a37b071e3413bcbf4f5/src/Instrumentation/Guzzle/src/GuzzleInstrumentation.php#L31

https://github.com/open-telemetry/opentelemetry-php-contrib/blob/491827d0c93921669e657a37b071e3413bcbf4f5/src/Instrumentation/ExtAmqp/src/ExtAmqpInstrumentation.php#L27-L30

### DIFF
--- a/src/Instrumentation/Slim/src/SlimInstrumentation.php
+++ b/src/Instrumentation/Slim/src/SlimInstrumentation.php
@@ -29,7 +29,7 @@ class SlimInstrumentation
 
     public static function register(): void
     {
-        $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.slim');
+        $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.slim', schemaUrl: TraceAttributes::SCHEMA_URL);
         /**
          * requires extension >= 1.0.2beta2
          * @see https://github.com/open-telemetry/opentelemetry-php-instrumentation/pull/136


brettmc (approved):

src/Instrumentation/Slim/src/SlimInstrumentation.php
```diff
@@ -29,7 +29,7 @@ class SlimInstrumentation
 
     public static function register(): void
     {
-        $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.slim');
+        $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.slim', schemaUrl: TraceAttributes::SCHEMA_URL);
```
ChrisLightfootWild, line 32: Should this be explicitly set as the schema that it was created against?

It may potentially be invalid against future versions of the spec 🤔

sudol: You're right, I hadn't considered this. Since this package (opentelemetry-auto-slim), pulls in open-telemetry/sem-conv, which is the source of TraceAttributes, is it fair to assume that this package is compatible with whatever version of the schema is defined in open-telemetry/sem-conv?

ChrisLightfootWild: Apologies for my delayed response to this @sudol 

I have had a look through the documentation, and I think this may be the relevant part that was somewhere in the back of my mind: [Schema-File Driven Telemetry Producers](https://opentelemetry.io/docs/specs/otel/telemetry-stability/#schema-file-driven-telemetry-producers)

I think that we would be at risk of changing the schema version URL in a potentially unsafe way if `TraceAttributes::SCHEMA_URL` is used as is.

@brettmc did mention in the SIG this week that maybe we could provide a series of constants for these versions instead (tagged for his input/thoughts on this!)

brettmc: how does this look? https://github.com/open-telemetry/opentelemetry-php/pull/1301

brettmc:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/telemetry-stability.md#schema-file-driven-telemetry-producers was just pointed out in the spec channel... specifically:

> Stable instrumentations that include the Schema URL in the produced telemetry are called Schema-File Driven Telemetry Producers.
Such instrumentations are prohibited from changing the produced telemetry until the moratorium on relying on schema transformations for telemetry stability is lifted

The restrictions aren't any better for instrumentations that don't include a schema URL, though...

src/Instrumentation/Slim/src/SlimInstrumentation.php
ChrisLightfootWild: Nice @brettmc! I added a comment on there for discussion

ChrisLightfootWild:
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/telemetry-stability.md#schema-file-driven-telemetry-producers was just pointed out in the spec channel... specifically:
> 
> > Stable instrumentations that include the Schema URL in the produced telemetry are called Schema-File Driven Telemetry Producers.
> > Such instrumentations are prohibited from changing the produced telemetry until the moratorium on relying on schema transformations for telemetry stability is lifted
> 
> The restrictions aren't any better for instrumentations that don't include a schema URL, though...

Happy to go with your judgement either way!

src/Instrumentation/Slim/src/SlimInstrumentation.php
brettmc: My interpretation of the spec re: schema_url is that we should explicitly set it to the latest schema version (`1.25.0`), confirm that we're setting the values correctly per that version, then leave it alone.

brettmc: @sudol - use `https://opentelemetry.io/schemas/1.25.0`, and the version should not change again after that.

sudol: Ah yes. Sorry, I misunderstood. Yes I will get this updated tomorrow.

brettmc (approved):

PR #260, opened 2024-04-30
link pdo prepare spans to execute + fetchAll spans

brettmc:
- generalise PdoAttributeTracker to also track statements (and rename to suit)
- set all possible values on spans before starting them (per spec)
- import classes
- fix psalm config warnings
- prefer stable composer packages

Fixes: https://github.com/open-telemetry/opentelemetry-php/issues/1286
Fixes: https://github.com/open-telemetry/opentelemetry-php/issues/1285

### DIFF
--- a/src/Instrumentation/PDO/composer.json
+++ b/src/Instrumentation/PDO/composer.json
@@ -7,6 +7,7 @@
   "readme": "./README.md",
   "license": "Apache-2.0",
   "minimum-stability": "dev",
+  "prefer-stable": true,
   "require": {
     "php": "^8.2",
     "ext-pdo": "*",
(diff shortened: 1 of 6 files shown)


brettmc:
> Saw some failed tests, is that expected

@bobstrecansky All failing due to codecov timeouts, which I resolved in another PR and should go green right...about....now.

brettmc:
> it might make sense to reconsider

Sure! Having a quick scan through that PR, we've definitely implemented some of the recommendations now - forcing covers annotations and ensuring every test class declares what it is covering.

auto-instrumentation tests are a little different though, in their current state they can really only to integration tested since they rely on the extension. That said, we could probably improve the coverage reliability by at least having each test declare which methods it is expecting to cover.

And, happy to see what mutation testing does for us. Anything that improves our quality or reliability (without adding too much ongoing maintenance burden) ) is welcome.

dkarlovi:
From my experience, the only overhead is it making you add the missing tests. 😀

bobstrecansky (approved):

PR #264, opened 2024-05-17
Restrict trace events to correct log level

alecsammon:
We noticed during testing that debug messages were appearing as events in our traces.

This change only adds events if the level would be handled by the logger - i.e. the level is at least the same as the defined minimum log level.


brettmc (approved):

brettmc:
Looks like a sensible change to me. Can you take a look at the phan issue? It might just need to be suppressed...

alecsammon:
Thanks Brett. Have added a phan supression - hopefully build will be 🟢 now.

brettmc:
@alecsammon progress, now some phpstan issues. FYI you can run all of the CI checks locally: `PHP_VERSION=8.3 PROJECT=Instrumentation/Laravel make all`

src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
```diff
@@ -6,23 +6,37 @@
...
 
     /**
      * Record a log.
      */
     public function recordLog(MessageLogged $log): void
     {
+        $underlyingLogger = $this->logger->getLogger();
+
+        /** @phan-suppress-next-line PhanUndeclaredMethod */
+        if (method_exists($underlyingLogger, 'isHandling') && !$underlyingLogger->isHandling($log->level)) {
```
alecsammon, line 36: looks like the `isHandling` is only available in the monolog - my understanding is that someone could pass a custom logger.

I'm not sure I like the `method_exists` - but maybe is the safest way of ensuring this doesn't break any code?

src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
```diff
@@ -6,23 +6,37 @@
 
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Log\Events\MessageLogged;
+use Illuminate\Log\LogManager;
 use OpenTelemetry\API\Trace\Span;
 use OpenTelemetry\Context\Context;
 
 class LogWatcher extends Watcher
 {
+    /** @var LogManager */
```
ChrisLightfootWild, line 15: Could drop this bit:
```suggestion
```

src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
ChrisLightfootWild: Hey Alec :wave:

I had the same understanding, as it looks like PSR-3 doesn't account for that functionality.

This isn't the first `method_exists()` usage at least across contrib. Another option could be an `instanceof` check to see if the method exists and is compatible.

I wonder if OTel has some filtering mechanism in general for this, as I imagine the PSR-3 instrumentation would log everything as well.

ChrisLightfootWild (approved):

src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
alecsammon: 👋🏻  Hi! 

I do think that this isn't perfect - but maybe prevents at least some leaking of debug logs into enviroments where it shouldn't be.

Looks like a processor could also be used - i.e. in a collector - to give more options for dropping events - see https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/filterprocessor/README.md.
However this is maybe more advanced - and for people with basic setups it's probably better to have this as a fallback.

alecsammon:
@brettmc - I'm hoping that the build should all be sorted now!

Are you able to re-run the workflow when you're able?

alecsammon:
Investigating build failures again - sorry :(

alecsammon:
I'm wondering if this is right approach at all.

It maybe better to use a environment variable to determine which levels should be logged

I.e. https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/logging/logging.html#envvar-OTEL_PYTHON_LOG_LEVEL

https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/ does describe a OTEL_LOG_LEVEL which is maybe the right one to use? Or `OTEL_PHP_LOG_LEVEL` may be better?

agoallikmaa:
I looked into how Java instrumentation handles this, and there for the most common logging library (see [Logback appender instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/logback/logback-appender-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/logback/appender/v1_0/LogbackInstrumentation.java#L37)), the instrumented method itself is only ever called if some log level checks already passed, so this filter technically exists there even if it's due to the instrumentation hook never being reached otherwise. So there is definitely also a precedent for this approach in other OpenTelemetry implementations.

As for what a global configuration option would do or if one is even needed, that could be a separate topic of discussion. If such an option were to be added, it could act merely as an additional filter, for example to exclude `debug` and lower even if they are configured to get logged somewhere otherwise.

brettmc:
I'm wondering whether attaching logs as trace events is the right thing to do at all? Is there any precedent? Otel has a logging signal, and we already have contrib modules which will take monolog/psr-3 log messages and forward them in the opentelemetry format, including trace+span id.

If we were to go with an environment variable, I don't think `OTEL_LOG_LEVEL` is the correct one, because that's for configuring the SDK logger, ie errors, warnings etc coming out of our SDK, which is unrelated to application log level.

src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
agoallikmaa: Is it expected that the underlying logger could change? If not, maybe it is possible to get the underlying one already in `register` and only assign to a property if `method_exists` passes there.

bobstrecansky:
@alecsammon - are you still interested in pushing this forward?

PR #267, opened 2024-05-23
fix: post hook invalid when symfony messenger sender throws exception

nesl247:
I still need to validate that this solves the problem but looking at the code this is the only thing that makes sense to me to solve it.

I’m opening this as a draft because we can’t submit issues, so while validating this if anyone else comes across it this may help them.

```
Symfony\Component\Messenger\Bridge\Amqp\Transport\AmqpTransport::send(): OpenTelemetry: post hook invalid signature, class=Symfony\Component\Messenger\Bridge\Amqp\Transport\AmqpTransport function=send
```

### DIFF
--- a/src/Instrumentation/Symfony/src/MessengerInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/MessengerInstrumentation.php
@@ -150,7 +150,7 @@ final class MessengerInstrumentation
             post: static function (
                 SenderInterface $sender,
                 array $params,
-                Envelope $result,
+                ?Envelope $result,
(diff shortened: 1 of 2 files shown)


brettmc:
Hi! Issues can be created over in our main repo, https://github.com/open-telemetry/opentelemetry-php

PR makes sense, and something similar was done for `MessageBusInterface::dispatch`. Although the interface should always return an Envelope, that's not true for a post hook function on exception.

Could you please add a test for this? The happy path is covered, but not exception.

brettmc (approved):

PR #268, opened 2024-05-23
Adding JsonFormatter Support to Monolog Handler

Spittal:
In the current iteration `$record['formatted']` is a string, which is then used later used in an attempt to access `message` from. I can only imagine this was typo, or an oversight.

I have tested locally and this works now.

### DIFF
--- a/src/Logs/Monolog/src/Handler.php
+++ b/src/Logs/Monolog/src/Handler.php
@@ -45,16 +45,15 @@ class Handler extends AbstractProcessingHandler
      */
     protected function write($record): void
     {
-        $formatted = $record['formatted'];
         $logRecord = (new API\LogRecord())
             ->setTimestamp((int) $record['datetime']->format('Uu') * 1000)
             ->setSeverityNumber(API\Map\Psr3::severityNumber($record['level_name']))
             ->setSeverityText($record['level_name'])
-            ->setBody($formatted['message'])
+            ->setBody($record['message'])
         ;
         foreach (['context', 'extra'] as $key) {
-            if (isset($formatted[$key]) && count($formatted[$key]) > 0) {
-                $logRecord->setAttribute($key, $formatted[$key]);
+            if (isset($record[$key]) && count($record[$key]) > 0) {
+                $logRecord->setAttribute($key, $record[$key]);
             }
         }
         $this->getLogger($record['channel'])->emit($logRecord);
--- a/src/Logs/Monolog/src/Handler.php
+++ b/src/Logs/Monolog/src/Handler.php
@@ -45,15 +45,18 @@ class Handler extends AbstractProcessingHandler
      */
     protected function write($record): void
     {
+        $formatted = ($this->getFormatter() instanceof JsonFormatter)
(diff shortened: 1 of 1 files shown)


brettmc:
Hi! Issues can be created over in our main repo, https://github.com/open-telemetry/opentelemetry-php

It's not a typo, but it could be a bug, depending on which Monolog [formatter](https://github.com/Seldaek/monolog/blob/main/doc/02-handlers-formatters-processors.md#formatters) you are using. We assume the default formatter: https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/FormattableHandlerTrait.php#L42 and I guess you're not using that?

Spittal:
> It's not a typo, but it could be a bug, depending on which Monolog [formatter](https://github.com/Seldaek/monolog/blob/main/doc/02-handlers-formatters-processors.md#formatters) you are using. A formatter must be used: https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/FormattableHandlerTrait.php#L42 and we configure `NormalizerFormatter` as the default, I am guessing you've chosen a different formatter?

That's correct, we're using the `JsonFormatter`. Our exact setup is in the context of Laravel of looks like this.

```
'otel' => [
    'driver' => 'monolog',
    'handler' => \OpenTelemetry\Contrib\Logs\Monolog\Handler::class,
    'with' => [
        'loggerProvider' => \OpenTelemetry\API\Globals::loggerProvider(),
        'level' => env('LOG_LEVEL', 'debug'),
        'bubble' => true,
    ],
    'replace_placeholders' => true,
    'formatter' => Monolog\Formatter\JsonFormatter::class,
],
```

Spittal:
Okay I've tested the following formatters

- LineFormatter  
- HtmlFormatter  
- NormalizerFormatter  
- ScalarFormatter  
- JsonFormatter  
- WildfireFormatter  
- ChromePHPFormatter  
- GoogleCloudLoggingFormatter  

The others [listed here](https://seldaek.github.io/monolog/doc/02-handlers-formatters-processors.html) aren't relevant as far as I can tell.

The most recent commit will decode the JSON before populating the `LogRecord`. This is _probably_ :100:% the desired behaviour, but it's notable that in my case using Loki that I am not required to use the Loki JSON Formatter. If folks _really_ want to have a JSON being received by Loki they can do that in the Otel collector with the Loki exporter I believe.

src/Logs/Monolog/src/Handler.php
```diff
@@ -45,7 +45,9 @@ protected function getDefaultFormatter(): FormatterInterface
      */
     protected function write($record): void
     {
-        $formatted = $record['formatted'];
+        $formatted = ($this->getFormatter() instanceof JsonFormatter)
```
agoallikmaa, line 48: Shouldn't there either be an `use` for `JsonFormatter` or fully qualified name?

brettmc:
@Spittal hi, sorry for the slow response. The build is failing on this PR, and it looks like the issue that Ago mentioned above

brettmc:
ping @Spittal

PR #269, opened 2024-06-10
👷 Laravel instrumentation 1.1.x

ChrisLightfootWild:
WIP: current test approach needs tweaking.

Using new functionality provided via https://github.com/open-telemetry/opentelemetry-php/pull/1304.

This should probably form the basis of a v1.1 package, so depends on a 1.0 release of the current Laravel instrumentation.


src/Instrumentation/Laravel/src/LaravelConfiguration.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    private function __construct(
+        public readonly bool $enabled,
+    ) {
+    }
+
+    public static function fromArray(array $properties): self
+    {
+        return new self(
+            enabled: $properties['enabled'] ?? (class_exists(Sdk::class) ? !Sdk::isDisabled() : true),
```
ChrisLightfootWild, line 20: I think properties is always populated by `withDefaults` functionality, so will need to double check this funtionality.

src/Instrumentation/Laravel/src/LaravelInstrumentation.php
```diff
@@ -4,33 +4,38 @@
...
-        Hooks\Illuminate\Queue\Worker::hook($instrumentation);
+        $config = $configuration->get(LaravelConfiguration::class) ?? LaravelConfiguration::default();
+
+        if (! $config->enabled) {
+            return;
+        }
+
+        $logger = $context->loggerProvider->getLogger(self::INSTRUMENTATION_NAME);
+        $meter = $context->meterProvider->getMeter(self::INSTRUMENTATION_NAME);
+        $tracer = $context->tracerProvider->getTracer(self::INSTRUMENTATION_NAME);
```
ChrisLightfootWild, line 29: CachedInstrumentation deferred this, but I was trying to avoid having to keep creating tracers.

src/Instrumentation/Laravel/src/LaravelInstrumentation.php
```diff
@@ -4,33 +4,38 @@
...
+
+        foreach (ServiceLoader::load(Hook::class) as $hook) {
+            /** @var Hook $hook */
+            $hook->instrument($hookManager, $config, $logger, $meter, $tracer);
+        }
```
ChrisLightfootWild, line 34: Lean into the power of SPI :eyes:

src/Instrumentation/Laravel/src/LaravelInstrumentation.php
```diff
@@ -4,33 +4,38 @@
...
     }
 
     public static function shouldTraceCli(): bool
     {
-        return PHP_SAPI !== 'cli' || (
-            class_exists(Configuration::class)
-            && Configuration::getBoolean('OTEL_PHP_TRACE_CLI_ENABLED', false)
-        );
+        return PHP_SAPI !== 'cli' || (new ConfigurationResolver())->getBoolean('OTEL_PHP_TRACE_CLI_ENABLED');
```
ChrisLightfootWild, line 39: This works without an SDK component now (I was previously unaware of the ConfigurationResolver).

src/Instrumentation/Laravel/_register.php
```diff
@@ -2,17 +2,23 @@
...
 
-if (class_exists(Sdk::class) && Sdk::isInstrumentationDisabled(LaravelInstrumentation::NAME) === true) {
-    return;
-}
+// In case the `tbachert/spi` plugin has not been allowed in composer allow-plugins (root-level).
+ServiceLoader::register(Instrumentation::class, LaravelInstrumentation::class);
 
-if (extension_loaded('opentelemetry') === false) {
-    trigger_error('The opentelemetry extension must be loaded in order to autoload the OpenTelemetry Laravel auto-instrumentation', E_USER_WARNING);
+ServiceLoader::register(Hooks\Hook::class, Hooks\Illuminate\Console\Command::class);
```
ChrisLightfootWild, line 13: Having mapped all these out here, I might as well have kept them in `LaravelInstrumentation` and just called them there without SPI.

ChrisLightfootWild: Thanks to the work from @Nevay (and [these instructions](https://github.com/Nevay/spi/issues/1#issuecomment-2198483721)), this has now been improved to de-dupe registration.

src/Instrumentation/Laravel/composer.json
```diff
@@ -9,18 +9,16 @@
   "minimum-stability": "dev",
   "prefer-stable": true,
   "require": {
-    "php": "^8.0",
+    "php": "^8.1",
     "ext-json": "*",
     "ext-opentelemetry": "*",
-    "laravel/framework": "^6.0 || ^7.0 || ^8.0 || ^9.0 || ^10.0 || ^11.0",
-    "open-telemetry/api": "^1.0",
-    "open-telemetry/sem-conv": "^1.24"
```
ChrisLightfootWild, line 17: Once instrumentation PR is merged/tagged, bump ensure `sem-conv:^1.25`

src/Instrumentation/Laravel/src/LaravelComponentProvider.php
```diff
@@ -0,0 +1,29 @@
...
+        return LaravelConfiguration::fromArray($properties);
+    }
+
+    public function getConfig(ComponentProviderRegistry $registry): ArrayNodeDefinition
+    {
+        $root = new ArrayNodeDefinition('laravel');
+
+        return $root
+            ->canBeDisabled()
+            ->addDefaultsIfNotSet()
```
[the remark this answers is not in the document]
brettmc: I think you can do that, the laravel configuration schema can accept any shape you can define with symfony/config. For example, create an array for each component that is instrumented, which can individually be turned on/off, configure other values, ...

src/Instrumentation/Laravel/src/LaravelInstrumentation.php
brettmc: We shouldn't need CachedInstrumentation going forward, and creating one tracer per hook registration is fine.

src/Instrumentation/Laravel/composer.json
ChrisLightfootWild: Leave as is for now.

src/Instrumentation/Laravel/src/Hooks/Hook.php
```diff
@@ -0,0 +1,22 @@
...
+use OpenTelemetry\Contrib\Instrumentation\Laravel\LaravelConfiguration;
+
+interface Hook
+{
+    public function instrument(
+        HookManagerInterface $hookManager,
+        LaravelConfiguration $configuration,
+        LoggerInterface $logger,
+        MeterInterface $meter,
+        TracerInterface $tracer,
```
ChrisLightfootWild, line 20: Should I actually be passing the `*ProviderInterfaces` around?

My current understanding would be that I'd have to duplicate the instantiation or provide a Factory to give me back the implementations I want.

brettmc: Yes, I think you should. I also think you should just accept the `InstrumentationContext` here, and retrieve the providers you need for this particular hook (less rework if we add new providers in future). Also, I think the Tracer/Logger/etc for this hook should have a unique name that includes the component being instrumented, eg `io.opentelemetry.contrib.php.laravel.kernel` - the reason for this is that we can use [scope configuration](https://github.com/open-telemetry/opentelemetry-php/pull/1353) to turn off parts of auto-instrumentation based on that name.

ChrisLightfootWild: Thanks @brettmc, all of that sounds sensible. I will look to make the switch.

I will have to have a tinker with the scope configuration here and see how that goes!

brettmc (approved):

src/Instrumentation/Laravel/composer.json
```diff
@@ -9,26 +9,24 @@
   "minimum-stability": "dev",
   "prefer-stable": true,
   "require": {
-    "php": "^8.0",
+    "php": "^8.1",
     "ext-json": "*",
-    "ext-opentelemetry": "*",
```
ChrisLightfootWild, line 14: Changed based on [similar issue](https://github.com/open-telemetry/opentelemetry-php/issues/1191) raised for PDO instrumentation.

src/Instrumentation/Laravel/src/Hooks/Hook.php
andrii-androshchuk, line 1: Great interface!

src/Instrumentation/Laravel/src/Hooks/Illuminate/Foundation/Console/ServeCommand.php
```diff
@@ -5,28 +5,27 @@
...
-    public function instrument(): void
-    {
-        /** @psalm-suppress UnusedFunctionCall */
-        hook(
+    public function instrument(
+        LaravelConfiguration $configuration,
+        HookManagerInterface $hookManager,
+        InstrumentationContext $context,
+    ): void {
+        $hookManager->hook(
```
andrii-androshchuk, line 25: I would add `$configuration->shouldTraceCli()` check here as well, as the `serve` command is being called from the CLI only.

src/Instrumentation/Laravel/composer.json
```diff
@@ -17,30 +17,23 @@
   "prefer-stable": true,
   "require": {
     "php": "^8.1",
-    "ext-json": "*",
-    "ext-opentelemetry": "*",
-    "laravel/framework": "^6.0 || ^7.0 || ^8.0 || ^9.0 || ^10.0 || ^11.0 || ^12.0",
+    "laravel/framework": "^10.0 || ^11.0 || ^12.0",
```
andrii-androshchuk, line 20: Lets add `^13.0`.

andrii-androshchuk (commented):
Great work, @ChrisLightfootWild! Left a few specific comments, and one more general comment / question:

The current release uses `Version::VERSION_1_38_0`, while this PR still points to `Version::VERSION_1_24_0`. 
- Should we keep the newer version? 
- Is there a way to define it once and for entire library, not in individual hook / component?

src/Instrumentation/Laravel/src/ComponentLoader/LaravelComponentLoader.php
andrii-androshchuk, line 1: `LaravelComponentLoader` and `LaravelComponentProvider` are currently placed in their own dedicated folders (namespaces). However, at the moment, I am not sure whether there will be additional classes alongside them in the future.

What do you think about moving them into the `Configuration` namespace instead (which is already the parent namespace for both `EnvComponentLoader` and `ComponentProvider`)?

That would look something like this:

- `OpenTelemetry\Contrib\Instrumentation\Laravel\Configuration\LaravelComponentProvider`
- `OpenTelemetry\Contrib\Instrumentation\Laravel\Configuration\LaravelComponentLoader`

src/Instrumentation/Laravel/src/Configuration/ConfigEnv/LaravelComponentLoader.php
ChrisLightfootWild: Thanks for that suggestion - implemented in 5f65f37 (although I also split into `Config` & `ConfigEnv` to reflect the API layout)

src/Instrumentation/Laravel/composer.json
ChrisLightfootWild: This got added when I merged main back in. This branch was just so old that it pre-dated 13 :laughing:

src/Instrumentation/Laravel/src/Watchers/RedisCommand/RedisCommandWatcher.php
```diff
@@ -25,15 +27,17 @@
...
     ) {
     }
 
     /** @psalm-suppress UndefinedInterfaceMethod */
+    #[\Override]
     public function register(Application $app): void
     {
-        /** @phan-suppress-next-line PhanTypeArraySuspicious */
-        $app['events']->listen(CommandExecuted::class, [$this, 'recordRedisCommand']);
+        $app->afterResolving('events', function (Dispatcher $dispatcher) {
```
ChrisLightfootWild, line 38: Defer container resolution of the `events` object.

brettmc (approved):

PR #270, opened 2024-06-13
[opentelemetry-auto-slim] fix schema url

weslenteche:
This pull request aims to fix an issue identified in which the value of `schemaUrl` is being inferred in the `$version` parameter instead of `$schemaUrl`

### DIFF
--- a/src/Instrumentation/Slim/src/SlimInstrumentation.php
+++ b/src/Instrumentation/Slim/src/SlimInstrumentation.php
@@ -30,8 +30,9 @@ class SlimInstrumentation
     public static function register(): void
     {
         $instrumentation = new CachedInstrumentation(
-            name: 'io.opentelemetry.contrib.php.slim',
-            schemaUrl: 'https://opentelemetry.io/schemas/1.25.0'
+            'io.opentelemetry.contrib.php.slim',
+            null,
+            'https://opentelemetry.io/schemas/1.25.0'
         );
         /**
          * requires extension >= 1.0.2beta2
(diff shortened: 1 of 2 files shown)


brettmc (approved):

brettmc:
The named params don't work with 8.0. The usage of this package for 8.0 is <1%, so I'm happy for either dropping 8.0 (update composer requirements and drop it from the build matrix), or maintaining support for a bit longer by not using named params.

weslenteche:
> The named params don't work with 8.0. The usage of this package for 8.0 is <1%, so I'm happy for either dropping 8.0 (update composer requirements and drop it from the build matrix), or maintaining support for a bit longer by not using named params.

I opted for the second option, so I removed the named parameters. What do you think ?

brettmc:
> I opted for the second option, so I removed the named parameters.

The change LGTM, however there are now psalm failures.

weslenteche:
> > I opted for the second option, so I removed the named parameters.
> 
> The change LGTM, however there are now psalm failures.

Could you help me resolve the psalm errors?

I tried to add extends to the params template code in [this code snippet](https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/Slim/tests/Integration/SlimInstrumentationTest.php#L167), but without success.

agoallikmaa:
> Could you help me resolve the psalm errors?
> 
> I tried to add extends to the params template code in [this code snippet](https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/Slim/tests/Integration/SlimInstrumentationTest.php#L167), but without success.

Slim added some psalm template annotations in 4.14.0. Trying to replicate the annotations they use internally (`@template-extends`) doesn't seem to be detected by psalm, not sure if it's because of the anonymous declaration, but it can be suppressed with `@psalm-suppress MissingTemplateParam` before that line.

agoallikmaa (approved):

PR #272, opened 2024-06-18
Fix Wordpress DB attributes

jeff-dewan-lumen:
The attribute for the DB name was being incorrectly set to the DB password, resulting in exposed credentials when viewing traces.

[`wp-includes/class-wpdb.php`](https://github.com/WordPress/WordPress/blob/6aa053dc01aeab4ca197091d0c9a4c3437591640/wp-includes/wp-db.php#L700)
```
public function __construct( $dbuser, $dbpassword, $dbname, $dbhost ) {
```

`params[1]` corresponds to `$dbpassword` which should never be exposed in a trace.

Screenshot 2024-06-18 at 1 59 24 PM redacted2


brettmc:
Hi @jeff-dewan-lumen this does look like a bug, can you provide some detail on why you closed the PR?

jeff-dewan-lumen:
Hi @brettmc , sorry but I realized after opening the PR that I am not currently prepared to agree to the EasyCLA terms with my organization right now.

brettmc:
No problem. Could you instead create an issue over in https://github.com/open-telemetry/opentelemetry-php/issues ? Somebody else can provide a fix.

jeff-dewan-lumen:
Will do, thanks!

PR #274, opened 2024-06-27
More signals for cakephp auto instrumentation

eriksamstag:
Closes open-telemetry/opentelemetry-php#1322


bobstrecansky:
@eriksamstag - are you still planning on continuing to work on this?

eriksamstag:
@bobstrecansky Sorry for the delay, I was pretty busy the last weeks. I added the suggestions to the other auto-instrumentation libraries and I think this is ready for review now.

src/Instrumentation/CakePHP/src/CakePHPInstrumentation.php
```diff
@@ -23,60 +16,8 @@ class CakePHPInstrumentation
     public static function register(): void
     {
         $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.cakephp');
```
brettmc, line 18: can you add schema URL? The latest version that will install should be 1.25 (since this package requires php 8.0+)

src/Instrumentation/CakePHP/src/Hooks/Cake/Http/Server.php
```diff
@@ -0,0 +1,199 @@
...
+
+        return [$parts[0], $parts[1]];
+    }
+
+    /**
+     * @param SpanInterface $span
+     * @param array $metricAttributes
+     * @return void
+     */
+    private function recordDuration(SpanInterface $span, array $metricAttributes): void
```
brettmc, line 141: this type of thing is usually left to post-processing by something like https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/spanmetricsconnector/README.md
Also, for a shared-nothing setup (apache, fcgi, etc) I think you're likely to run into https://github.com/open-telemetry/opentelemetry-php/issues/970#issuecomment-1513427738

eriksamstag: Thank you for pointing this out. I wasn't aware of the problems with shared-nothing setups. 

Just to make sure I understand this correctly: You would also create the metrics for request and response body size in post-processing or using something like the upcoming [TraceQL metrics](https://grafana.com/docs/tempo/latest/traceql/metrics-queries/) of Grafana Tempo?

In that case, I will remove all the metrics related code.

brettmc: > Just to make sure I understand this correctly: You would also create the metrics for request and response body size in post-processing or using something like the upcoming [TraceQL metrics queries](https://grafana.com/docs/tempo/latest/traceql/metrics-queries/) of Grafana Tempo?

Yes. If your root span contains those attributes, then you can add more dimensions to the span metrics connector to capture them. Metrics queries, and also https://grafana.com/docs/tempo/latest/metrics-generator/ I think would do the same thing (I believe OTEL's span metrics connector was inspired by Grafana's one)

src/Instrumentation/CakePHP/src/CakePHPInstrumentation.php
ChrisLightfootWild: I think `1.25` is the first on `php: ^8.1`, so we should use `sem-conv: 1.24` here.

eriksamstag: I now changed it to `TraceAttributes::SCHEMA_URL`, the same way it is done in the [PSR-18 auto-instrumentation](https://github.com/open-telemetry/opentelemetry-php-contrib/blob/c701e167156ea78ed54361318224bd00d8340325/src/Instrumentation/Psr18/src/Psr18Instrumentation.php#L30) library. I would think that this way, when installing the CakePHP auto-instrumentation, composer would install the latest compatible version of open-telemetry/sem-conv which would have the constant set to the appropriate schema url. Or am I missing something?

brettmc: That's how it should work in the future, but there is actually a moratorium on changing versions for telemetry: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/telemetry-stability.md#schema-file-driven-telemetry-producers
We're _trying_ to follow that, but it looks like psr-18 didn't get the memo :(

So, in the meantime, we go with the latest version that we can, and we're stuck on that version for a while

weslenteche: The current code in the [master](https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/CakePHP/src/CakePHPInstrumentation.php#L25-L29) already has the value of the URL schema defined. Could you resolve current conflicts?

eriksamstag: > That's how it should work in the future, but there is actually a moratorium on changing versions for telemetry: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/telemetry-stability.md#schema-file-driven-telemetry-producers We're _trying_ to follow that, but it looks like psr-18 didn't get the memo :(
> 
> So, in the meantime, we go with the latest version that we can, and we're stuck on that version for a while

I see, thanks again for the information

brettmc (approved):

ChrisLightfootWild (approved):
Looks good :+1: 

Minor nit is that some blank line end-of-files have been removed; we should probably check the linting for that separately though!

brettmc:
released as v0.0.3

PR #277, opened 2024-07-01
Fix namespace to exclude ApiPlatform test HttpClient from instrumentation

ralfkuehnel:
We ran into the issue described in #235 and I was very happy when I realized that the fix was merged just two days ago! Thanks @cedricziel! But I think there is a small typo in the namespace used to exclude the ApiPlatform test HttpClient from instrumentation. So I created this PR to fix it!


cedricziel (approved):

cedricziel:
Thank you @ralfkuehnel !

brettmc (approved):

brettmc:
merged, and released as a new version

PR #278, opened 2024-07-04
feat: track dbname and other pdo attributes

cedricziel:
This change tracks db name and user / host, if set per PDO object so it gets attached to every related span.

### DIFF
--- a/src/Instrumentation/PDO/src/PDOTracker.php
+++ b/src/Instrumentation/PDO/src/PDOTracker.php
@@ -112,9 +112,9 @@ final class PDOTracker
      * Extracts attributes from a DSN string
      *
      * @param string $dsn
-     * @return array
+     * @return iterable<non-empty-string, bool|int|float|string|array|null>
      */
-    private static function extractAttributesFromDSN(string $dsn): array
+    private static function extractAttributesFromDSN(string $dsn): iterable
     {
         $attributes = [];
         if (str_starts_with($dsn, 'sqlite::memory:')) {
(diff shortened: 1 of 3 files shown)


src/Instrumentation/PDO/src/PDOTracker.php
```diff
@@ -106,4 +107,35 @@ private static function mapDriverNameToAttribute(?string $driverName): string
...
+     * Extracts attributes from a DSN string
+     *
+     * @param string $dsn
+     * @return array
+     */
+    private static function extractAttributesFromDSN(string $dsn): array
+    {
+        $attributes = [];
+        if (str_starts_with($dsn, 'sqlite')) {
+            $attributes[TraceAttributes::DB_NAME] = $dsn;
```
ChrisLightfootWild, line 121: Should the `sqlite:` prefix be dropped, assuming `TraceAttributes::DB_SYSTEM` would be `sqlite` already?

src/Instrumentation/PDO/src/PDOTracker.php
```diff
@@ -106,4 +107,35 @@ private static function mapDriverNameToAttribute(?string $driverName): string
...
+
+            return $attributes;
+        }
+
+        if (preg_match('/user=([^;]*)/', $dsn, $matches)) {
```
brettmc, line 126: should some of these fields be left unset rather than 'unknown' ?

cedricziel: yes! - done

src/Instrumentation/PDO/src/PDOTracker.php
cedricziel: I wanted to avoid an additional extraction dance for the database name - we can probably strip the prefix though

brettmc (approved):

brettmc (approved):

PR #282, opened 2024-07-12
wip: capture spans for controllers automatically

cedricziel:


src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
```diff
@@ -165,5 +167,160 @@ public static function register(): void
...
+                }
+
+                if ($eventName === 'kernel.controller' && \method_exists($event, 'getController')) {
+                    $controller = $event->getController();
+                    if (!($controller instanceof \Closure)) {
+                        if (\is_callable($controller, false, $controllerName) && $controllerName !== null) {
+                            if (\strpos($controllerName, '::') > 0) {
+                                list($class, $method) = \explode('::', $controllerName);
+                                if (isset($class, $method)) {
+                                    hook(
```
brettmc, line 210: I don't think this will work with modern runtimes, where one long-lived application might handle many requests.
Maybe it should just manually create and end span on Symfony events?

cedricziel: I think I'd rather memoize the hooks already in place. - this hook in particular is for the controller callable

PR #284, opened 2024-07-17
ext-rdkafka instrumentation

tom-diacono:
# RdKafka instrumentation

## About RdKafka
[librdkafka](https://github.com/confluentinc/librdkafka) is a C library for interfacing with Kafka, of which PHP has an [extension](https://github.com/arnaud-lb/php-rdkafka) to interface with it. The goal with hooking into this library is that (atleast from what I've read) most higher level PHP Kafka packages are using rdkafka under the hood (enqueue for example), so hopefully instrumenting this will have a good impact in covering many use cases

This auto-instrumentation allows the following to happen:
* A span to be created for each message consumed
  * Done by hooking into `consume` method to start a span, and `commit`/`commitAsync` methods to end the span
* A traceparent to be propagated into this application from Kafka message headers
  * Again, this one is done by hooking into `consume`
* A traceparent header to be set on the outbound message headers, to allow downstream systems to pickup the context
  * Done by hooking into the `producev` method (note that `produce` doesn't support headers, hence this only works with `producev`

PS- This is my first PR in this repository, so not expecting to get everything right first time!


src/Instrumentation/ExtRdKafka/composer.json
```diff
@@ -0,0 +1,50 @@
...
+  "description": "OpenTelemetry auto-instrumentation for psr18 http clients.",
+  "keywords": ["opentelemetry", "otel", "open-telemetry", "tracing", "ext-rdkafka", "instrumentation"],
+  "type": "library",
+  "homepage": "https://opentelemetry.io/docs/php",
+  "readme": "./README.md",
+  "license": "Apache-2.0",
+  "minimum-stability": "dev",
+  "prefer-stable": true,
+  "require": {
+    "php": "^8.0",
```
brettmc, line 12: the tests only run on 8.2+ (because of exclusions added the matrix in .github/workflows/php.yml) - so this should either be 8.2 or the tests should run on 8.0+

src/Instrumentation/ExtRdKafka/README.md
```diff
@@ -0,0 +1,41 @@
...
+  messages.
+    - Note there is an old method called `produce`. This does not support headers, and thus there was no point hooking
+      into this
+- `\RdKafka\KafkaConsumer::commit` / `\RdKafka\KafkaConsumer::commitAsync` - For ending the span when the message offset
+  is committed.
+
+## Versions
+
+* Tested on PHP 8.2 and 8.3 with success
+* PHP 8.1 didn't seem to work
```
brettmc, line 32: adding observers to internal methods (eg from extensions) was introduced in 8.2, so you can safely target only 8.2+

src/Instrumentation/ExtRdKafka/composer.json
brettmc, line 12:
```suggestion
    "php": "^8.2",
```

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
```diff
@@ -0,0 +1,131 @@
...
+                if (!$message instanceof Message || $message->err !== RD_KAFKA_RESP_ERR_NO_ERROR) {
+                    return;
+                }
+
+                $builder = $instrumentation
+                    ->tracer()
+                    // @phan-suppress-next-line PhanTypeSuspiciousStringExpression - Doesn't seem to know this has to be a string
+                    ->spanBuilder('Consumer: ' . $message->topic_name)
+                    ->setSpanKind(SpanKind::KIND_CONSUMER)
+                    ->setAttribute(TraceAttributes::MESSAGING_SYSTEM, 'kafka')
```
brettmc, line 109: there is a const for this in `TraceAttributeValues`

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
```diff
@@ -0,0 +1,131 @@
...
+                    ->setAttribute(TraceAttributes::MESSAGING_OPERATION, 'consume')
```
brettmc, line 110: note that messaging.operation has some well-known values listed in https://opentelemetry.io/docs/specs/semconv/attributes-registry/messaging/#messaging-attributes - maybe `receive` or `process` is more appropriate here?

is there anything else of value in https://opentelemetry.io/docs/specs/semconv/attributes-registry/messaging/#messaging-kafka-attributes that you could add?

src/Instrumentation/Psr15/composer.json
```diff
@@ -34,5 +34,10 @@
     "open-telemetry/sdk": "^1.0",
     "phpunit/phpunit": "^9.5",
     "vimeo/psalm": "^5.0"
+  },
+  "config": {
+    "allow-plugins": {
+      "php-http/discovery": false
+    }
```
brettmc, line 41: can you revert this?

docker-compose.yaml
```diff
@@ -42,3 +43,22 @@ services:
...
+    environment:
+      KAFKA_PROCESS_ROLES: 'broker,controller'
+      KAFKA_NODE_ID: 1
+      KAFKA_ADVERTISED_LISTENERS: ${KAFKA_ADVERTISED_LISTENERS:-PLAINTEXT://kafka:29092,PLAINTEXT_HOST://kafka:9092}
+      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,CONTROLLER:PLAINTEXT,PLAINTEXT_HOST:PLAINTEXT
+      KAFKA_LISTENERS: 'PLAINTEXT://kafka:29092,CONTROLLER://kafka:29093,PLAINTEXT_HOST://0.0.0.0:9092'
+      KAFKA_CONTROLLER_LISTENER_NAMES: 'CONTROLLER'
+      KAFKA_CONTROLLER_QUORUM_VOTERS: '1@kafka:29093'
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
+    command: "bash -c 'if [ ! -f /tmp/update_run.sh ]; then echo \"ERROR: Did you forget the update_run.sh file that came with this docker-compose.yml file?\" && exit 1 ; else /tmp/update_run.sh && /etc/confluent/docker/run ; fi'"
```
brettmc, line 60: can this happen, given it's volume-mounted in?

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
```diff
@@ -0,0 +1,131 @@
...
+                $carrier = [];
+                TraceContextPropagator::getInstance()->inject($carrier);
+                $propagator = Globals::propagator();
+                $propagator->inject($carrier);
+
+                if (array_key_exists('traceparent', $carrier)) {
+                    $params[4]['traceparent'] = $carrier['traceparent'];
+                }
+
+                return $params;
```
weslenteche, line 81: A code suggestion on how the span generated at this moment could be composed when producing a new message.

```suggestion
               $builder = $instrumentation
                   ->tracer()
                   ->spanBuilder('publish') // topicName + publish, example foo publish
                   ->setSpanKind(SpanKind::KIND_PRODUCER)
                   ->setAttribute(TraceAttributes::CODE_FUNCTION, $function)
                   ->setAttribute(TraceAttributes::CODE_NAMESPACE, $class)
                   ->setAttribute(TraceAttributes::CODE_FILEPATH, $filename)
                   ->setAttribute(TraceAttributes::CODE_LINENO, $lineno)
                   ->setAttribute(TraceAttributes::MESSAGING_SYSTEM, 'kafka')
                   ->setAttribute(TraceAttributes::MESSAGING_OPERATION, 'publish')
               $parent = Context::getCurrent();
               
               $span = $builder
                    ->setParent($parent)
                    ->startSpan();
               
               // Kafka message key
               if (array_key_exists(3, $params)) {
                     $span->setAttribute(TraceAttributes::MESSAGING_KAFKA_MESSAGE_KEY, $params[3]);
               }
               
               // Headers are the 5th argument for the producev function
               if (sizeof($params) >= 5 && array_key_exists(4, $params)) {
                   $headers = $params[4];
                   
                   $propagator = Globals::propagator();
                   $propagator->inject($headers, ArrayAccessGetterSetter::getInstance(), $context);
               }
               
                return $params;
```

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
```diff
@@ -0,0 +1,131 @@
...
+
+                $parent = Context::getCurrent();
+                $spanBuilder = $builder
+                    ->setParent($parent);
+
+                if (is_array($message->headers) &&  array_key_exists('traceparent', $message->headers)) {
+                    $propagator = TraceContextPropagator::getInstance();
+                    $otelContext = $propagator->extract(['traceparent' => $message->headers['traceparent']]);
+                    $spanBuilder->setParent($otelContext);
+                }
```
weslenteche, line 121: In this code snippet, other propagators must be extracted in addition to the `TraceContextPropagator`.

```suggestion
                if (is_array($message->headers) {
                    $parent = Globals::propagator()->extract($this->headers);
                    $spanBuilder->setParent($parent);
                }
```

weslenteche (changes requested):

src/Instrumentation/ExtRdKafka/README.md
tom-diacono: That explains a lot, will remove from docs then

docker-compose.yaml
tom-diacono: In terms of is there a chance that the file won't be there? If so, answer is no in honesty! So could just make it run the command without the check if you think that's cleaner?

brettmc: yeah, simpler is better IMO

src/Instrumentation/ExtRdKafka/composer.json
```diff
@@ -0,0 +1,50 @@
...
+    "ext-rdkafka": "*",
+    "ext-opentelemetry": "*",
+    "open-telemetry/api": "^1.0",
+    "open-telemetry/sem-conv": "^1.23",
+    "psr/http-client": "^1"
+  },
+  "require-dev": {
+    "friendsofphp/php-cs-fixer": "^3",
+    "nyholm/psr7": "*",
```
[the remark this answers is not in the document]
tom-diacono: Nope, don't think so, leftover from copy-paste!

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
tom-diacono: Just trying to work this one out (apologies, still getting used to OTEL in PHP and in general, so terminology may be a bit off).

So this adds a span for the actual production of the message (I like it). I can't see it actually activate that span though, guessing I'll want to do that? I'm assuming from there, I'll need to close the span off in the `post` hook right? (which isn't yet there, but fine to add). Otherwise I'll get an _unclosed span_ error (OWTTE)

Also, I see that the `$context` variable is not actually defined, when injecting the context into the headers (appreciate this was probably written on GitHub and not in an IDE). Should I use the same context that I'm using for the parent of this span, currently defined as `$parent`, or use the context of the newly created span (which I'm assuming I'll need to activate prior). Guessing the newly created span, but wanted to make sure.

Cheers :)

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
tom-diacono: Changed the attribute to `process`. Also added a couple more attributes

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
weslenteche: This is correct, in this example that I added as a suggestion, the span should be closed in the `post hook`.
Regarding the written code, I ended up writing directly through Github and there ended up being confusion when declaring the variables, but below is what the `pre hook` would be like.

```php
$builder = $instrumentation
    ->tracer()
    ->spanBuilder('publish') // topicName + publish, example foo publish
    ->setSpanKind(SpanKind::KIND_PRODUCER)
    ->setAttribute(TraceAttributes::CODE_FUNCTION, $function)
    ->setAttribute(TraceAttributes::CODE_NAMESPACE, $class)
    ->setAttribute(TraceAttributes::CODE_FILEPATH, $filename)
    ->setAttribute(TraceAttributes::CODE_LINENO, $lineno)
    ->setAttribute(TraceAttributes::MESSAGING_SYSTEM, 'kafka')
    ->setAttribute(TraceAttributes::MESSAGING_OPERATION, 'publish')

$span = $builder
    ->setParent($parent)
    ->startSpan();

$context = $span->storeInContext($parent);

// Kafka message key
if (array_key_exists(3, $params)) {
    $span->setAttribute(TraceAttributes::MESSAGING_KAFKA_MESSAGE_KEY, $params[3]);
}

// Headers are the 5th argument for the producev function
if (sizeof($params) >= 5 && array_key_exists(4, $params)) {
    $headers = $params[4];
    $propagator = Globals::propagator();
    $propagator->inject($headers, ArrayAccessGetterSetter::getInstance(), $context);
}

Context::storage()->attach($context);

return $params;
```

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
```diff
@@ -0,0 +1,133 @@
...
+                // This is to ensure that there is data. Packages periodically poll this method in order to
+                // determine if there is a message there. If there is not, we don't want to create a span.
+                if (!$message instanceof Message || $message->err !== RD_KAFKA_RESP_ERR_NO_ERROR) {
+                    return;
+                }
+
+                $builder = $instrumentation
+                    ->tracer()
+                    // @phan-suppress-next-line PhanTypeSuspiciousStringExpression - Doesn't seem to know this has to be a string
+                    ->spanBuilder('Consumer: ' . $message->topic_name)
```
weslenteche, line 108: Taking a look at the [semantic convention specification](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#operation-types:~:text=The%20span%20name%20SHOULD%20be%20set%20to%20the%20message%20destination%20name%20and%20the%20name%20of%20the%20operation%20being%20performed%20(as%20captured%20in%20messaging.operation.name)%20in%20the%20following%20format%3A).

> The span name SHOULD be set to the message destination name and the name of the operation being performed (as captured in messaging.operation.name) in the following format:
`<destination name> <operation name>`

> If the operation name is not specified by the messaging system, then the operation type as defined in [Operation types](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#operation-types) SHOULD be used:
`<destination name> <operation type>`

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
```diff
@@ -0,0 +1,133 @@
...
+                if (is_array($message->headers)) {
+                    $parent = Globals::propagator()->extract($message->headers);
+                    $spanBuilder->setParent($parent);
+                }
+
+                $span = $spanBuilder->startSpan();
+
+                $context = $span->storeInContext($parent);
+
+                Context::storage()->attach($context);
```
weslenteche, line 129: Suggestion removing `setParent()` code duplication.

```suggestion
                $parent = Context::getCurrent();
                
                if (is_array($message->headers)) {
                    $parent = Globals::propagator()->extract($message->headers);
                }
                
                $spanBuilder->setParent($parent);
                $span = $spanBuilder->startSpan();

                $context = $span->storeInContext($parent);
                Context::storage()->attach($context);
```

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
weslenteche: I'm sorry, now your initial approach is clear to me. See if the code below helps you with this.

```php
$carrier = [];
$propagator = Globals::propagator();
$propagator->inject($carrier);

// Headers are the 5th argument for the producev function
$params[4] = isset($params[4]) ? array_merge($params[4], $carrier) : $carrier;

return $params;
```

tom-diacono: All good, I've ended up going for the first one anyway. So it's now creating a span for the lifecycle for the produce method, and still works end to end.

Had to modify the tests to expect more spans, but that's all fine, all passing

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
tom-diacono: Good point! I've made it be `topic_name operation_type`. This matches the semconv and to sanity check it, the Java one aswell

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
tom-diacono: Just dug into this a bit. I changed it around a bit to use an `if-else`. When I did that, I got an thing from `phpstan` saying that the docblock said it would never be null.....

Did some testing (I've got this hooked into a local repo), and rdkafka does indeed give `null` as headers when there's nothing there (not empty array or anything). This is in contrast to the `php-librdkafka` stubs, which say a string.

For this reason, I've left the if statement in, and put an `ignore` on that line for `phpstan`. Not ideal, but I think that's the only way

src/Instrumentation/ExtRdKafka/src/ExtRdKafkaInstrumentation.php
```diff
@@ -0,0 +1,181 @@
...
+                array $params,
+                string $class,
+                string $function,
+                ?string $filename,
+                ?int $lineno
+            ) use ($instrumentation) : array {
+                /** @var CachedInstrumentation $instrumentation */
+                $builder = $instrumentation
+                    ->tracer()
+                    ->spanBuilder('publish') // topicName + publish, example foo publish
```
weslenteche, line 78: Would it be possible to incorporate the span value with the name of the topic and operation as recommended by the semantic convention?

weslenteche (approved):

src/Instrumentation/ExtRdKafka/tests/Integration/ExtRdKafkaInstrumentationTest.php
```diff
@@ -0,0 +1,202 @@
...
+        // The traceparent header is separated into 4 sections with dashes. I want to get the second segment of that
+        // header, which is the TraceID
+        $traceId = explode('-', $message->headers['traceparent'])[1];
+        $this->assertEquals($span->getContext()->getTraceId(), $traceId);
+    }
+
+    public function test_produce_creates_new_span()
+    {
+        $this->assertCount(0, $this->storage);
+        // Overriding topic as all other tests consume their own message. If we don't do this, the message will be
```
brettmc, line 122: this sounds like a post-test cleanup/teardown issue - could a `tearDown` method empty all queues?

tom-diacono: Yep, it is exactly that. Had a bit of a look around yesterday, and unfortunately the `ext-rdkafka` extension [doesn't have support for the AdminClient](https://github.com/arnaud-lb/php-rdkafka/issues/215) (the thing used to manage topics) yet.

So from what I can see, there's no way that we can manually remove the topic (or messages, as deletion is handled by the Kafka broker) after a given test run- atleast via PHP. Open to any other ideas, but without having that element of control via the broker, I'm struggling to come up with something better

brettmc: can you consume and discard any remaining messages from all queues as part of teardown?

tom-diacono: Yeah that will also work, leaves it in a slightly better state than what we've currently got IMO.... Unfortunate that we can't make these tests properly atomic, but I think that's the best we'll get.

Will push up a change asap

src/Instrumentation/ExtRdKafka/tests/Integration/ExtRdKafkaInstrumentationTest.php
```diff
@@ -0,0 +1,218 @@
...
+                break;
+            }
+        }
+    }
+
+    /**
+     * @psalm-suppress InvalidNullableReturnType
+     * @return int|Message
+     *
+     * Couldn't use union types on method as psalm wants php 7.4 compatible code. For some reason, it also thinks
```
brettmc, line 170: that doesn't sound right - is it somehow looking at the top-level composer.json to determine php version? if all else fails, you can force it in the local psalm.xml.dist: https://psalm.dev/docs/running_psalm/configuration/#phpversion

src/Instrumentation/ExtRdKafka/tests/Integration/ExtRdKafkaInstrumentationTest.php
```diff
@@ -0,0 +1,218 @@
...
+            ->withPropagator(new TraceContextPropagator())
+            ->activate();
+    }
+
+    public function tearDown(): void
+    {
+        $this->scope->detach();
+    }
+
+    public static function tearDownAfterClass(): void
```
brettmc, line 48: it should probably do this after every test. I'm not sure if we're doing it in this repo yet, but over in the main repo we run all tests in random order to shake out any inter-test dependencies.

src/Instrumentation/ExtRdKafka/tests/Integration/ExtRdKafkaInstrumentationTest.php
tom-diacono: Found it in the logs- I wasn't specifying PHP Version when doing the `make all`- as tests didn't require it. Fixed now

src/Instrumentation/ExtRdKafka/tests/Integration/ExtRdKafkaInstrumentationTest.php
tom-diacono: Good point- toyed with the idea but was worried about the overhead (these tests already take a little while to run). After a test, it's not really noticeable speed-wise. Have changed it

brettmc (approved):

brettmc:
https://packagist.org/packages/open-telemetry/opentelemetry-auto-ext-rdkafka

PR #290, opened 2024-09-03
Fix span status for 4XX responses

alecsammon:
For servers then Span Status should only be set to `Error` for responses >= 500, and 400 should be left unset.

See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md?plain=1#L108-L112


brettmc (approved):

ChrisLightfootWild (approved):

alecsammon:
❤️  - thanks all!

PR #291, opened 2024-09-04
Fix: ignore redirection error

MilesChou:
According to the definition in [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-redirection-3xx):

> The 3xx (Redirection) class of status code indicates that further action needs to be taken by the user agent in order to fulfill the request.

The 3xx status code only signifies that additional steps are required to determine the result, but merely encountering a 3xx status code does not necessarily indicate an error. Therefore, it is recommended to make this adjustment.


MilesChou:
EasyCLA is submitted.

ChrisLightfootWild (approved):

PR #292, opened 2024-09-05
Feat: Add contextual attributes register for queue

MilesChou:
I have been using the package [vladimir-yuldashev/laravel-queue-rabbitmq](https://github.com/vyuldashev/laravel-queue-rabbitmq) in my work. However, I noticed that the auto Laravel feature only handles Laravel's native queues, without any interface for developers to add custom queue parsing.

This PR introduces an interface that allows the registration of custom queues. By creating a class that extends `AbstractContextualAttributesBuilder` and registering it with `AttributesBuilderRegister` , it will automatically be recognized and processed during the hook phase.

### DIFF
--- a/src/Instrumentation/Laravel/src/Hooks/Illuminate/Queue/AttributesBuilder.php
+++ b/src/Instrumentation/Laravel/src/Hooks/Illuminate/Queue/AttributesBuilder.php
@@ -5,11 +5,7 @@ declare(strict_types=1);
 namespace OpenTelemetry\Contrib\Instrumentation\Laravel\Hooks\Illuminate\Queue;
 
 use Illuminate\Contracts\Queue\Queue as QueueContract;
-use Illuminate\Queue\BeanstalkdQueue;
-use Illuminate\Queue\RedisQueue;
-use Illuminate\Queue\SqsQueue;
 use OpenTelemetry\SemConv\TraceAttributes;
-use OpenTelemetry\SemConv\TraceAttributeValues;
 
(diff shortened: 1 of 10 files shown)


ChrisLightfootWild:
> I have been using the package [vladimir-yuldashev/laravel-queue-rabbitmq](https://github.com/vyuldashev/laravel-queue-rabbitmq) in my work. However, I noticed that the auto Laravel feature only handles Laravel's native queues, without any interface for developers to add custom queue parsing.
> 
My initial thinking around this is that it would be less automatic at that point, if developers have to do additional wiring in their own service providers.

> This PR introduces an interface that allows the registration of custom queues. By creating a class that extends `AbstractContextualAttributesBuilder` and registering it with `AttributesBuilderRegister` , it will automatically be recognized and processed during the hook phase.

I'm not totally against this but would potentially like to see this as a feature that waits until we could target a `1.1` release with support for this to be provided by the newer SPI mechanism. See https://github.com/open-telemetry/opentelemetry-php/pull/1304 and #269

MilesChou:
> I'm not totally against this but would potentially like to see this as a feature that waits until we could target a `1.1` release with support for this to be provided by the newer SPI mechanism. See [open-telemetry/opentelemetry-php#1304](https://github.com/open-telemetry/opentelemetry-php/pull/1304) and #269
  
I understand, and I’ll go ahead and close this PR. If there's any testing or help needed when targeting the 1.1 release or implementing this feature with the newer SPI mechanism, I’d be happy to assist.

PR #294, opened 2024-09-16
Use Coroutine::list instead of Coroutine::listCoroutines

thomasbeaujean:
Since OpenSwoole: 4.4.0, 
the method Coroutine::list is available: https://openswoole.com/docs/modules/swoole-coroutine

In the 22 version, the method Coroutine::listCoroutines does not exist.

### DIFF
--- a/src/Context/Swoole/src/SwooleContextHandler.php
+++ b/src/Context/Swoole/src/SwooleContextHandler.php
@@ -41,7 +41,7 @@ final class SwooleContextHandler
     public function splitOffChildCoroutines(): void
     {
         $pcid = Coroutine::getCid();
-        foreach (Coroutine::listCoroutines() as $cid) {
+        foreach (Coroutine::list() as $cid) {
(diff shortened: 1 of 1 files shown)


src/Context/Swoole/src/SwooleContextHandler.php
```diff
@@ -41,7 +41,7 @@ public function switchToActiveCoroutine(): void
     public function splitOffChildCoroutines(): void
     {
         $pcid = Coroutine::getCid();
-        foreach (Coroutine::listCoroutines() as $cid) {
+        foreach (Coroutine::list() as $cid) {
```
brettmc, line 44: Do we need to take BC into account here?
```suggestion
        foreach (method_exists(Coroutine::class, 'list') ? Coroutine::list() : Coroutine::listCoroutines() as $cid) {
```

brettmc (approved):

PR #295, opened 2024-09-18
[Symfony] Fix: Set span status according to otel convention

technimad-splunk:
The span status was always set to error for responses with status code 400 and up.  
According to the otel semantic conventions, for 4xx status, the span status MUST be left unset for spans with SpanKind.SERVER This commit adds the logic to implement this behaviour.

Ref: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status

### DIFF
--- a/src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
@@ -113,7 +113,12 @@ final class SymfonyInstrumentation
                     return;
                 }
 
-                if ($response->getStatusCode() >= Response::HTTP_BAD_REQUEST) {
+                // Do not set error status on server spans for 4xx responses
+                if (!$span->getKind(SpanKind::KIND_SERVER) && $response->getStatusCode() >= Response::HTTP_BAD_REQUEST) {
+                    $span->setStatus(StatusCode::STATUS_ERROR);
+                }
+                // Do set error status on all spans for 5xx responses
+                if ($response->getStatusCode() >= Response::HTTP_INTERNAL_SERVER_ERROR) {
                     $span->setStatus(StatusCode::STATUS_ERROR);
                 }
                 $span->setAttribute(TraceAttributes::HTTP_RESPONSE_STATUS_CODE, $response->getStatusCode());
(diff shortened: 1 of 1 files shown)


technimad-splunk:
PR #290 fixed a similar issue for Laravel.

src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
```diff
@@ -113,7 +113,12 @@ public static function register(): void
                     return;
                 }
 
-                if ($response->getStatusCode() >= Response::HTTP_BAD_REQUEST) {
+                // Do not set error status on server spans for 4xx responses
+                if (!$span->getKind(SpanKind::KIND_SERVER) && $response->getStatusCode() >= Response::HTTP_BAD_REQUEST) {
```
agoallikmaa, line 117: This only seems possible if this is a subrequest (internal redirect) in which case span kind is `INTERNAL`. I assume it was an intentional choice to use error status specifically for 400 codes from internal redirects, as the specification does not cover this scenario at all. Since you already have a comment here, perhaps it could mention that this case is for subrequests/internal spans.

agoallikmaa (approved):

src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
technimad-splunk: Good point. This case can only happen with span kind Internal. The intention was to have this case mainly for client span kinds. Which can't happen here. 
I will remove this case, to ONLY set error status on >= 500 status codes. This is the same behaviour as implemented in Laravel.

brettmc: looks ok to me, can it be tested by adding/modifying our existing tests?

mvanduijker:
We would love to see this merged (and released) how can I help to make this happen?

brettmc:
@mvanduijker merged. This package has been in beta for a long time, I think it's ready for a 1.0 release? Perhaps you could test it out and report back?

mvanduijker:
we gonna migrate to opentelemetry, with a pretty big project, this month. I can report back how it behaves.

PR #297, opened 2024-10-02
psr3: handle non-string attribute keys

brettmc:
non-string attribute keys cause an error, and stop log messages from being exported. cast the numeric keys to a string to ensure they make it through alive. update tests, skipping cake because it does weird stuff when presented with a numeric key which completely changes the structure of the incoming context.

Closes: https://github.com/open-telemetry/opentelemetry-php/issues/1374

### DIFF
--- a/src/Instrumentation/Psr3/src/Formatter.php
+++ b/src/Instrumentation/Psr3/src/Formatter.php
@@ -10,21 +10,21 @@ class Formatter
 {
     public static function format(array $context): array
     {
-        $return = [];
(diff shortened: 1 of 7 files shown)


ChrisLightfootWild (approved):

agoallikmaa (approved):

bobstrecansky (commented):
I noticed that we had a couple failures here related to PHPCS and Phan in this PR.  If those are expected, feel free to merge at will.

PR #300, opened 2024-10-03
Add support for Doctrine auto-instrumentation

DominicDetta:


brettmc:
@DominicDetta one more thing to add is an entry to the top-level `.gitsplit.yaml`. When this is merged, that'll be used to split the package into a read-only repository which we can point packagist at.
You'll also need to add an entry to .github/workflows/php.yml so that the tests will run.

DominicDetta:
@brettmc I included the Doctrine directory in the files you specified

DominicDetta:
I agreed the EasyCLA and waiting for the merge approval

brettmc:
Can you also update workflows/php to exclude 7.4 from the version matrix? The php version requirement is ^8.2, but I think it would work back to 8.0 since it doesn't hook internal functions. Either way, we need to either drop the requirement back to 8.0 or exclude pre-8.2 versions from the test matrix.

DominicDetta:
it's done

DominicDetta:
In the end I integrated again the PHP 7.4 version and excluded Doctrine from the pre-8.2 versions tests

brettmc:
Green build is an excellent start. I'm happy to push this out as a 0.0.1 release. Since this works against doctrine 3 (according to the composer.json), and doesn't hook any internal/extension functions, I think it would probably work in 8.0 and 8.1 as well. But, that's not a blocker to getting this out if you're happy with the PR as it is.

DominicDetta:
I'm pretty confident it works even with PHP 8.0 version.
I can confirm that the classes `\Doctrine\DBAL\Driver\Connection` and `\Doctrine\DBAL\Driver` exist even in the next major versions of doctrine. (I'm using doctrine 3 at the moment). I'm glad I was able to contribute.

Nevay (commented):
Hooking driver methods will cause duplicated spans if a middleware like [Symfony Debug Middleware](https://github.com/symfony/doctrine-bridge/tree/3e223cc6869b8677bd74ba508eca77a2e87b9e4d/Middleware/Debug) is used.
An alternative approach to avoid this problem would be to move the implementation into a middleware and use a hook to inject this middleware into every connection; see also [Nevay/otel-instrumentation-doctrine-dbal](https://github.com/Nevay/otel-instrumentation-doctrine-dbal/blob/main/src/DoctrineInstrumentation.php).

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
```diff
@@ -0,0 +1,186 @@
...
+                Context::storage()->attach($span->storeInContext($parent));
+            },
+            post: static function (\Doctrine\DBAL\Driver\Connection $connection, array $params, mixed $statement, ?Throwable $exception) {
+                self::end($exception);
+            }
+        );
+
+        hook(
+            \Doctrine\DBAL\Driver\Connection::class,
+            'beginTransaction',
```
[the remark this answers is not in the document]
DominicDetta: Do you have an idea how to solve it because sincerely I dont know how to approach it.  When I looked up your repository I dont see any changes. Anyway I dont think it's a problem to be worry about.

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
```diff
@@ -0,0 +1,186 @@
...
+            'connect',
+            pre: static function (\Doctrine\DBAL\Driver $driver, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                /** @psalm-suppress ArgumentTypeCoercion */
+                $builder = self::makeBuilder($instrumentation, 'Doctrine\DBAL\Driver::connect', $function, $class, $filename, $lineno)
+                    ->setSpanKind(SpanKind::KIND_CLIENT);
+                $builder
+                    ->setAttribute(TraceAttributes::SERVER_ADDRESS, $params[0]['host'] ?? 'unknown')
+                    ->setAttribute(TraceAttributes::SERVER_PORT, $params[0]['port'] ?? 'unknown')
+                    ->setAttribute(TraceAttributes::DB_SYSTEM, $params[0]['driver'] ?? 'unknown')
+                    ->setAttribute(TraceAttributes::DB_NAMESPACE, $params[0]['dbname'] ?? 'unknown');
```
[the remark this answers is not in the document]
DominicDetta: I prefer to use the default value already used by the [official pdo](https://github.com/opentelemetry-php/contrib-auto-pdo/blob/main/src/PDOInstrumentation.php) at the line 42

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
```diff
@@ -0,0 +1,186 @@
...
+        );
+
+        hook(
+            \Doctrine\DBAL\Driver\Connection::class,
+            'query',
+            pre: static function (\Doctrine\DBAL\Driver\Connection $connection, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                /** @psalm-suppress ArgumentTypeCoercion */
+                $builder = self::makeBuilder($instrumentation, 'Doctrine\DBAL\Driver\Connection::query', $function, $class, $filename, $lineno)
+                    ->setSpanKind(SpanKind::KIND_CLIENT);
+                $builder->setAttribute(TraceAttributes::DB_QUERY_TEXT, $params[0] ?? 'undefined');
```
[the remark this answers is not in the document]
DominicDetta: I would not use this approach for these reasons: 
- the official pdo repository does not do any sanitize
- the https://github.com/phpmyadmin/sql-parser library you propose focuses on Mysql syntax
- The target applications should use the prepared statements and therefore the problem should not arise
- The convention indicates that it is a SHOULD rule

dkarlovi:
> Does OpenTelemetry have a feature to configure and filter duplicated spans?

AFAIK the idea is to allow the agent to do that:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/filterprocessor/README.md

src/Instrumentation/Doctrine/_register.php
```diff
@@ -0,0 +1,18 @@
...
+declare(strict_types=1);
+
+use OpenTelemetry\Contrib\Instrumentation\Doctrine\DoctrineInstrumentation;
+use OpenTelemetry\SDK\Sdk;
+
+if (class_exists(Sdk::class) && Sdk::isInstrumentationDisabled(DoctrineInstrumentation::NAME) === true) {
+    return;
+}
+
+if (extension_loaded('opentelemetry') === false) {
```
ostrolucky, line 12: Is this necessary, considering you have ext-opentelemetry as a requirement in composer.json? If you want to allow project to still function without having the extension loaded, you should then also make it as optional dependency in composer.json.

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
```diff
@@ -0,0 +1,182 @@
...
+use OpenTelemetry\API\Trace\SpanKind;
+use OpenTelemetry\API\Trace\StatusCode;
+use OpenTelemetry\Context\Context;
+
+use function OpenTelemetry\Instrumentation\hook;
+
+use OpenTelemetry\SemConv\TraceAttributes;
+use Throwable;
+
+class DoctrineInstrumentation
```
ostrolucky, line 19: You should call this DBAL instrumentation. Possibly whole package as well. Have you thought about Doctrine ORM? Is that going to be a separate package, or part of this package? If the former, this package should be renamed. If the latter, code should be structured in a way to be ready for future extension to cover doctrine ORM. I'm pretty sure it will be important because object hydration is often the bottleneck.

src/Instrumentation/Doctrine/_register.php
brettmc: We've got this test in all auto-instrumentations, because folks will often install with composer's `--ignore-platform-reqs` (reported usages have been for CI pipelines, or a multi-developer project where not all developers care about auto-instrumentation/telemetry). We're open to revisiting this, but AFAIR those are the edge cases to consider.

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
cedricziel: Well I think it's fine to have this umbrella-instrumentation. I'd be happy to add some more ORM instrumentation later, but I think it makes sense to stick with this naming as ORM depends on DBAL

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.doctrine');
+
+        hook(
+            \Doctrine\DBAL\Driver::class,
+            'connect',
+            pre: static function (\Doctrine\DBAL\Driver $driver, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                /** @psalm-suppress ArgumentTypeCoercion */
+                $builder = self::makeBuilder($instrumentation, 'Doctrine\DBAL\Driver::connect', $function, $class, $filename, $lineno)
+                    ->setSpanKind(SpanKind::KIND_CLIENT);
```
cedricziel, line 33: How do we think about the duality of PDO and dbal here? Most instrumented methods have a pendant in the PDO instrumentation i.e. - a compatible pdo driver will cause two client spans and the hierarchy will probably be

```
my_func (internal)
 -> connect (client) # from dbal
   -> PDO::__construct (client ) # from pdo
```

IMO the two should be either mutually exclusive or enrich each other. WDYT?

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
```diff
@@ -0,0 +1,182 @@
...
+                self::end($exception);
+            }
+        );
+
+        hook(
+            \Doctrine\DBAL\Driver\Connection::class,
+            'query',
+            pre: static function (\Doctrine\DBAL\Driver\Connection $connection, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                /** @psalm-suppress ArgumentTypeCoercion */
+                $builder = self::makeBuilder($instrumentation, 'Doctrine\DBAL\Driver\Connection::query', $function, $class, $filename, $lineno)
```
cedricziel, line 53: The recommended span name here would be something like `SELECT my_table`.

The duplication with the pdo instrumentation is the same here.

https://opentelemetry.io/docs/specs/semconv/database/database-spans/#name

cedricziel (changes requested):
My proposal is to figure out if this instrumentation should depend on the existence of PDO and then define how the two interact. My understanding is that _if_ a PDO is used and the PDO instrumentation is active, there will be 2 nested CLIENT spans for the same interaction. - IMHO this is severely impacting the user experience.

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
brettmc: We do not have a way for two auto-instrumentations to cooperate like this (it's an idea that has come up before, and would be cool). So I think mutually exclusive is the best we have - or at least it's up to the user to install only one instrumentation, I don't think we need to go as far as setting them up as conflicting in composer.
Is documenting that recommendation good enough for now?

cedricziel: Could the PDO instrumentation re-write the doctrine span from CLIENT to internal or bail if it recognizes the parent is already client?

cedricziel: To your point - i think documenting it makes the experience worse because users then need to make a whole lot of decisions when instrumenting

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
ostrolucky: Ok. Just a note though: doctrine/orm depends on doctrine/dbal, but not doctrine/mongodb-odm

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
brettmc: > i think documenting it makes the experience worse

I only meant documenting such as "use pdo or doctrine auto-instrumentation, but probably not both" - I expect users to make decisions about which auto-instrumentations to install based on their workload/stack...just installing everything available is probably too noisy.

> Could the PDO instrumentation re-write the doctrine span from CLIENT to internal or bail if it recognizes the parent is already client?

I think the issue here would be that pre/post hooks operate in isolation, so there's currently no way for a post hook to know whether the pre hook created a span or just modified the active span. All the implementations we have assume that `pre` created and activated a span, and `post` just closes whichever is the active span at the time.
We have previously brainstormed whether we could manage some state between a pre and post hook, but we thought it was going to be a hard problem.

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
DominicDetta: > You should call this DBAL instrumentation. Possibly whole package as well. Have you thought about Doctrine ORM? Is that going to be a separate package, or part of this package? If the former, this package should be renamed. If the latter, code should be structured in a way to be ready for future extension to cover doctrine ORM. I'm pretty sure it will be important because object hydration is often the bottleneck.

Right. I didn't think at the possibility to monitor Doctrine ORM Hydration process., only the database invocations.




> Well I think it's fine to have this umbrella-instrumentation. I'd be happy to add some more ORM instrumentation later, but I think it makes sense to stick with this naming as ORM depends on DBAL

I really like that idea. I will maintain the current nomenclature

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
DominicDetta: Ok. Do you know a good agnostic SQL parser to achieve this? The [one I found ](https://github.com/phpmyadmin/sql-parser) is focusing on Mysql syntax.

dkarlovi: Doctrine already knows the table name(s).

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
DominicDetta: > I expect users to make decisions about which auto-instrumentations to install based on their workload/stack...just installing everything available is probably too noisy.

I completely agree on this matter.

Nevay: > Could the PDO instrumentation ... bail if it recognizes the parent is already client?

Bailing out requires storing additional data within the `Context` to be able to detect duplicate instrumentations, see Javas span suppression strategies:
[Instrumentation span suppression behavior](https://opentelemetry.io/docs/zero-code/java/agent/disable/#instrumentation-span-suppression-behavior)
[SpanSuppressionStrategy.java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressionStrategy.java)
[SpanSuppressors.java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressors.java)

> AFAIK the idea is to allow the agent to do that

Dropping at the agent is too late as it may lead to broken traces "Dropping a span may lead to orphaned spans if the dropped span is a parent." [ref](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/filterprocessor/README.md#warnings).

> there's currently no way for a post hook to know whether the pre hook created a span or just modified the active span

Scopes attached via `Context::storage()->attach()` implement `ArrayAccess` to allow propagating state from `pre` hook to `post` hook [ref](https://github.com/open-telemetry/opentelemetry-php/pull/675).

```php
hook(
    null,
    'example',
    static function() use ($tracer): void {
        $context = Context::getCurrent();
        if (lcg_value() > .5 /* some suppression condition */) {
            $span = $tracer
                ->spanBuilder('example')
                ->startSpan();
            $context = $span->storeInContext($context);
        }
        $scope = Context::storage()->attach($context);
        $scope[SpanInterface::class] = $span ?? null;
    },
    static function(): void {
        if (!$scope = Context::storage()->scope()) {
            return;
        }
        $scope->detach();
        /** @var SpanInterface|null $span */
        $span = $scope[SpanInterface::class] ?? null;
        $span?->end();
    },
);
```

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
DominicDetta: > Doctrine already knows the table name(s).

With the approach I'm following right now, I dont have this information. Are you referring to Doctrine ORM? I'm creating hooks on the methods of `\Doctrine\DBAL\Driver` and `\Doctrine\DBAL\Driver\Connection`

dkarlovi: I see, that's a good point.

The issue with this approach is whichever parser you choose is bound to be quite slow and will probably be slowest part of the call (which includes the actual query), making it unusable for production use. ORM itself has a DQL query parser (which is close enough for high level comparison) and it's basically unusable in prod without it being cached with [ORM query cache](https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/caching.html#query-cache).

Running a random query parser on every DB query is unlikely to be viable.

dkarlovi: IMO we could consider the ORM instrumentation even if it's still not here yet.

For example, since ORM will know what the table names are, you could assume they will be put in context by the ORM instrumentation and then only try to detect it if not there? This would future proof this implementation and allow extensions when they happen.

DominicDetta: Sorry for the late response but I'm pretty busy at the moment. 



> IMO we could consider the ORM instrumentation even if it's still not here yet.
> 
> For example, since ORM will know what the table names are, you could assume they will be put in context by the ORM instrumentation and then only try to detect it if not there? This would future proof this implementation and allow extensions when they happen.

I see your point but my focus is to implement hooks for Doctrine DBAL and in the future time permitted we could refactor the functions to behave as you suggested.

src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
LionsAd: @brettmc Can you post-process the spans and just remove / enrich the duplicated ones?

e.g. Is there something that runs before the exporter?

brettmc:
@DominicDetta I didn't notice that you'd signed the CLA now. Let's get this merged, can you merge in main and resolve conflicts?

DominicDetta:
@brettmc I resolved the conflict, I didnt have time to put more effort in this project. I'm too busy at the moment.

DominicDetta:
@brettmc  Maybe you didnt see but I merged to the main branch

.github/workflows/php.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           'Instrumentation/Slim',
           'Instrumentation/Symfony',
           'Instrumentation/Yii',
+          'Instrumentation/Doctrine',
```
ChrisLightfootWild, line 48: Please can you move this into alphabetical order?

.github/workflows/php.yml
```diff
@@ -67,6 +68,8 @@ jobs:
             php-version: 8.1
           - project: 'Instrumentation/ExtRdKafka'
             php-version: 8.1
+          - project: 'Instrumentation/Doctrine'
+            php-version: 8.1
```
ChrisLightfootWild, line 72: As above.

.github/workflows/php.yml
```diff
@@ -157,4 +160,4 @@ jobs:
         directory: src/${{ matrix.project }}
         files: ./coverage.clover
         flags: ${{ matrix.project }}
-        verbose: false
+        verbose: false
```
ChrisLightfootWild, line 163: Revert to newline EOF.

ChrisLightfootWild (commented):
Just a couple of tidy-ups from me 👍

brettmc (approved):

brettmc:
https://packagist.org/packages/open-telemetry/opentelemetry-auto-doctrine

PR #301, opened 2024-10-03
[contrib-aws] Check for presence of @metadata element in Aws\Result object before accessing within mapResult instrumentation hook.

gwharton:
Added check to Instrumentation mapResult hook to check for presence of @metadata element in the Aws\Result object, as @metadata is not always set for all instrumented functions.

fixes https://github.com/open-telemetry/opentelemetry-php/issues/1390


bobstrecansky (approved):
Thank you for this bug fix!

brettmc (approved):

PR #303, opened 2024-10-05
Add peer.service attribute to psr18 spans

caetanev:
The peer.service attribute helps to relate traces between services. The attribute is defined in the semantic conventions: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/peer.md?plain=1#L15

Add the peer.service attribute to the Psr18 spans that request known services.

### DIFF
--- a/src/Instrumentation/Psr18/src/Psr18Instrumentation.php
+++ b/src/Instrumentation/Psr18/src/Psr18Instrumentation.php
@@ -49,6 +49,7 @@ class Psr18Instrumentation
                     ->spanBuilder(sprintf('%s', $request->getMethod()))
                     ->setParent($parentContext)
                     ->setSpanKind(SpanKind::KIND_CLIENT)
+                    ->setAttribute(TraceAttributes::PEER_SERVICE, $request->getUri()->getHost())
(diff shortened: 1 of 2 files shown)


brettmc:
> SHOULD be equal to the actual `service.name` resource attribute of the remote service if any.

This change would apply to all http requests, which may or may not have a `service.name`. What are other SIGs doing here?

bobstrecansky (approved):

bobstrecansky (commented):
We should follow the Java SIG pattern here.

bobstrecansky:
Removed my review - inadvertently approved.

brettmc:
I think Java's approach is a great one to follow - and I also remember now that it's supported in yaml-based config: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/kitchen-sink.yaml#L384

> This looks like it would suffer the same problem but was recently accepted: https://github.com/open-telemetry/opentelemetry-php-contrib/pull/302

Ugh, sorry I didn't think about that during review. Once we get a pattern we're happy with in this PR, can apply the same to Symfony...

caetanev:
I will close this PR and work towards replicating the Java SIG solution.

PR #304, opened 2024-10-11
Update Symfony Messenger instrumentation to use SpanKind::KIND_PRODUCER, SpanKind::KIND_CONSUMER #1314

RichardChukwu:
This implements instrumentation for the Symfony Messenger component in the OpenTelemetry PHP SDK. It fixes the spanKind label to follow https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-kind for both the dispatch and send methods, allowing for better tracing and monitoring of message handling within Symfony applications.

This PR addresses Issue [](https://github.com/open-telemetry/opentelemetry-php/issues/1314) regarding the need for appropriate span kinds and attributes for message processing in the Symfony Messenger. @brettmc 

Closes: https://github.com/open-telemetry/opentelemetry-php/issues/1314

### DIFF
--- a/src/Instrumentation/Symfony/src/MessengerInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/MessengerInstrumentation.php
@@ -125,7 +125,7 @@ final class MessengerInstrumentation
                 $builder = $instrumentation
                     ->tracer()
                     ->spanBuilder(\sprintf('SEND %s', $messageClass))
-                    ->setSpanKind(SpanKind::KIND_CONSUMER)
+                    ->setSpanKind(SpanKind::KIND_PRODUCER)
                     ->setAttribute(TraceAttributes::CODE_FUNCTION, $function)
                     ->setAttribute(TraceAttributes::CODE_NAMESPACE, $class)
                     ->setAttribute(TraceAttributes::CODE_FILEPATH, $filename)
--- a/src/Instrumentation/Symfony/src/MessengerInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/MessengerInstrumentation.php
@@ -58,7 +58,7 @@ final class MessengerInstrumentation
                 $builder = $instrumentation
                     ->tracer()
                     ->spanBuilder(\sprintf('DISPATCH %s', $messageClass))
-                    ->setSpanKind(SpanKind::KIND_INTERNAL)
+                    ->setSpanKind(SpanKind::KIND_PRODUCER)
                     ->setAttribute(TraceAttributes::CODE_FUNCTION, $function)
                     ->setAttribute(TraceAttributes::CODE_NAMESPACE, $class)
                     ->setAttribute(TraceAttributes::CODE_FILEPATH, $filename)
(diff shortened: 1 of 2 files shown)


src/Instrumentation/Symfony/src/MessengerInstrumentation.php
```diff
@@ -125,7 +125,7 @@ public static function register(): void
                 $builder = $instrumentation
                     ->tracer()
                     ->spanBuilder(\sprintf('SEND %s', $messageClass))
-                    ->setSpanKind(SpanKind::KIND_INTERNAL)
```
brettmc, line 128: I'm no expert here, but aren't dispatch and send roughly the same thing (ie, outbound messages)? I would have thought that KIND_CONSUMER would be for receiving messages?

RichardChukwu: Thank you for the feedback, @brettmc ! You’re right that both dispatch and send in Symfony Messenger handle outbound messages.

Regarding the use of KIND_CONSUMER, you bring up a good point. In OpenTelemetry's messaging semantic conventions, KIND_PRODUCER is typically used for the sender (dispatch or send), and KIND_CONSUMER is used for receiving messages.

Given that this instrumentation focuses on tracing the message dispatch and sending processes (outbound), I guess the correct span kind would be KIND_PRODUCER for both events right?.

RichardChukwu:
Friendly ping on this @brettmc good to go?

src/Instrumentation/Symfony/src/MessengerInstrumentation.php
brettmc: Is this resolved? It's still KIND_CONSUMER.

RichardChukwu: Oh, surprised that didn't update, thought I pushed already. Updated now

brettmc: I think you need to update a test as well...

RichardChukwu: Test updated

RichardChukwu:
@brettmc thank you for the feedback.

Any further updates please or is this good to go?

brettmc (approved):

RichardChukwu:
Noticed some CI tests failures though, but are unrelated to this issue.🤔

I may be wrong, but I'm guessing this can be merged and those looked at separately, right? @brettmc

brettmc:
> I'm guessing this can be merged and those looked at separately

Yes, those failures are unrelated and won't affect this PR.

cedricziel:
I agree with the sentiment of this PR - I introduced the INTERNAL spans, because I added instrumentation for ext-amqp in addition to this one that would create the relevant PRODUCER spans. I wonder if this PR should also address this possible span duplication by modifying the ext-amqp instrumentation to enrich the spans created here?

RichardChukwu:
Thank you for your feedback! @cedricziel 

I agree that modifying the ext-amqp instrumentation to enrich the spans created here would be a valuable enhancement to avoid span duplication. 

However, I believe this could be addressed as a separate issue to keep the scope of this PR focused on resolving the KIND_PRODUCER instrumentation for dispatch and send.

I’m happy to open a new issue to track this work and contribute to implementing it once this PR is merged. Let me know if that approach works for you!

@brettmc would love to hear your thoughts too

agoallikmaa (approved):

PR #305, opened 2024-10-11
Add request hostname to span name 

NeilWhitworth:
addresses https://github.com/open-telemetry/opentelemetry-php/issues/1399

### DIFF
--- a/src/Instrumentation/Symfony/src/HttpClientInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/HttpClientInstrumentation.php
@@ -48,16 +48,34 @@ final class HttpClientInstrumentation
                 ?int $lineno,
             ) use ($instrumentation): array {
                 /** @psalm-suppress ArgumentTypeCoercion */
+                $requestOptions = $params[2] ?? [];
+
+                if (array_key_exists('base_uri', $requestOptions)) {
+                    $baseuri = $requestOptions['base_uri'];
+                    $parsedUrl = parse_url($baseuri);
+                    if (!array_key_exists('port', $parsedUrl)) {
+                        $parsedUrl['port'] = $parsedUrl['scheme'] === 'https' ? 443 : 80;
+                    }
+                } else {
+                    $parsedUrl = parse_url((string) $params[1]);
+                    if (!array_key_exists('port', $parsedUrl)) {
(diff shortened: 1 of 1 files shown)


brettmc:
This has come up before in various SIGs across the otel ecosystem, eg https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/6526 and https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/8339

The responses I've seen seem to boil down to "no, that's not what the spec says, here are some workarounds". The workarounds can include allowing developers to provide some custom logic to set the span name, or using a custom Span Processor or a collector component.
I'd also suggest another option for our SIG - we have the foundations of [file-based configuration](https://github.com/open-telemetry/opentelemetry-configuration), which allows for a much richer configuration of the SDK and auto-instrumentation. So perhaps we could do something with that? I'd be much more comfortable if any deviation from the spec was opt-in.

brettmc:
Here's a half-baked idea. I'm currently working on enabling [SDK configuration through SPI](https://github.com/open-telemetry/opentelemetry-php/pull/1412) and wondering if we can use SPI to solve this problem...

How about if we come up with a new API/SDK interface, say something like `HttpSpanNameExtractor`, with a method like `extract(RequestInterface $request): string`.
We can build a spec-compliant implementation (`{method}`) which would be the default, and one that does `{method} {host}` as in this PR. Using SPI also allows folks to write their own super-customized version as suits their needs, and non-default behaviour is opt-in.

NeilWhitworth:
> Here's a half-baked idea. I'm currently working on enabling [SDK configuration through SPI](https://github.com/open-telemetry/opentelemetry-php/pull/1412) and wondering if we can use SPI to solve this problem...
> 
> How about if we come up with a new API/SDK interface, say something like `HttpSpanNameExtractor`, with a method like `extract(RequestInterface $request): string`. We can build a spec-compliant implementation (`{method}`) which would be the default, and one that does `{method} {host}` as in this PR. Using SPI also allows folks to write their own super-customized version as suits their needs, and non-default behaviour is opt-in.

It would be useful to have an API interface to control this. Internally we have several Libraries (build from OpenAPI specs) that _could_ build a better name, including the low-cardinality route information, but I could not see any way to do this within the current implementation.

Falling back to using `{method} {host}` seamed a better option that worked most of the time, and provided more information whilst staying within the spirit of the specs.

PR #306, opened 2024-10-16
fix(contrib-auto-laravel): fix log level type issue

flc1125:
fix: https://github.com/open-telemetry/opentelemetry-php-contrib/pull/264#discussion_r1802328849

### DIFF
--- a/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
+++ b/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
@@ -29,10 +29,7 @@ class LogWatcher extends Watcher
         $this->logger = $app['log'];
     }
 
-    /**
-     * Record a log.
-     */
-    public function recordLog(MessageLogged $log): void
+    public function shouldRecordLog(MessageLogged $log): bool
     {
         $underlyingLogger = $this->logger->getLogger();
 
@@ -41,6 +38,18 @@ class LogWatcher extends Watcher
             && method_exists($underlyingLogger, 'toMonologLevel')
             && !$underlyingLogger->isHandling($underlyingLogger->toMonologLevel($log->level))
         ) {
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * Record a log.
+     */
+    public function recordLog(MessageLogged $log): void
+    {
+        if (!$this->shouldRecordLog($log)) {
             return;
         }
 
--- a/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
+++ b/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
@@ -37,7 +37,9 @@ class LogWatcher extends Watcher
         $underlyingLogger = $this->logger->getLogger();
 
         /** @phan-suppress-next-line PhanUndeclaredMethod */
-        if (method_exists($underlyingLogger, 'isHandling') && !$underlyingLogger->isHandling($log->level)) {
+        if (method_exists($underlyingLogger, 'isHandling') && !$underlyingLogger->isHandling(
+                $underlyingLogger->toMonologLevel($log->level))
+        ) {
             return;
         }
 
--- a/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
+++ b/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
@@ -37,8 +37,9 @@ class LogWatcher extends Watcher
         $underlyingLogger = $this->logger->getLogger();
 
         /** @phan-suppress-next-line PhanUndeclaredMethod */
-        if (method_exists($underlyingLogger, 'isHandling') && !$underlyingLogger->isHandling(
-                $underlyingLogger->toMonologLevel($log->level))
+        if (method_exists($underlyingLogger, 'isHandling')
+            && method_exists($underlyingLogger, 'toMonologLevel')
+            && !$underlyingLogger->isHandling($underlyingLogger->toMonologLevel($log->level))
         ) {
             return;
         }
--- a/src/Instrumentation/Laravel/composer.json
+++ b/src/Instrumentation/Laravel/composer.json
@@ -48,7 +48,8 @@
     "lock": false,
     "sort-packages": true,
     "allow-plugins": {
-      "php-http/discovery": false
+      "php-http/discovery": false,
+      "tbachert/spi": true
     }
   }
 }
--- a/src/Instrumentation/Laravel/composer.json
+++ b/src/Instrumentation/Laravel/composer.json
@@ -48,8 +48,7 @@
     "lock": false,
     "sort-packages": true,
     "allow-plugins": {
-      "php-http/discovery": false,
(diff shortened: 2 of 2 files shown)


src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
```diff
@@ -37,7 +37,9 @@ public function recordLog(MessageLogged $log): void
         $underlyingLogger = $this->logger->getLogger();
 
         /** @phan-suppress-next-line PhanUndeclaredMethod */
-        if (method_exists($underlyingLogger, 'isHandling') && !$underlyingLogger->isHandling($log->level)) {
+        if (method_exists($underlyingLogger, 'isHandling') && !$underlyingLogger->isHandling(
+                $underlyingLogger->toMonologLevel($log->level))
```
brettmc, line 41: this seems like it would now _only_ work with monolog as the underlying logger, but there could be others, right? is `toMonologLevel` a part of any common interface between loggers?

flc1125: @brettmc 
I'm not quite sure how many of the third-party libraries in the industry that comply with `\Psr\Log\LoggerInterface` support the method `method_exists($underlyingLogger, 'isHandling')`.

- Suppose there are many. Then I wonder if adding `method_exists($underlyingLogger, 'toMonologLevel')` is too specific.
- Suppose there are few, and it is more for supporting `Monolog`. Then I think it is acceptable to add `method_exists($underlyingLogger, 'toMonologLevel')`.

Of course, there is another way, which is to add an adapter to solve this problem, but the extension might be a bit cumbersome.

So, I'd like to know your thoughts.

flc1125: Also, there is another method, that is, I define an interface of ShouldHandle and bring it in when doing `new LogWatcher($this->instrumentation)`. It's probably like this:

```php
<?php

declare(strict_types=1);

use Illuminate\Log\Events\MessageLogged;
use Illuminate\Log\LogManager;

interface ShouldHandle
{
    public function shouldHandle(LogManager $logger, MessageLogged $log): bool;
}
```

flc1125: @brettmc 
Another processing method. Because` method_exists($underlyingLogger, 'isHandling')` it is too special. This method is not supported in `\Psr\Log\LoggerInterface`. 

As an official extension package, for this special case, I suggest that users can consider overriding the `recordLog` method of `LogWatcher`.

And for this package, we only need to add a `shoudRecordLog` method.

----

For this, what we may have to do:

- Roll back https://github.com/open-telemetry/opentelemetry-php-contrib/pull/264
- Add a `shoudRecordLog` method 

> Of course, there are some other additions. That is, it can be retrieved more conveniently through some configurations. Currently, our register entry has some limitations. But this may take some time to continuously improve.

bobstrecansky:
@flc1125 - will you add some relevant tests to this PR as well please?

flc1125:
> @flc1125 - will you add some relevant tests to this PR as well please?

No problem. Give me some time.

ChrisLightfootWild:
> fix: [#264 (comment)](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/264#discussion_r1802328849)

@flc1125 this link appears to be broken for me. Though if I hover over it, I see the following:
image

I can't actually see that review though.

---

Have you seen an example of this failure? (Do you have a stacktrace to hand?)

flc1125:
> > fix: [#264 (comment)](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/264#discussion_r1802328849)
> 
> @flc1125 this link appears to be broken for me. Though if I hover over it, I see the following: image
> 
> I can't actually see that review though.
> 
> Have you seen an example of this failure? (Do you have a stacktrace to hand?)

@ChrisLightfootWild Ch

image

> It should be collapsed.  

----

**This is an example of an error I encountered:**  

- PHP: 8.1.1
- Laravel Framework 9.52.16

```
Uncaught TypeError: Monolog\Logger::isHandling(): Argument #1 ($level) must be of type int, string given, called in /webroot/project/vendor/open-telemetry/opentelemetry-auto-laravel/src/Watchers/LogWatcher.php on line 40 and defined in /webroot/project/vendor/monolog/monolog/src/Monolog/Logger.php:530
Stack trace:
#0 /webroot/project/vendor/open-telemetry/opentelemetry-auto-laravel/src/Watchers/LogWatcher.php(40): Monolog\Logger->isHandling()
#1 /webroot/project/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php(421): OpenTelemetry\Contrib\Instrumentation\Laravel\Watchers\LogWatcher->recordLog()
#2 /webroot/project/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php(249): Illuminate\Events\Dispatcher->Illuminate\Events\{closure}()
#3 /webroot/project/vendor/laravel/framework/src/Illuminate/Log/Logger.php(245): Illuminate\Events\Dispatcher->dispatch()
#4 /webroot/project/vendor/laravel/framework/src/Illuminate/Log/Logger.php(186): Illuminate\Log\Logger->fireLogEvent()
#5 /webroot/project/vendor/laravel/framework/src/Illuminate/Log/Logger.php(94): Illuminate\Log\Logger->writeLog()
#6 /webroot/project/vendor/laravel/framework/src/Illuminate/Log/LogManager.php(645
```

> Besides, I checked the latest version of Laravel. Its type hasn't changed and the same problem is estimated to occur.

flc1125:
@bobstrecansky @brettmc I think I already know where the problem is. 

The occurrence of this problem mainly lies in the version of `Monolog`.

- Under version `2.x`, the data type problem I mentioned will occur.
- Under version `3.x`, the policy is used. And Laravel has upgraded to this version since `v10.x` (related [PR](https://github.com/laravel/framework/pull/42311)). 

> https://github.com/Seldaek/monolog/pull/1656/files#diff-1e7fd545cec457de96f5ed6bd7249ba091cd9e699b4057db15ff1e2e0364025bR429-R434

-----

I want to know if our maintenance policy has any requirements for the version of Laravel? This will facilitate me in dealing with this problem.  

> Of course, for my problem, upgrading my framework is a good choice.

brettmc:
> I want to know if our maintenance policy has any requirements for the version of Laravel?

Our policy is to try to support the versions that are actively supported

flc1125:
> > I want to know if our maintenance policy has any requirements for the version of Laravel?
> 
> Our policy is to try to support the versions that are actively supported

@brettmc OK, just checked official maintenance info this pr should be closed. thanks everyone for the help. 
https://laravel.com/docs/11.x/releases#support-policy

image

ChrisLightfootWild:
@flc1125 thanks for the additional info. I'm still going to try and find some time to look into this a bit further and see if we need to tackle this now.

I have the basis of v1.1 for Laravel instrumentation in #269 and that targets the supported versions of Laravel which may encounter this.

PR #308, opened 2024-10-18
[symfony] Fix: Set span status according to otel convention

technimad-splunk:
Only set error status on span  if we know the response code is 500 or higher.

Addresses https://github.com/open-telemetry/opentelemetry-php/issues/1403

In Symfony any error is handled as an exception. The instrumentation added an error status to any span when an exception occurred. This is not in line with the spec. For span kinds of server, the error status must ONLY be set for response codes of 500 and higher.
Earlier improvement was done in PR #295, but that fix wasn't sufficient.


brettmc (approved):

brettmc:
Thanks, @technimad-splunk . It looks like there is a linting error in Symfony (see checks), can you please fix that?

agoallikmaa (approved):

PR #309, opened 2024-10-22
Fix broken Serializer when using Laravel Illuminate\Queue\LuaScripts

gdaszuta:
Since introduction of RedisCommandWatcher, auto-instrumentation for Laravel fails when using Lua scripts for Queue management, since command argument array can contains arrays.

I don't have the time right now to prepare failing PoC, problematic payload is created in Illuminate\Queue\RedisQueue, it seems that it fails when multiple jobs are passed into argument as an object.


brettmc:
Looks sensible to me - can you add (or modify) a test for this?

gdaszuta:
@brettmc added a simple test for the case, is this enough?

brettmc (approved):

ChrisLightfootWild:
@MilesChou would you be happy with these changes?

ChrisLightfootWild (approved):

PR #311, opened 2024-10-26
Laravel: handle possible LogWatcher TypeError's

ChrisLightfootWild:
Based on reported issues in #306 (with thanks to @flc1125 for raising it).

Given we are close to moving to package 1.1.x and dropping support of Laravel 6-9, then it may be preferable to restore the previous behaviour and allow the log messages through. This would keep the package viable for the older Laravel versions.

Unfortunately, there's no log level filtering at that point but I think it's better to handle this gracefully.

The approach in #306 could also provide a stable v1, while we look to tackle this more elegantly in future.

### DIFF
--- a/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
+++ b/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
@@ -10,6 +10,7 @@ use Illuminate\Log\LogManager;
 use OpenTelemetry\API\Instrumentation\CachedInstrumentation;
 use OpenTelemetry\API\Logs\LogRecord;
 use OpenTelemetry\API\Logs\Map\Psr3;
+use TypeError;
 
 class LogWatcher extends Watcher
 {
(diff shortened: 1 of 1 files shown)


flc1125:
If we plan to only support Laravel versions 10 + and above. 

I think the Laravel dependency in `composer.json` can also be evaluated if it needs to be adjusted.

https://github.com/open-telemetry/opentelemetry-php-contrib/blob/2d2c37b7783f59967de79eba7cad8225e8662dd8/src/Instrumentation/Laravel/composer.json#L15

brettmc (approved):

ChrisLightfootWild:
> If we plan to only support Laravel versions 10 + and above. 
> 
> I think the Laravel dependency in `composer.json` can also be evaluated if it needs to be adjusted.
> 
> https://github.com/open-telemetry/opentelemetry-php-contrib/blob/2d2c37b7783f59967de79eba7cad8225e8662dd8/src/Instrumentation/Laravel/composer.json#L15
> 
> > Another message: Aside from the fix this time, I'm currently running fine on version 9.x.

Yes to v1.1 moving to support 10 + 11 👍

This package *should* still work down to Laravel 6; it is just hard to make progress when having to test so many end-of-life versions!

flc1125:
> > If we plan to only support Laravel versions 10 + and above.
> > I think the Laravel dependency in `composer.json` can also be evaluated if it needs to be adjusted.
> > https://github.com/open-telemetry/opentelemetry-php-contrib/blob/2d2c37b7783f59967de79eba7cad8225e8662dd8/src/Instrumentation/Laravel/composer.json#L15
> > 
> > > Another message: Aside from the fix this time, I'm currently running fine on version 9.x.
> 
> Yes to v1.1 moving to support 10 + 11 👍
> 
> This package _should_ still work down to Laravel 6; it is just hard to make progress when having to test so many end-of-life versions!

Very good plan. If multi-version compatibility needs to be considered, it is a challenge. 

Currently, accelerating the advancement of Laravel in OTEL may be a more priority task to do. 💪🏻

PR #316, opened 2024-11-07
cURL instrumentation: support partially successful curl_setopt_array

vikmovcan:
When `curl_setopt_array` is partially successful we end up not recording curl options which were set successfully.
In order to work around this, the fix tries to emulate the behaviour of how `curl_setopt_array` works, and iterates over the passed options using `curl_opt`, recording each valid one, until an invalid option is encountered (reference: https://github.com/php/php-src/blob/fb257ee83c405fecf449571bfcd1cc0fb4910336/ext/curl/interface.c#L2381).

### DIFF
--- a/src/Instrumentation/Curl/src/CurlInstrumentation.php
+++ b/src/Instrumentation/Curl/src/CurlInstrumentation.php
@@ -82,6 +82,14 @@ class CurlInstrumentation
             pre: null,
             post: static function ($obj, array $params, mixed $retVal) use ($curlHandleToAttributes) {
                 if ($retVal != true) {
+                    if (curl_error($params[0])) {
+                        foreach ($params[1] as $option => $value) {
+                            if (!curl_setopt($params[0], $option, $value)) {
+                                break;
+                            }
+                        }
+                    }
+
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


src/Instrumentation/Curl/src/CurlInstrumentation.php
```diff
@@ -82,6 +82,14 @@ public static function register(): void
             pre: null,
             post: static function ($obj, array $params, mixed $retVal) use ($curlHandleToAttributes) {
                 if ($retVal != true) {
+                    if (curl_error($params[0])) {
+                        foreach ($params[1] as $option => $value) {
+                            if (!curl_setopt($params[0], $option, $value)) {
+                                break;
```
ChrisLightfootWild, line 88: If we `continue` here instead, would it permit setting the subsequent options too? (in case there is more than one error)

vikmovcan: yes it would, however I dont see a reason to do that since I think we should only be interested in recording options which would be used later (in the following curl handle execution).
trying to record all valid options might be useful (and could be done without `continue` too I think), but that is outside of scope of this PR.

brettmc: I think I agree with @vikmovcan here. We shouldn't set the options if curl itself wouldn't - so if `curl_setopt_array` wouldn't set them, neither should we.

intuibase (approved):
Looks good, thanks for catching that.

brettmc (approved):

brettmc:
released v0.0.2

PR #317, opened 2024-11-28
Fix boundaries for Symfony HttpKernel auto instrumentation

dciprian-petrisor:
Fixes https://github.com/open-telemetry/opentelemetry-php/issues/1443


src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
```diff
@@ -80,6 +80,30 @@ public static function register(): void
                 array $params,
                 ?Response $response,
                 ?\Throwable $exception
+            ): void {
+                $scope = Context::storage()->scope();
+                if (null === $scope || null === $exception) {
```
brettmc, line 85: So if an exception occurs, this post callback will run, otherwise the `terminate` one will? If so, we should document this.

dciprian-petrisor: That is correct.
I'm assuming this should be documented in the README.md, right?

ChrisLightfootWild: The examples on [this page](https://symfony.com/doc/current/components/http_kernel.html#http-kernel-working-example) seem to suggest that the `handle` and `terminate` can both run.

So it looks like you'd expect 2 spans to be created and then with these changes only close one of them? I'm not so sure that behaviour is correct.

ChrisLightfootWild: There appears to be a parameter to configure whether `handle` should throw [here](https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/HttpKernel/HttpKernel.php#L69).

So when `$catch = false` the handle post-hook might have an exception and `terminate` will not have been reached yet. I think this would be fine as it handles the recording of the exception and detaches the scope.

However, if `$catch = false` and there is no exception, then `terminate()` would be called after `handle` and the `terminate` post-hook could try to close a scope which does not belong to it.

dciprian-petrisor: @ChrisLightfootWild 

1. If `handle` throws an exception which is never caught, then `terminate` will never be reached either, according to the[ `HttpKernelRunner` implementation.](https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/Runtime/Runner/Symfony/HttpKernelRunner.php)
In this situation, it is correct to close the scope on the `post` hook of `handle`.

2. If `handle` does not throw an exception (or has `$catch = true`), then `terminate`  would run afterwards, closing the scope.

The only problem I see is there is not a 100% guarantee that `terminate` will always be executed after `handle`.

For example, something else might go wrong in `HttpKernelRunner` after `handle` was called, e.g. `fastcgi_finish_request()/litespeed_finish_request()` might crash, `flush()` might crash, etc, leaving the scope still open.

Unsure how much of a real issue this is, in most cases.


Normally this would be solved by hooking into `HttpKernelRunner::run` with both `pre` and `post` hooks, as it covers both `handle` and `terminate`, however `$request` and `$response` are not available as parameters there, so I am at a loss right now.


We need either a compromise or another solution enitrely

ChrisLightfootWild: Thanks for that @dciprian-petrisor - I had another look with less bleary eyes and I finally caught up to you guys 💪

LionsAd: I am having the same problem for Drupal, but opted to do an auto root span for now, which obviously duplicates the logic to do that.

That way I have a duplicated span for the Kernel::handle() + Kernel::terminate(), but that does not matter as it's still within one root trace.

The Auto Root Span uses the SDK ShutdownHandler::register so that might be also an option here.

The other thing is:

You could hook into KernelRunner::run(), but not open the scope here, but instead close it just during the post() [if ::run() was called in pre].

While that is more logic to handle, because handle() needs to check that ::run was actually called, so that it can close it, cross hook frontiers are a good way to handle this.

```
::run pre -> $runCalled = 1
::handle pre -> Open root scope like normal
::handle post -> if ($runCalled) {} -> Do not bother to close scope, else fallback to current logic
::terminate pre -> Open span like normal
::terminate post -> Close span like normal
::run post -> Close scope
```

In general I personally will still call the Auto Root Scope hook as early as possible (e.g. DrupalKernel::__construct in my case), because else there is always an ::IO extension call, which escapes.

Until that's converted to SPI obviously there are still race conditions.

LionsAd: This is my implementation:

```php
<?php

namespace OpenTelemetry\Contrib\Instrumentation\Drupal;

use Drupal\Core\DrupalKernel;
use OpenTelemetry\SDK\Trace\AutoRootSpan;
use function OpenTelemetry\Instrumentation\hook;

class DrupalAutoRootSpan {

  public static function register(): void {

    hook(
      DrupalKernel::class,
      '__construct',
      static::registerAutoRootSpan(),
      NULL,
    );
  }

  /**
   * @param \OpenTelemetry\API\Instrumentation\CachedInstrumentation $instrumentation
   *
   * @return \Closure
   */
  public static function registerAutoRootSpan(): \Closure {
    return static function (DrupalKernel $kernel, array $params, string $class, string $function, ?string $filename, ?int $lineno) {
      $request = AutoRootSpan::createRequest();

      if ($request) {
        AutoRootSpan::create($request);
        AutoRootSpan::registerShutdownHandler();
      }
    };
  }

}
```

It works really well.

dciprian-petrisor:
> > Just running into the exact same issue in a project I am working on. Is this on the horizon to be merged any time soon?
> 
> No. I'm not sure of the current state. It at least needs to be rebased given its age, and there's some discussion that didn't conclude (and I don't know enough Symfony to make a call myself)
> @dciprian-petrisor ?

Hi, I'll look into this over the weekend and if you all agree we could merge it.

dciprian-petrisor:
I've rebased this PR.

After a brief look with fresh eyes, given it's been some time since this PR had activity, here are my conclusions:

1. The implementation in this PR covers the 'normal' execution of any Symfony HTTP Request, minus the situation where any of the following calls in [HttpKernelRunner](https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/Runtime/Runner/Symfony/HttpKernelRunner.php) would fail miserably: 
```
fastcgi_finish_request();
```
or
```
litespeed_finish_request();
```
or
```
Response::closeOutputBuffers(0, true);
flush();
```

All three of these failing indicate a **severe** problem and I have personally never encountered failures of them.
These calls failing would indicate issues with FASTCGI/PHP internals or the SAPI.


2. In **any** other situation, one of the following two outcomes can happen:
- the `post` hook of `handle` runs whenever **an exception occurs/is thrown** while processing the Symfony request.
  otherwise, a `Response` object is returned from `handle`.
- the `terminate` call is always made when `handle` **does not throw**.

Both of these will finalize the trace.


In summary, I believe this implementation is good as-is.
Regarding the documentation, if you can give me a few hints as to where to document it, I would gladly do it.

dciprian-petrisor:
@brettmc  just did

PR #318, opened 2024-12-04
Auto instrumentation for mysqli

intuibase:
Example view in Kibana


src/Instrumentation/MySqli/src/MySqliTracker.php
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    public function untrackMySqliTransaction(mysqli $mysqli)
+    {
+        if ($this->mySqliTransaction->offsetExists($mysqli)) {
+            unset($this->mySqliTransaction[$mysqli]);
+        }
+    }
+
+    private function splitQueries(string $sql)
```
ChrisLightfootWild, line 164: Is it possible to add a test that covers the query splitting?

intuibase: Sure thing

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
```diff
@@ -0,0 +1,815 @@
...
+            return; //TODO create error span?
+        }
+        $tracker->addMySqliAttribute($obj ? $obj : $params[0], TraceAttributes::DB_NAMESPACE, $params[$obj ? 0 : 1]);
+    }
+
+    private static function preparePostHook(CachedInstrumentation $instrumentation, MySqliTracker $tracker, $obj, array $params, mixed $stmtRetVal, ?\Throwable $exception)
+    {
+
+        if ($exception || !$stmtRetVal instanceof mysqli_stmt) {
+            self::logDebug('mysqli::prepare failed', ['exception' => $exception, 'obj' => $obj, 'retVal' => $stmtRetVal, 'params' => $params]);
```
brettmc, line 563: there are a few of these logDebug calls - should they be logged like this, or recorded as an exception on the active span, which [the spec requires](https://opentelemetry.io/docs/specs/otel/trace/exceptions/#recording-an-exception)

intuibase: Thank you. I’ve implemented instrumentation for prepare, as it makes sense because it involves IO with the database.

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
```diff
@@ -0,0 +1,832 @@
...
+
+    private static function beginTransactionPostHook(CachedInstrumentation $instrumentation, MySqliTracker $tracker, $obj, array $params, mixed $retVal, ?\Throwable $exception)
+    {
+        $mysqli = $obj ? $obj : $params[0];
+        $transactionName = $params[$obj ? 1 : 2] ?? null;
+
+        $attributes = $tracker->getMySqliAttributes($mysqli);
+
+        if ($transactionName) {
+            $attributes['db.transaction.name'] = $transactionName;
```
intuibase, line 609: In reference to the discussion on SIG and the custom attribute db.transaction.name, I looked into it and found that there is no corresponding entity in sem-conv. I started digging deeper into the topic and discovered that this is typical for the MySQLi extension in PHP. PDO does not allow setting a transaction name.

I checked the code of mysqlnd (the base driver for MySQLi and PDO), and it seems that there is no such concept as a transaction name. In mysqlnd, this is implemented by setting a comment in the SQL code—it is not interpreted by the database itself as a name but is treated as a regular comment.

brettmc: Discussed this in SIG, and we decided that transaction name is useful telemetry metadata and should be added. Since this instrumentation is unstable, the spec says we can add additional telemetry: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/telemetry-stability.md#unstable-instrumentations

It would be worth pursuing having `db.transaction.name` or something like it added to semconv.

brettmc (approved):

brettmc:
@intuibase are you happy for me to merge this and get a 0.0.1 version out?

intuibase:
> @intuibase are you happy for me to merge this and get a 0.0.1 version out?
Yes, please :)

bobstrecansky (approved):

src/Instrumentation/MySqli/src/MySqliTracker.php
```diff
@@ -0,0 +1,214 @@
...
+    {
+        // Normalize line endings to \n
+        $sql = preg_replace("/\r\n|\n\r|\r/", "\n", $sql);
+
+        $queries = [];
+        $buffer = '';
+        $blockDepth = 0;
+        $tokens = preg_split('/(;)/', $sql, -1, PREG_SPLIT_DELIM_CAPTURE); // Keep semicolons as separate tokens
```
ChrisLightfootWild, line 172: How should this parse SQL which contains semicolons in queries such as: 

```sql
INSERT INTO `tableA` (columnA) VALUES ('hello; world'); SELECT LAST_INSERT_ID();
```

I guess this might only be applicable if someone is writing raw SQL, as I'd probably expect parameterized SQL otherwise:

```sql
INSERT INTO `tableA` (columnA) VALUES (?); SELECT LAST_INSERT_ID();
```

I didn't test this out myself, but thought it was worth mentioning in case anybody has any input on it.

src/Instrumentation/MySqli/src/MySqliTracker.php
```diff
@@ -0,0 +1,214 @@
...
+            if (preg_match('/\bEND\b(\s+[a-zA-Z0-9_]+)?\s*$/i', $token)) {
+                $blockDepth--;
+            }
+
+            // If we are outside a block and encounter a semicolon, split the query
+            if ($blockDepth === 0 && $token === ';') {
+                $trimmedQuery = trim($buffer);
+                if ($trimmedQuery !== ';') { // Ignore empty queries
+                    $queries[] = $trimmedQuery;
+                    //substr($trimmedQuery, 0, -1); // Remove the trailing semicolon
```
ChrisLightfootWild, line 200:
```suggestion
```

src/Instrumentation/MySqli/src/MySqliTracker.php
intuibase: It will fail and split into two queries "....hello;" and "world...".
I tried to use mysql parser libraries and even those one from phpmysqladmin failed in some of the tests.
Of course I can try to fix that case but I'm pretty sure that i'll fail with other example sooner or later.

intuibase:
@ChrisLightfootWild  thanks, I've improved parser a bit, now it should work with semicolons inside quotes.

ChrisLightfootWild (approved):

brettmc:
https://packagist.org/packages/open-telemetry/opentelemetry-auto-mysqli

PR #322, opened 2025-01-07
fix symfony breakages

brettmc:
* update TraceableTracerProvider to support SDK 1.1.0+ (add updateConfigurator method).
* update sdk dependency to 1.1 (which drops PHP 7.4+8.0 support)
* update SystemClock refs from sdk to api
* update psalm config
* use stable versions

This makes the build green again.

Fixes: https://github.com/open-telemetry/opentelemetry-php/issues/1461

### DIFF
--- a/src/Symfony/composer.json
+++ b/src/Symfony/composer.json
@@ -7,7 +7,7 @@
     ],
     "prefer-stable": true,
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": "^8.1",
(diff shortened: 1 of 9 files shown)


src/Symfony/composer.json
agoallikmaa, line 1: Can the `symfony/polyfill-php80` dependency removed from this file as well?

agoallikmaa (approved):

PR #324, opened 2025-01-08
ext-mongodb 2.0 support

brettmc:
Being able to retrieve host, port + info from a CommandStartedEvent is deprecated in 1.20.0 and will be removed in 2.0. we have been advised to use SDAM subscription instead.

- update tests to allow defining mongo server from env
- require ext-mongodb 1.13 (which provides SDAM subscriber for server info)
- use SDAM subscriber to get server info attributes. Subscribe to the serverChanged event and store the attributes in a class variable, organised by host/port.

Fixes: https://github.com/open-telemetry/opentelemetry-php/issues/1467


src/Instrumentation/MongoDB/src/MongoDBInstrumentationSubscriber.php
```diff
@@ -118,4 +135,58 @@ private static function endSpan(?Throwable $exception = null): void
 
         $span->end();
     }
+
+    public function serverChanged(ServerChangedEvent $event): void
+    {
+        $host = $event->getHost();
+        $port = $event->getPort();
+        $info = $event->getNewDescription()->getHelloResponse();
```
agoallikmaa, line 143: There is a corner case mentioned in MongoDB extension documentation at https://www.php.net/manual/en/mongodb-driver-serverdescription.gethelloresponse.php
> When the driver is connected to a load balancer, this method will return an empty array since load balancers are not monitored. This is in contrast to [MongoDB\Driver\Server::getInfo()](https://www.php.net/manual/en/mongodb-driver-server.getinfo.php), which would return the backing server's [» hello](https://www.mongodb.com/docs/manual/reference/command/hello/) command response from the initial connection handshake.

The code comment regarding this still seems present in `2.x` branch of mongo extension, so this is probably still the case even though the above documentation is for `1.x`. I don't see an obvious workaround for `2.x`, but possibly if this `$info` here is empty array and we detect that mongo extension version is 1.x, there could be a fallback to the previous behavior.

brettmc: Do you know what they mean by "load balancer"? I set up a replicaSet and it correctly fetches info from all servers via the `serverChanged` event.

agoallikmaa: I checked what the extension tests with for tests that run "with load balancer", and it seems they use `haproxy` as seen [here](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/run-load-balancer.sh). On the driver side it seems adding `loadBalanced=true` to connection URL parameters should make the driver behave as if it is connecting through a load balancer, so maybe no actual load balancer is required to trigger this behavior, just that parameter? The specification is [here](https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.md).

brettmc: ugh, `loadBalanced=true` seems to require a real, working, load-balanced setup, and I cannot get a load-balanced mongo cluster running.

I've got a 3-node replicaset, and haproxy in front. I can connect through haproxy and round-robin to the different nodes, however it's not "real" load balancing as adding `loadBalanced=true` gives me a `Driver attempted to initialize in load balancing mode, but the server does not support this mode.`

Since our mongo package is not stable (latest=`0.0.6`), the easiest path forward is to go ahead with this change, and wait for somebody with a load-balanced setup to show up (and/or create a todo/help-wanted issue).

Another point to consider is that all of these extra fields are not part of the official semconv, and according to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/telemetry-stability.md#stable-instrumentations they should not be used in a stable instrumentation. It's not a problem now, but we may end up needing to remove them in the future anyway, if we're following the spec dogmatically:

> Stable instrumentations authored by OpenTelemetry SHOULD NOT produce telemetry that is not described by OpenTelemetry semantic conventions

agoallikmaa (approved):

src/Instrumentation/MongoDB/src/MongoDBInstrumentationSubscriber.php
agoallikmaa: That seems reasonable. :+1:

jmikola: Regarding load balancers, some drivers do use HAPorxy for local testing but the actual feature is more involved and entails the backing servers also being aware of the load balancer (and returning a `serviceId` field in their `hello` responses). The [Load Balancer](https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.md) specification goes into more detail about the feature.

PR #326, opened 2025-02-04
fix(Aws): fixes wrong aws client name and region binding when appending middlewares

rtreffler:
Binding happened outside the foreach loop resulting in binding the same values for all instrumented clients. Similar problem was affecting span and scope object variables, where multiple clients could overwrite each other spans.

### DIFF
--- a/src/Aws/tests/Integration/UsesServiceTrait.php
+++ b/src/Aws/tests/Integration/UsesServiceTrait.php
@@ -4,14 +4,10 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Tests\Aws\Integration;
 
-use Aws\Api\Service;
 use Aws\AwsClientInterface;
-use Aws\CommandInterface;
-use Aws\Exception\AwsException;
 use Aws\MockHandler;
 use Aws\Result;
 use Aws\Sdk;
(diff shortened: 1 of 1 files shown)


src/Aws/tests/Integration/UsesServiceTrait.php
```diff
@@ -0,0 +1,149 @@
+<?php
```
rtreffler, line 1: this file is copied almost exactly how it was written in AWS PHP SDK: https://github.com/aws/aws-sdk-php/blob/master/tests/UsesServiceTrait.php

not sure that is important, thought its worth mentioning

bobstrecansky:
Thank you @rtreffler!  Once you've signed the CLA we'll be able to review and merge.

rtreffler:
> Thank you @rtreffler! Once you've signed the CLA we'll be able to review and merge.

The company I'm currently working for is already working on the CLA topic.

src/Aws/tests/Integration/UsesServiceTrait.php
rtreffler: its now a modified subset of that file

rtreffler:
is there a chance we could progress with it this week ?

brettmc (approved):

PR #338, opened 2025-03-29
fix: Track attributes for new PDO subclasses & span hierarchy

cedricziel:
This adds connection tracking to the new pdo subclasses in PHP 8.4

It also fixes an issue where the hierarchy of a trace was not correctly maintained.

Also combines attribute tracking into one cohesive method.

Fixes: https://github.com/open-telemetry/opentelemetry-php/issues/1469

Context: https://www.php.net/releases/8.4/en.php#pdo_driver_specific_subclasses


src/Instrumentation/PDO/src/PDOInstrumentation.php
```diff
@@ -213,9 +263,10 @@ public static function register(): void
                 if ($spanContext = $pdoTracker->getSpanForPreparedStatement($statement)) {
                     $builder->addLink($spanContext);
                 }
+                $parent = Context::getCurrent();
                 $span = $builder->startSpan();
 
-                Context::storage()->attach($span->storeInContext(Context::getCurrent()));
+                Context::storage()->attach($span->storeInContext($parent));
```
brettmc, line 269: I'm pretty confident this is equivalent - it the change for readability?

brettmc (approved):

src/Instrumentation/PDO/src/PDOInstrumentation.php
ChrisLightfootWild: I think this was the fix for:

> It also fixes an issue where the hierarchy of a trace was not correctly maintained.

I'd guess `$builder->startSpan()` was resolving context in an incorrect way which changed the hierarchy?

cedricziel: That's also my assumption.

bobstrecansky (approved):

src/Instrumentation/PDO/src/PDOInstrumentation.php
```diff
@@ -31,6 +31,56 @@ public static function register(): void
...
+                    ?int $lineno,
+                ) use ($instrumentation) {
+                    /** @psalm-suppress ArgumentTypeCoercion */
+                    $builder = self::makeBuilder($instrumentation, 'PDO::connect', $function, $class, $filename, $lineno)
+                        ->setSpanKind(SpanKind::KIND_CLIENT);
+                    if ($class === PDO::class) {
+                        $builder
+                            ->setAttribute(TraceAttributes::SERVER_ADDRESS, $params[0] ?? 'unknown')
+                            ->setAttribute(TraceAttributes::SERVER_PORT, $params[0] ?? null);
+                    }
```
Nevay, line 54: The `if` condition is not required (only needed for `__construct` as constructors can be overridden with arbitrary parameters).

Should add the `//@todo` from `__construct()` / IMO we should not populate the `server.` attributes until the todo is fixed.

intuibase (approved):

brettmc:
@cedricziel - is this ready to merge?

cedricziel:
@brettmc ready!

PR #348, opened 2025-04-10
fix: matching will be strict

cedricziel:
This changes the behavior with the strict parameter:

* by default, all attributes must match
* with the strict parameter, the assertion will do an exhaustive check on _all_ attributes in the actual, not just the expected attributes, making it more suitable for integration tests


brettmc (approved):

cedricziel:
Good to go :)

PR #349, opened 2025-04-10
Allow return integer type when resolving attributes

DominicDetta:


src/Instrumentation/Doctrine/src/AttributesResolver.php
```diff
@@ -41,7 +41,7 @@ final class AttributesResolver
         'trino',
     ];
 
-    public static function get(string $attributeName, array $params): string
+    public static function get(string $attributeName, array $params)
```
brettmc, line 44: could this be `string|int` ?

DominicDetta: I could transform to `string|int|null`.

DominicDetta: done

src/Instrumentation/Doctrine/src/AttributesResolver.php
```diff
@@ -63,9 +63,9 @@ private static function getServerAddress(array $params): string
     /**
      * Resolve attribute `server.port`
      */
-    private static function getServerPort(array $params): string
+    private static function getServerPort(array $params): int
```
brettmc, line 66: This might need some more thought. `int(null)` is 0, but that's probably not what we want if port is not set. I think it would be better to return `?int`...null attributes will be discarded but not so `0` integer values

DominicDetta: Ok. I will change the logic.

brettmc (approved):

PR #357, opened 2025-04-18
[aws-contrib] fix bug where span is not ended/scope is not detached, when client operation promise rejects

mitchierichie:
If a client operation fails for whatever reason (in my case, I called `$kmsClient->decrypt()` with an invalid payload), the span will not be recorded.

It should be recorded with an error, but the `mapResult` middleware is only meant for mapping fulfilled promises, and does not allow for any kind of interaction with rejected promises.

Here, I am using a custom middleware instead of `mapResult`, so I can handle both fulfillment and rejection.

To reproduce this issue, here's an example:

```
$kmsClient = new KmsClient(['version' => 'latest', 'region' => 'us-east-1']);

// setup your instrumentation 
$instrumentation->instrumentClients([$kmsClient]);

$bar = $kmsClient->decrypt(['CiphertextBlob' => 'foo']);
```

This will fail and throw an exception that we were unable to decrypt, which is expected. but, no span will be recorded. and, if you do not have debug scopes disabled _and_ you have [zend assertions on](https://www.php.net/manual/en/function.assert.php), you'll also see an error like this:

```
Scope: missing call to Scope::detach() for scope #2534, created 
        at OpenTelemetry.Context.Context.activate(Context.php:79)
        at OpenTelemetry.API.Trace.Span.activate(Span.php:51)
        at OpenTelemetry.Aws.AwsSdkInstrumentation.OpenTelemetry.Aws.{closure}(AwsSdkInstrumentation.php:97)
        at Aws.Middleware.Aws.{closure}(Middleware.php:189)
        at Aws.AwsClient.executeAsync(AwsClientTrait.php:64)
        at Aws.AwsClient.execute(AwsClientTrait.php:58)
        at Aws.AwsClient.__call(AwsClientTrait.php:86)
``` 

Closes https://github.com/open-telemetry/opentelemetry-php/issues/1568

### DIFF
--- a/src/Aws/src/AwsSdkInstrumentation.php
+++ b/src/Aws/src/AwsSdkInstrumentation.php
@@ -4,14 +4,21 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Aws;
 
+use Aws\CommandInterface;
 use Aws\Middleware;
 use Aws\ResultInterface;
+use Closure;
+use Error;
+use GuzzleHttp\Promise;
 use OpenTelemetry\API\Instrumentation\InstrumentationInterface;
 use OpenTelemetry\API\Instrumentation\InstrumentationTrait;
 use OpenTelemetry\API\Trace\SpanKind;
+use OpenTelemetry\API\Trace\StatusCode;
 use OpenTelemetry\API\Trace\TracerInterface;
 use OpenTelemetry\API\Trace\TracerProviderInterface;
 use OpenTelemetry\Context\Propagation\TextMapPropagatorInterface;
+use Psr\Http\Message\RequestInterface;
+use Throwable;
 
 /**
  * @experimental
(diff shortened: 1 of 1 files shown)


mitchierichie:
fixes https://github.com/open-telemetry/opentelemetry-php/issues/1568

sahil-io:
been waiting for this fix. Thank you. FInally....

brettmc (approved):

mitchierichie:
@brettmc thank you for the speedy merge!!!

brettmc:
@mitchierichie - I was a bit hasty in merging this, there are a couple of minor CI issues. I can fix them, but it might be faster if you could take a look (since I can't approve my own code for merge) ?

mitchierichie:
@brettmc no problem, where should I look?

brettmc:
> no problem, where should I look?

I think you've found it now, but in the "checks" tab of your PR. Don't worry about any failures that aren't related to Aws...

PR #360, opened 2025-04-23
Fix mongo-ext requirements condition to support both ^1.13 and ^2.0 v…

mlazarov:
Fix mongo-ext requirements condition to support both ^1.13 and ^2.0 versions

### DIFF
--- a/src/Instrumentation/MongoDB/composer.json
+++ b/src/Instrumentation/MongoDB/composer.json
@@ -8,7 +8,7 @@
   "license": "Apache-2.0",
   "require": {
     "php": ">=8.1",
-    "ext-mongodb": "^1.13 | ^2.0",
+    "ext-mongodb": "^1.13 || ^2.0",
     "ext-json": "*",
     "mongodb/mongodb": "^1.15 || ^2.0",
     "open-telemetry/api": "^1.0",


ChrisLightfootWild (approved):
I'm not sure that this is broken, as I thought a single pipe still worked, but for consistency (and personal preference) the standard `|| (or)` gets a 👍 from me.

ChrisLightfootWild:
> > I'm not sure that this is broken, as I thought a single pipe still worked, but for consistency (and personal preference) the standard `|| (or)` gets a 👍 from me.
> 
> Currently when I try to run `composer require open-telemetry/opentelemetry-auto-mongodb` I get this error:
> 
> ```
> Cannot use open-telemetry/opentelemetry-auto-mongodb's latest version 0.0.8 as it requires ext-mongodb ^1.13 which is not satisfied by your platform.
> 
> open-telemetry/opentelemetry-auto-mongodb[0.0.7, ..., 0.0.8] require ext-mongodb ^1.13 -> it has the wrong version installed (2.0.0).
> ```
> 
> May be I'm wrong but with single pipe this isn't working as expected?

Thanks for highlighting this!

Perhaps there's some nuance here, as it notes the use of spacing for `logical AND` but then does also mention backwards compatibility with `logical OR` and `|`:
https://getcomposer.org/doc/articles/versions.md#version-range

Everyday is a school day; thanks for the fix 💪

cedricziel (approved):

PR #366, opened 2025-05-10
fix Pre Hook Exception in case _controller is not stringable

trigrab:
closes #1569

### DIFF
--- a/src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
@@ -46,8 +46,24 @@ final class SymfonyInstrumentation
                 $request = ($params[0] instanceof Request) ? $params[0] : null;
                 $type = $params[1] ?? HttpKernelInterface::MAIN_REQUEST;
                 $method = $request?->getMethod() ?? 'unknown';
+                $controller = $request?->attributes?->get('_controller');
+                                
+                if (is_string($controller)) {
+                    $controllerName = $controller;
+                } elseif (is_array($controller) && is_object($controller[0]) && isset($controller[1])) {
+                    // [$object, 'method']
+                    $controllerName = get_class($controller[0]) . '::' . $controller[1];
+                } elseif (is_array($controller) && is_string($controller[0]) && isset($controller[1])) {
+                    // ['ClassName', 'method']
+                    $controllerName = $controller[0] . '::' . $controller[1];
+                } elseif (is_object($controller)) {
+                    $controllerName = get_class($controller);
+                } else {
+                    $controllerName = 'sub-request';
+                }
+                
                 $name = ($type === HttpKernelInterface::SUB_REQUEST)
-                    ? sprintf('%s %s', $method, $request?->attributes?->get('_controller') ?? 'sub-request')
+                    ? sprintf('%s %s', $method, $controllerName)
                     : $method;
                 /** @psalm-suppress ArgumentTypeCoercion */
                 $builder = $instrumentation


src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
```diff
@@ -46,8 +46,24 @@ public static function register(): void
...
+                    // [$object, 'method']
+                    $controllerName = get_class($controller[0]) . '::' . $controller[1];
+                } elseif (is_array($controller) && is_string($controller[0]) && isset($controller[1])) {
+                    // ['ClassName', 'method']
+                    $controllerName = $controller[0] . '::' . $controller[1];
+                } elseif (is_object($controller)) {
+                    $controllerName = get_class($controller);
+                } else {
+                    $controllerName = 'sub-request';
+                }
```
Nevay, line 63:
```suggestion
                if (!is_callable($controller, true, $controllerName)) {
                    $controllerName = 'sub-request';
                }
```

trigrab: I like your suggestion, as it is cleaner. I am just not certain if we can assume that every controller out there passes the `is_callable` check. 
But I don't want to create an academic discussion here and therefor slow down this fix. 
Can you maybe just elaborate a bit on your suggestion?

umulmrum: The problem with applying this suggestion is that there is no $controllerName when it is a callable.

But I think you are right in suggesting to listen to an event (or hooking to the end of the dispatch method - I'm not familiar with C-level PHP, so can't see what is possible and makes sense). The moment the actual controller is definitely resolved is when returning from the dispatching of the ControllerArgumentsEvent. At this point it is also guaranteed that the controller is callable (Symfony will throw an exception if it is not, but this should hopefully only ever happen in a dev environment, so should be ok to ignore here). This would avoid getting problems with any custom logic.

But looking at the resolved callable would mean that the original if-else construct in the PR is needed to get meaningful values.

Nevay: > that there is no $controllerName when it is a callable

`is_callable()` sets `$controllerName`

trigrab: @Nevay that means you still think your suggestion is the best solution?

I want to move this forward. If you think that is the best solution I will apply it so that we can go on.

brettmc: I think you should try Nevay's suggestion, and add integration tests to confirm all the variances of $controller work as expected

trigrab: I applied the change and added integarion tests. There is one test that would have a different result then the initial solution. 
Now the name falls back to `sub-request` for anonymous classes. In the inital solution it was the class name. But I think that's fine.
Thanks for your support.
If there is anything else that I can do here, please let me know.

brettmc: minor CI break to be looked at before we can see the test results.

trigrab: I changed the tests because psalm was unhappy. I think now it should work. Can you trigger the ci one more time?

trigrab: Thanks for triggering the CI again. Now everything is green.
Is there still anything to do from my side?

brettmc (approved):

PR #369, opened 2025-05-13
update instrumentations to semconv 1.32

brettmc:
SemConv 1.32 deprecates a couple of attributes which we use extensively:
- code.function.name is now an FQN (including class name)
- code.namespace is deprecated (included in code.function.name)
- code.linenumber is deprecated and replaced with code.line.number

Updated all packages to conform, updated min semconv to 1.32, and made a couple of minor changes"


brettmc:
Note that Curl and OpenAIPHP look to be broken in main, so unrelated to this PR.

ChrisLightfootWild (approved):

PR #374, opened 2025-05-21
add Doctrine span links and more SemConv

brettmc:
- track doctrine prepared statements, and when they are executed create a span link back to the span that prepared.
- add some more SemConv to db spans, particularly db.operation.name
- align span names with how PDO does it (eg Doctrine::prepare)

Fixes: https://github.com/open-telemetry/opentelemetry-php/issues/1578

### DIFF
--- a/src/Instrumentation/Doctrine/src/DoctrineTracker.php
+++ b/src/Instrumentation/Doctrine/src/DoctrineTracker.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Contrib\Instrumentation\Doctrine;
+
+use Doctrine\DBAL\Driver\Statement;
+use OpenTelemetry\API\Trace\SpanContextInterface;
+use WeakMap;
+use WeakReference;
+
+/**
+ * @internal
+ */
+class DoctrineTracker
+{
+    private WeakMap $statementToSpanContextMap;
+
+    public function __construct()
+    {
+        $this->statementToSpanContextMap = new WeakMap();
+    }
+
+    public function trackStatement(Statement $statement, SpanContextInterface $context): void
+    {
+        $this->statementToSpanContextMap[$statement] = WeakReference::create($context);
+    }
+
+    public function getSpanContextForStatement(Statement $statement): ?SpanContextInterface
+    {
+        return $this->statementToSpanContextMap[$statement]?->get();
+    }
+}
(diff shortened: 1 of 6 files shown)


brettmc:
@open-telemetry/php-approvers ping!

src/Instrumentation/Doctrine/src/DoctrineTracker.php
```diff
@@ -0,0 +1,33 @@
...
+ * @internal
+ */
+class DoctrineTracker
+{
+    private WeakMap $statementToSpanContextMap;
+
+    public function __construct()
+    {
+        $this->statementToSpanContextMap = new WeakMap();
+    }
```
ChrisLightfootWild:
```suggestion
    public function __construct(
        private WeakMap $statementToSpanContextMap = new WeakMap(),
    ) {
    }
```

ChrisLightfootWild (approved):

src/Instrumentation/Doctrine/src/DoctrineTracker.php
```diff
@@ -0,0 +1,33 @@
...
+
+    public function trackStatement(Statement $statement, SpanContextInterface $context): void
+    {
+        $this->statementToSpanContextMap[$statement] = WeakReference::create($context);
```
Nevay: This looks like it will return `null` after the `::prepare()` span is flushed and exported? Should be testable by something like:
```php
$stmt = $connection->prepare('SELECT * FROM `technology` WHERE name = :name');
$this->storage->exchangeArray([]);

$stmt->bindValue('name', 'PHP');
$stmt->executeQuery();
$execute = $this->storage->offsetGet(0);
$this->assertCount(1, $execute->getLinks());
```

```suggestion
        $this->statementToSpanContextMap[$statement] = $context;
```

brettmc: Yes, you're right. I've used SplObjectStorage now, and added a test. It doesn't look like prepared statements can be destroyed easily, so I guess holding on to span context forever is okay...

Nevay: Should revert the `WeakMap`->`SplObjectStorage` change to avoid leaking memory; only the usage of `WeakReference` was problematic.

brettmc: Reverted to `WeakMap`

PR #377, opened 2025-05-22
update rule-based sampler to config 0.4

brettmc:
- update to work with config 0.4
- rename key to php_rule_based
- update documentation


src/Sampler/RuleBased/src/ComponentProvider/SamplerRuleBased.php
```diff
@@ -49,17 +50,17 @@ public function createPlugin(array $properties, Context $context): SamplerInterf
         );
     }
 
-    public function getConfig(ComponentProviderRegistry $registry): ArrayNodeDefinition
+    public function getConfig(ComponentProviderRegistry $registry, NodeBuilder $builder): ArrayNodeDefinition
     {
-        $node = new ArrayNodeDefinition('contrib_rule_based');
+        $node = $builder->arrayNode('php_rule_based');
```
brettmc, line 55: My thinking here is that this is less likely to conflict with a sampler from another SIG, given that they would be unlikely to accept the same config

bobstrecansky (approved):

PR #379, opened 2025-05-27
[ReactPHP] Add user-option for redacting query parameters

smaddock:
Related to the general issue tracked in https://github.com/open-telemetry/semantic-conventions/issues/877

There have been related PRs for OpenTelemetry Semantic Conventions in an attempt to define this, but it seems they were too broad to gain consensus. However, I have not seen any comments contrary to the specific feature implemented by this PR, exemplified here:

> For users who want to change the default behavior, OpenTelemetry should provide a consistent way instead of let the user struggle in a jungle of 20+ instrumentation libraries where each have different ways of doing redaction. [ref](https://github.com/open-telemetry/semantic-conventions/pull/961#issuecomment-2079761854)

and here:

>> we discussed the idea of having generalized SDK capabilities for redaction, is this still something we want? If so I am happy to work towards a proposal for that.
> 
> I'd definitely answer that with a yes. Users might want to have redaction capabilities they can use for data that doesn't conform to OTel semantic conventions. [ref](https://github.com/open-telemetry/semantic-conventions/pull/971#issuecomment-2082811589)

Since that has not yet come to fruition, but these is still a need, I propose implementing this PR for at a minimum this PHP auto-instrumentation library, with the goal of aligning the mechanism with what SemConv decides in the future.

Note that this PR does not allow for overriding the required query keys defined in footnote 6 here: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/url.md#url-query but allows the user of this library to add additional keys to the redaction list.

Slack conversation specifically for this PR: https://cloud-native.slack.com/archives/C01NFPCV44V/p1748367527296729

### DIFF
--- a/src/Instrumentation/ReactPHP/src/ReactPHPInstrumentation.php
+++ b/src/Instrumentation/ReactPHP/src/ReactPHPInstrumentation.php
@@ -55,6 +55,15 @@ class ReactPHPInstrumentation
      */
     private const ENV_HTTP_RESPONSE_HEADERS = 'OTEL_PHP_INSTRUMENTATION_HTTP_RESPONSE_HEADERS';
     /**
+     * The environment variable which adds to the URL query parameter keys to redact the values for.
+     * This supports a comma-separated list of case-sensitive known HTTP methods.
+     *
+     * Note that this is not currently defined in OTel SemConv, and therefore subject to change.
+     *
+     * @see https://github.com/open-telemetry/semantic-conventions/issues/877
+     */
+    private const ENV_URL_SANITIZE_FIELD_NAMES = 'OTEL_PHP_INSTRUMENTATION_URL_SANITIZE_FIELD_NAMES';
+    /**
      * The `{method}` component of the span name when the original method is not known to the instrumentation.
      *
      * @see https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name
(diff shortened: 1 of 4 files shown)


src/Instrumentation/ReactPHP/src/ReactPHPInstrumentation.php
```diff
@@ -54,6 +54,15 @@ class ReactPHPInstrumentation
      * @see https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span
      */
     private const ENV_HTTP_RESPONSE_HEADERS = 'OTEL_PHP_INSTRUMENTATION_HTTP_RESPONSE_HEADERS';
+    /**
+     * The environment variable which adds to the URL query parameter keys to redact the values for.
+     * This supports a comma-separated list of case-sensitive known HTTP methods.
```
Nevay, line 59:
```suggestion
     * This supports a comma-separated list of case-sensitive query parameter keys.
```

src/Instrumentation/ReactPHP/src/ReactPHPInstrumentation.php
```diff
@@ -259,6 +268,12 @@ private static function sanitizeUrl(UriInterface $uri): string
             $uri = $uri->withUserInfo(self::URL_REDACTION);
         }
 
+        $sanitizeFields = self::URL_QUERY_REDACT_KEYS;
+        $customFields = $_ENV[self::ENV_URL_SANITIZE_FIELD_NAMES] ?? '';
+        if (!empty($customFields)) {
```
Nevay, line 273:
```suggestion
        if ($customFields !== '') {
```

smaddock:
@Nevay are you saying it would be redundant for me to extract my current implementation to a shared utility, and eventually you’ll have something committed that I could replace mine with?

smaddock:
These checks should pass now with #371 merged.

PR #382, opened 2025-06-04
New resource detector for DigitalOcean

smaddock:
There are only a couple existing contrib resource detectors, so I could be doing something wildly off here.

Current things I'm iffy on:

- I've seen notes about abandoning Composer's `file` autoload and replacing with @Nevay's SPI loader, but I wasn't sure if that was possible to implement now.
- I opened https://github.com/open-telemetry/opentelemetry-php/issues/1617; if that happens this class will need to be updated (not a big deal).
- Similarly I used some odd methodology for the PHPUnit tests to get around dependencies on `file_get_contents` and `error_log`... I've seen a few SDK utility methods built around these, but the readfile was private so not reusable in the production code, and I was unsure how to best handle the SDK logging for PHPUnit.
- I'm taking advantage of API credentials if they are provided, as DigitalOcean doesn't expose nearly as much "for free" as the larger cloud providers, but it's also not required... if it's not okay, though, I'll strip it out. I don't think I'll ever use it myself, just thought if it was available I'd make it work.
- I'm sure there's more I'll remember later.

I'm marking this as draft until I get a chance to test it live.


brettmc:
> I've seen notes about abandoning Composer's file autoload and replacing with @Nevay's SPI loader, but I wasn't sure if that was possible to implement now.

For now, stick with autoload->files mechanism. It's pretty trivial to switch when we get to that part (and autoload->files can be backwards-compatible to SDK 1.x)

smaddock:
This should be good to merge now.

brettmc (approved):

PR #383, opened 2025-06-05
Add http configuration package

Nevay:
Provides general configuration options for HTTP instrumentation packages.

Env-based configuration supports:
- `OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS`
- `OTEL_PHP_INSTRUMENTATION_URL_SANITIZE_FIELD_NAMES` from https://github.com/open-telemetry/opentelemetry-php-contrib/pull/379

File-based configuration supports:
- overriding known HTTP methods
- overriding URI sanitization
- enabling opt-in attributes for server and client spans

Depends on https://github.com/open-telemetry/opentelemetry-php/pull/1621 to load configuration from config file or env.


src/Instrumentation/HttpConfig/composer.json
```diff
@@ -0,0 +1,61 @@
...
+    "psr/http-message": "^2.0"
+  },
+  "require-dev": {
+    "friendsofphp/php-cs-fixer": "^3.7",
+    "league/uri": "^7.5",
+    "phan/phan": "^5.4",
+    "phpstan/phpstan": "^2.1",
+    "phpstan/phpstan-phpunit": "^2.0",
+    "psalm/plugin-phpunit": "^0.19.3",
+    "open-telemetry/api": "^1.4",
```
ChrisLightfootWild: I *think* the API dependency should be promoted from dev :thinking:

Nevay: Moved to `require`.

The API dependency is only needed for the file-based and env-based configuration classes -> dependency was originally in `require-dev` to allow adopting `UriSanitizer` without forcing an API bump. Moved to `require` as API `1.4.0` was released over half a year ago.

brettmc (approved):

src/Instrumentation/HttpConfig/composer.json
```diff
@@ -0,0 +1,61 @@
...
+  "description": "OpenTelemetry HTTP configuration options",
+  "keywords": ["opentelemetry", "otel", "open-telemetry", "instrumentation"],
+  "type": "library",
+  "homepage": "https://opentelemetry.io/docs/php",
+  "readme": "./README.md",
+  "license": "Apache-2.0",
+  "minimum-stability": "dev",
+  "prefer-stable": true,
+  "require": {
+    "php": "^8.2",
```
brettmc, line 12: since this requires 8.2, 8.1 should be dropped from the test matrix

Nevay: Should IMO be done in a separate PR for all instrumentations as `8.1` is EOL.

PR #384, opened 2025-06-07
Update dependa-bot and composer

weslenteche:


brettmc:
> If the goal is consistency for all contrib packages

@weslenteche will you include those other suggestions? If not, I'm happy to accept as-is

smaddock:
@brettmc I can do a separate PR for those

weslenteche:
> > If the goal is consistency for all contrib packages
> 
> @weslenteche will you include those other suggestions? If not, I'm happy to accept as-is

Hi everyone, the last week was a bit busy at work and I ended up not having time to adapt the suggestions from @smaddock 

Thank you very much for following up on the suggestions in other pull requests.

PR #385, opened 2025-06-09
Fix: Handle null request in CakePHP Server::run hook

tyler36:
## Issue

On CakePHP `5.2.4`, I get the following error:

```
Warning: Undefined array key 0 in /var/www/html/vendor/open-telemetry/opentelemetry-auto-cakephp/src/Hooks/Cake/Http/Server.php on line 29
```

## Solution

This PR introduces a null check for `$params[0]` before attempting to cast it to a `ServerRequestInterface`. If `$params[0]` is null, the `$request` variable is set to null, preventing the error and allowing the instrumentation to continue gracefully (albeit without request-specific attributes).


tyler36:
Appolgies for the noise. 
I renamed the branch and it closed the PR.

bobstrecansky (approved):

brettmc (approved):

PR #387, opened 2025-06-10
handle guzzle response exceptions

brettmc:
if a guzzle client is configured with exceptions=true, try to retrieve the http status code, body size, and network protocol version and set as span attributes

Closes: https://github.com/open-telemetry/opentelemetry-php/issues/1612

### DIFF
--- a/src/Instrumentation/Guzzle/tests/Integration/GuzzleInstrumentationTest.php
+++ b/src/Instrumentation/Guzzle/tests/Integration/GuzzleInstrumentationTest.php
@@ -6,6 +6,7 @@ namespace OpenTelemetry\Instrumentation\HttpAsyncClient\tests\Integration;
 
 use ArrayObject;
 use GuzzleHttp\Client;
+use GuzzleHttp\Exception\ConnectException;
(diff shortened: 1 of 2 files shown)


src/Instrumentation/Guzzle/tests/Integration/GuzzleInstrumentationTest.php
```diff
@@ -165,4 +166,47 @@ public function test_headers_propagation(): void
...
+            'http_errors' => true,
+            'exceptions' => true,
+        ]);
+        $this->mock->append($response);
+        $this->assertCount(0, $this->storage);
+
+        try {
+            $client->send(new Request('GET', 'https://example.com/error'));
+        } catch (\Exception $e) {
+            // Expected exception
```
cedricziel, line 187: i wonder if we could `expectException` here

brettmc: I think not, because it would bail out if we didn't catch, and not run all the code after this point.

bobstrecansky (approved):

PR #390, opened 2025-06-12
fix: openentelemetry-auto-laravel threw warning on \Illuminate\Database\Eloquent\Model::destroy called.

rozeo:
see more details: https://github.com/open-telemetry/opentelemetry-php/issues/1630
I checked to see if other functions were also affected, but only this function was affected.

Closes: https://github.com/open-telemetry/opentelemetry-php/issues/1630

### DIFF
--- a/src/Instrumentation/Laravel/tests/Integration/LaravelInstrumentationTest.php
+++ b/src/Instrumentation/Laravel/tests/Integration/LaravelInstrumentationTest.php
@@ -165,7 +165,6 @@ class LaravelInstrumentationTest extends TestCase
         $this->assertSame('test_models', $updateSpan->getAttributes()->get('laravel.eloquent.table'));
         $this->assertSame('update', $updateSpan->getAttributes()->get('laravel.eloquent.operation'));
         
-	
         // Delete span
         $deleteSpan = array_values(array_filter($eloquentSpans, function ($span) {
             return $span->getAttributes()->get('laravel.eloquent.operation') === 'delete';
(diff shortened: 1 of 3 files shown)


src/Instrumentation/Laravel/tests/Integration/LaravelInstrumentationTest.php
```diff
@@ -102,16 +102,16 @@ public function test_eloquent_operations(): void
...
-                
+
                 // Test update
                 $found->update(['name' => 'updated']);
-                
+
                 // Test delete
                 $found->delete();
-                
+
```
rozeo, line 114: PHPStorm has removed whitespace from lines consisting only of whitespace.

brettmc: No problem. I would have thought php-cs-fixer would have picked this up earlier...

rozeo:
I am still checking about the CLA...

brettmc (approved):

src/Instrumentation/Laravel/tests/Integration/LaravelInstrumentationTest.php
```diff
@@ -186,6 +175,58 @@ public function test_eloquent_operations(): void
...
+        // Create the test_models table
+        DB::statement('CREATE TABLE IF NOT EXISTS test_models (
+            id INTEGER PRIMARY KEY AUTOINCREMENT,
+            name TEXT,
+            created_at DATETIME,
+            updated_at DATETIME
+        )');
+
+        $this->router()->get('/eloquent', function () {
```
ChrisLightfootWild, line 191: Is there any reason to use the HTTP kernel to test model functionality? 🤔

rozeo: Thank you for review!

The main reason is to test in the same way as much as possible, because test cases with similar intentions were defined in other cases and added as additional cases, but as you reviewed, the TraceProvider setting is done on the TestCase class, so it seems that the test itself is possible without going through the HTTP kernel. 
However, it is defined as an integration test, I personally think the current way makes more sense, since the test is to be conducted in the same way as the actual use case and a database is required in this case. 
But a little redundant.

If a better testing method is exists, I will be create a new PR.

ChrisLightfootWild: Thanks @rozeo - we can split these out later then 👍

PR #391, opened 2025-06-16
Pdo metrics

GrzegorzDrozd:
- add duration tracking
- add returned rows tracking
- add pending operations metric
- add active connections metric
- minor code fixes


brettmc:
@GrzegorzDrozd I've finally had some time to look at metrics generation. I started down the path of tracking everything in pre and post hooks, in #396 - when I got something working I started to realise that I was duplicating a lot of effort from tracing (and some attributes were really hard to get for a metric since you can start a span and add to it while it's in progress, whereas a metric is a single call).

Since in my case (an HTTP server span), everything I need seems to be already stored in a span, I tried a different approach (I guess loosely based on [span metrics connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector) and implemented a Span Processor to emit metrics: https://github.com/open-telemetry/opentelemetry-php/pull/1651/files#diff-9ed562697c2c477790ec9394f926eeb261f2433f39fb3eeca3e7999ee7c390dbR54

Looking at your implementation here, I see a similar pattern:
1. start a span
2. start some timers and track some objects
3. end a span
4. calculate duration/s and counts from timers and trackers, gather attributes, emit metric/s

The metrics are closely related to the spans generated nearby.

Given that duration and most of the attributes you're interested in are already in a span, do you think you could achieve your goals with a span processor?

GrzegorzDrozd:
> @GrzegorzDrozd I've finally had some time to look at metrics generation. I started down the path of tracking everything in pre and post hooks, in #396 - when I got something working I started to realise that I was duplicating a lot of effort from tracing (and some attributes were really hard to get for a metric since you can start a span and add to it while it's in progress, whereas a metric is a single call).
> 
> Since in my case (an HTTP server span), everything I need seems to be already stored in a span, I tried a different approach (I guess loosely based on [span metrics connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector) and implemented a Span Processor to emit metrics: https://github.com/open-telemetry/opentelemetry-php/pull/1651/files#diff-9ed562697c2c477790ec9394f926eeb261f2433f39fb3eeca3e7999ee7c390dbR54
> 
> Looking at your implementation here, I see a similar pattern:
> 
>     1. start a span
> 
>     2. start some timers and track some objects
> 
>     3. end a span
> 
>     4. calculate duration/s and counts from timers and trackers, gather attributes, emit metric/s
> 
> 
> The metrics are closely related to the spans generated nearby.
> 
> Given that duration and most of the attributes you're interested in are already in a span, do you think you could achieve your goals with a span processor?

But that would force someone to use spans with metrics? What if they want only to use metrics? What about sampling?

brettmc:
> But that would force someone to use spans with metrics? What if they want only to use metrics? What about sampling? So with this argument I think I should move my code outside of hooks with spans and create separate code, so that we would allow using metrics without spans ...

Yes it would...but we're creating spans anyway (even if they're no-op). Anyway, I'm not saying we have to do it that way, I just want to make sure we _need_ the extra tracking/timing classes before accepting the ongoing maintenance.

src/Instrumentation/PDO/src/PDOInstrumentation.php
```diff
@@ -73,22 +76,41 @@ public static function register(): void
                     }
 
                     self::end($exception);
+                    
+                    $attributes = $pdoTracker->get($result);
+                    $parent = Context::getCurrent();
+
+                    $instrumentation->meter()
+                        ->createUpDownCounter('db.client.connection.count', '1')
```
brettmc: Should we be creating a new instrument each time, or one that is reused?

src/Instrumentation/PDO/src/PDOInstrumentation.php
```diff
@@ -327,6 +433,39 @@ private static function isDistributeStatementToLinkedSpansEnabled(): bool
...
+    }
+
+    protected static function createDurationMetric(
+        CachedInstrumentation $instrumentation,
+        array $attributes,
+        float $value,
+    ): void {
+        $parent = Context::getCurrent();
+        $instrumentation->meter()
+            ->createHistogram('db.client.operation.duration', 'ms')
```
alanwest, line 464: The [`db.client.operation.duration` metric](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#metric-dbclientoperationduration) should use seconds `s` as the unit.

GrzegorzDrozd:
Hi @brettmc

Question (I am tagging you because you answered in my PR but if you think this is a topic for more wide audience please tag people that need to see it):
There is not much metric related code in contrib modules, and because of that: do we want to add metrics to existing span processing or create separate files to allow developers to chose if they want to use metrics, spans or metrics + spans? 
In this example: PDOInstrumentation should contain code for both or should I create PDOMetrics with separate hook and register and deal with metrics just there? 
I have no idea how other languages handle those :/ I could not find much in go-contrib, java-contrib. In js-contrib i see that PostgreSQL exposes duration and code is included in main instrumentation flow.

If you wish I can create a thread in Slack for more interactive discussion.

src/Instrumentation/PDO/src/PDOInstrumentation.php
GrzegorzDrozd: All examples I found always created new meter. I assume this is a cheap operation. Re-using that would be much more complicated I think.

PR #394, opened 2025-06-18
Improve openentelemetry-auto-laravel test definitions for Eloquent hooks.

rozeo:
Based on [this review](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/390#discussion_r2144455109), I try to split the test cases related to Database\Eloquent\Model Hooks to a smaller dependency and simple test cases.
(Also added test cases for hooks that were not tested.)


src/Instrumentation/Laravel/tests/Integration/Database/Eloquent/ModelTest.php
```diff
@@ -0,0 +1,189 @@
...
+        ');
+    }
+
+    public function tearDown(): void
+    {
+        // Reset table
+        DB::statement('DROP TABLE IF EXISTS test_models');
+
+        parent::tearDown();
+    }
```
ChrisLightfootWild: Does the in-memory DB get dropped between tests without this? 🤔

ChrisLightfootWild (approved):
This is great, thanks for tidying it up! ⭐

src/Instrumentation/Laravel/tests/Integration/Database/Eloquent/ModelTest.php
rozeo: It was implemented to explicitly initialize the tables for each test case, but the in-memory DB behavior was already fully initialized for each test case. I  will remove tearDown. Thanks! 😄

PR #402, opened 2025-07-08
Update monorepo meta files

smaddock:
Followup to this comment: https://github.com/open-telemetry/opentelemetry-php-contrib/pull/384#issuecomment-2951550652

If any of these changes are undesirable let me know and I'll update the PR.

Also, I noticed the [src/Aws/](src/Aws/) and [src/ResourceDetectors/Azure/](src/ResourceDetectors/Azure/) libraries have inconsistent namespaces with the rest of the project; should those be updated? They would be breaking changes, but it looks like neither has hit a stable semantic version yet.


smaddock:
Reverted the last commit; `AutoInstrumentationInstaller` and `MetaPackages/opentelemetry` will never pass CI tests

composer.json
```diff
@@ -15,12 +15,14 @@
   },
   "autoload": {
     "psr-4": {
-      "OpenTelemetry\\Contrib\\Aws\\": "src/Aws/src",
+      "OpenTelemetry\\Aws\\": "src/Aws/src",
```
brettmc: This one looks odd, I think that it should still contain `Contrib` in the namespace...

edit: oh, it's to match current reality. Yeah, I think those two packages' namespaces should be updated.

smaddock: Agreed, I was updating it to match the subpackage composer.json namespace, which I think needs updating: https://github.com/open-telemetry/opentelemetry-php-contrib/blob/872475913cfc30684bc302a0b74890a5897cee9b/src/Aws/composer.json#L36 
(that's what my question at the end of the PR comment was about)

smaddock: Should I update those here or in separate PRs?

smaddock: If #403 and #404 are accepted, this should be resolved with 7786aa8

smaddock:
Brett has mentioned they’re used in the Packagist deployment, at least.

brettmc:
> The dev-tools on the root composer - I've never used them but would presume it was there for something.

It's used here for one function that I can see, `make packages-composer`.

> but even the composer.json that’s part of this PR… is it ever used/needed?

I don't think that it is used nowadays. I _think_ we could get rid of it. :crossed_fingers:

brettmc (approved):

PR #403, opened 2025-07-08
Update AWS namespace

smaddock:
Following conversation here: https://github.com/open-telemetry/opentelemetry-php-contrib/pull/402#discussion_r2193623703


brettmc (approved):

ChrisLightfootWild:
> > Please could you add `/examples` to `.gitattributes` export-ignore list? 👍
> 
> @ChrisLightfootWild should all examples be export-ignored? E.g., `/src/**/examples`
> 
> Edit: I see the package has its own .gitattributes, which is likely what you were referring to; my question applies to the monorepo .gitattributes.

You got it 👍 

I think we can ignore the root level one as that package is abandoned:
https://packagist.org/packages/open-telemetry/opentelemetry-php-contrib

ChrisLightfootWild (approved):

PR #405, opened 2025-07-10
[sampler-xray]Added AWS X-Ray Sampler implementation

AsakerMohd:
## Description:
Currently, there isn't any implementation of the AWS X-Ray Remote Sampler that allows customers to fetch their sampling rules and sample accordingly. This PR adds the implementation for that sampler. The customer is able to configure their sampling rules through AWS CloudWatch: https://docs.aws.amazon.com/xray/latest/devguide/xray-console-sampling.html. This implementation fetches the sampling rules using a set polling interval which is set when creating a new instance of AWSXRayRemoteSampler. This implementation is based on similar sampler implementations in other OTEL languages.
- .NET for example: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Sampler.AWS

The sampling rule has two components: a reservoir and a fixed sampling rate. The idea here is once the sampling rule is matched and the reservoir is exhausted, the spans are sampled at a fixed rate according to their trace ids. As a result, the remote sampler samples the spans using two samplers: A reservoir sampler and a fixed rate sampler and chooses which one according to the reservoir quota.

### Unit Testing:
Wrote unit tests for the major classes to assert for correctness and they're all passing.
### Manual Testing:
Wrote a sample app with manual instrumentation and hooked up the X-Ray Sampler to the Tracer. I then verified that requests are being sampled as expected using different sampling rules.


src/Sampler/Xray/src/Clock.php
ChrisLightfootWild, line 1: The API has a clock, which looks like it should give you what you need:

```php
(int) (\OpenTelemetry\API\Common\Time\Clock::getDefault()->now() / \OpenTelemetry\API\Common\Time\ClockInterface::NANOS_PER_MILLISECOND)
```

src/Sampler/Xray/src/SamplingRuleApplier.php
```diff
@@ -0,0 +1,199 @@
...
+                if ($this->borrowing) {
+                    $this->statistics->borrowCount++;
+                }
+                $this->statistics->sampleCount++;
+
+                return $res;
+            }
+        }
+        $res = $this->fixedRateSampler->shouldSample($parentContext, $traceId, $spanName, $spanKind, $attributes, $links);
+        ;
```
ChrisLightfootWild:
```suggestion

```

src/Sampler/Xray/src/SamplingStatisticsDocument.php
```diff
@@ -0,0 +1,26 @@
...
+    
+    public function __construct(string $clientId, string $ruleName, int $req, int $samp, int $borrow, float $ts)
+    {
+        $this->ClientID     = $clientId;
+        $this->RuleName     = $ruleName;
+        $this->RequestCount = $req;
+        $this->SampleCount  = $samp;
+        $this->BorrowCount  = $borrow;
+        $this->Timestamp    = $ts;
+    }
```
ChrisLightfootWild: Given this package supports PHP ^8.1:

```suggestion
    public function __construct(
        public string $ClientID,
        public string $RuleName,
        public int $RequestCount,
        public int $SampleCount,
        public int $BorrowCount,
        public float $Timestamp,
    ) {
    }
```

src/Sampler/Xray/src/SamplingRule.php
```diff
@@ -0,0 +1,63 @@
...
+        $this->ReservoirSize= $reservoirSize;
+        $this->Host         = $host;
+        $this->HttpMethod   = $httpMethod;
+        $this->ResourceArn  = $resourceArn;
+        $this->ServiceName  = $serviceName;
+        $this->ServiceType  = $serviceType;
+        $this->UrlPath      = $urlPath;
+        $this->Version      = $version;
+        $this->Attributes   = $attributes;
+    }
```
ChrisLightfootWild:
```suggestion
    public function __construct(
        public string $RuleName,
        public int $Priority,
        public float $FixedRate,
        public int $ReservoirSize,
        public string $Host,
        public string $HttpMethod,
        public string $ResourceArn,
        public string $ServiceName,
        public string $ServiceType,
        public string $UrlPath,
        /** @psalm-suppress PossiblyUnusedProperty */
        public int $Version,
        public array $Attributes,
    ) {
    }
```

src/Sampler/Xray/src/SamplingRuleApplier.php
```diff
@@ -0,0 +1,199 @@
...
+        if ($rule->ReservoirSize > 0) {
+            $this->reservoirSampler = new RateLimitingSampler($rule->ReservoirSize);
+            $this->borrowing = true;
+        } else {
+            $this->reservoirSampler = new AlwaysOffSampler();
+            $this->borrowing = false;
+        }
+        
+        $this->fixedRateSampler = new TraceIdRatioBasedSampler($rule->FixedRate);
+        $this->reservoirEndTime = new \DateTimeImmutable('@' . PHP_INT_MAX);
```
ChrisLightfootWild: I've seen this done elsewhere in the PR as:
`new \DateTimeImmutable('9999-12-31T23:59:59+00:00');`

AsakerMohd: Using the otel clock now.

src/Sampler/Xray/src/SamplingRule.php
AsakerMohd: fixed

src/Sampler/Xray/src/SamplingStatisticsDocument.php
AsakerMohd: fixed

src/Sampler/Xray/src/SamplingRuleApplier.php
AsakerMohd: fixed

src/Sampler/Xray/src/Clock.php
AsakerMohd: thanks for calling that out. Updated all the classes to use the otel clock instead since it's basically the same functionality.

brettmc:
Please add .gitsplit config so it can be published

PR #410, opened 2025-07-24
Record exception with stack trace on log records

ol1s:
Currently, exceptions on spans ARE presenting a useful stack trace whereas exceptions in logs ARE NOT presenting the stack trace.

As per otel semantic conventions:

> Exceptions SHOULD be recorded as attributes on the LogRecord passed to the Logger emit operations

https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/

This proposal records the exception with stack trace on the log record, using a similar implementation to the `Span.php`

### DIFF
--- a/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
+++ b/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
@@ -54,9 +54,21 @@ class LogWatcher extends Watcher
             // Should this fail, we should continue to emit the LogRecord.
         }
 
+        $contextToEncode = array_filter($log->context);
+
+        $exception = $this->getExceptionFromContext($log->context);
+
+        if ($exception !== null) {
+            unset($contextToEncode['exception']);
+        }
+
         $attributes = [
-            'context' => json_encode(array_filter($log->context)),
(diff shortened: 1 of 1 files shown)


src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
```diff
@@ -54,6 +56,7 @@ public function recordLog(MessageLogged $log): void
 
         $attributes = [
             'context' => json_encode(array_filter($log->context)),
+            ...$this->parseExceptionFromContext($log->context),
```
ChrisLightfootWild: It probably makes sense to remove the `exception` from the `context` key to avoid the duplication.

ol1s: I have changed that

brettmc (approved):

src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
ChrisLightfootWild: Nice work @ol1s 💪

PR #411, opened 2025-07-26
pgsql driver auto-instrumentation

jf-leblancrichard:


intuibase:
Hey
Just wanted to mention that I'm working on same instrumentation here:
https://github.com/intuibase/opentelemetry-php-contrib/blob/postgresql_instrumentation/src/Instrumentation/PostgreSql/src/PostgreSqlInstrumentation.php

I just have to implement unit tests.

PR #414, opened 2025-07-29
Update wordpress span Name to include low cardinal target

HeenaBansal20:
Fixes https://github.com/open-telemetry/opentelemetry-php/issues/1671
However this change is  required for multiple instrumented libraries.

### DIFF
--- a/src/Instrumentation/Wordpress/composer.json
+++ b/src/Instrumentation/Wordpress/composer.json
@@ -14,7 +14,8 @@
     "open-telemetry/api": "^1.0",
     "open-telemetry/sem-conv": "^1.32",
     "nyholm/psr7": "^1",
-    "nyholm/psr7-server": "^1"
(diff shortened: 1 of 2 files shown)


Soean:
I tested it in my local environment, works for me. ✅ 

I have also been considering an alternative value. We could have changed the name in the shutdown function. But there is not really a good value in WordPress.

 For `/example/`:
```php
<?php
register_shutdown_function(function () use ($span) {
  
  ...
  
  global $wp;
  echo $wp->query_string; // ‘pagename=example’
  echo $wp->request; // ‘example’
  echo $wp->matched_rule; // ‘(.?.+?)(?:/([0-9]+))?/?$’
  echo $wp->matched_query; // ‘pagename=example&page=’
  
  ...
  
  $span->updateName( ... )`

```

▶️ We should go with `SCRIPT_NAME` for now.

brettmc (approved):

PR #422, opened 2025-08-06
Fix/code igniter span name as per otel spec

HeenaBansal20:


src/Instrumentation/CodeIgniter/src/CodeIgniterInstrumentation.php
```diff
@@ -58,12 +61,14 @@ public static function register(): void
             ) use ($instrumentation, $requestProperty): void {
                 $extractedRequest = $requestProperty->getValue($igniter);
                 $request = ($extractedRequest instanceof RequestInterface) ? $extractedRequest : null;
-
+                 
+                // Get the HTTP method from the request and store it for later use
+                self::$httpMethod = $request ? $request->getMethod() : 'unknown';
```
ChrisLightfootWild: How about using the available `$_SERVER` request method, even as part of fallback? Not sure it is ever missing, but kept the `unknown`.

```suggestion
                self::$httpMethod = $request?->getMethod() ?? $_SERVER['REQUEST_METHOD'] ?? 'unknown';
```

Edit: use null-safe operator given we are in PHP 8.1+.

HeenaBansal20: I Agree @ChrisLightfootWild ,  fallback to $_SERVER['REQUEST_METHOD'] is smart move. :-)

HeenaBansal20:
@brettmc @ChrisLightfootWild Thanks for reviewing the PR. Is it good to merge to master ?
If yes , please merge 
Thanks .

PR #433, opened 2025-08-27
Auto instrumentation hooks for PHP Session 

HeenaBansal20:
This PR adds auto instrumentation hooks for session_start() and session_destroy().


src/Instrumentation/Session/_register.php
```diff
@@ -0,0 +1,18 @@
...
+
+use OpenTelemetry\Contrib\Instrumentation\Session\PhpSessionInstrumentation;
+use OpenTelemetry\SDK\Sdk;
+
+if (class_exists(Sdk::class) && Sdk::isInstrumentationDisabled(PhpSessionInstrumentation::NAME) === true) {
+    return;
+}
+
+if (extension_loaded('opentelemetry') === false) {
+    trigger_error('The opentelemetry extension must be loaded in order to autoload the OpenTelemetry IO auto-instrumentation', E_USER_WARNING);
```
brettmc:
```suggestion
    trigger_error('The opentelemetry extension must be loaded in order to autoload the OpenTelemetry session auto-instrumentation', E_USER_WARNING);
```

.gitsplit.yml
```diff
@@ -60,6 +60,8 @@ splits:
     target: "https://${GH_TOKEN}@github.com/opentelemetry-php/contrib-auto-psr18.git"
   - prefix: "src/Instrumentation/ReactPHP"
     target: "https://${GH_TOKEN}@github.com/opentelemetry-php/contrib-auto-reactphp.git"
+  - prefix: "src/Instrumentation/Session"
+    target: "https://${GH_TOKEN}@github.com/opentelemetry-php/contrib-auto-php-session.git"
```
brettmc:
```suggestion
    target: "https://${GH_TOKEN}@github.com/opentelemetry-php/contrib-auto-session.git"
```

composer.json
```diff
@@ -156,6 +159,7 @@
     "open-telemetry/opentelemetry-auto-psr16": "self.version",
     "open-telemetry/opentelemetry-auto-psr18": "self.version",
     "open-telemetry/opentelemetry-auto-reactphp": "self.version",
+    "open-telemetry/opentelemetry-auto-php-session": "self.version",
```
brettmc, line 162:
```suggestion
    "open-telemetry/opentelemetry-auto-session": "self.version",
```

src/Instrumentation/Session/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Spans and Attributes
+
+### session.start
+
+When `session_start()` is called, a span named `session.start` is created with the following attributes:
+
+- `code.function_name`: The function name (`session_start`)
+- `code.file_path`: The file path where the function was called
+- `code.line_number`: The line number where the function was called
```
brettmc: since session function are provided by an extension, hence they're internal functions, you won't have access to file path or line number

src/Instrumentation/Session/README.md
```diff
@@ -0,0 +1,57 @@
...
+- `session.options.*`: Any options passed to `session_start()`
+- `session.id`: The session ID (if session was successfully started)
+- `session.name`: The session name (if session was successfully started)
+- `session.status`: Either "active" or "inactive"
+- `session.cookie.*`: Session cookie parameters
```
brettmc: these aren't official SemConv attributes, so perhaps they should be prefixed with `php.` ?

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
```diff
@@ -0,0 +1,134 @@
...
+class PhpSessionInstrumentation
+{
+    public const NAME = 'php-session';
+
+    public static function register(): void
+    {
+        $instrumentation = new CachedInstrumentation(
+            'io.opentelemetry.contrib.php.session',
+            null,
+            Version::VERSION_1_32_0->url(),
```
brettmc: Could you use the latest version of semconv (1.36.0)?

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
```diff
@@ -0,0 +1,134 @@
...
+        ?string $filename,
+        ?int $lineno
+    ): SpanBuilderInterface {
+        /** @psalm-suppress ArgumentTypeCoercion */
+        return $instrumentation->tracer()
+            ->spanBuilder($name)
+            ->setSpanKind(SpanKind::KIND_INTERNAL)
+            ->setAttribute(TraceAttributes::CODE_FUNCTION_NAME, $function)
+            ->setAttribute(TraceAttributes::CODE_FILE_PATH, $filename)
+            ->setAttribute(TraceAttributes::CODE_LINE_NUMBER, $lineno);
```
brettmc:
```suggestion
            ->setAttribute(TraceAttributes::CODE_FUNCTION_NAME, $function);
```
Will always be null for an internal function

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
```diff
@@ -0,0 +1,134 @@
...
+            return;
+        }
+        $scope->detach();
+        $span = Span::fromContext($scope->context());
+        
+        // Add session-specific attributes in the post hook
+        if ($function === 'session_start') {
+            $span->setAttribute('session.id', session_id());
+            $span->setAttribute('session.name', session_name());
+            $isSessionActive = !empty(session_id());
```
brettmc: would a session ever not be active after calling session_start ?

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
```diff
@@ -0,0 +1,134 @@
...
+            $span->setAttribute('session.status', $isSessionActive ? 'active' : 'inactive');
+
+            // Add session cookie parameters
+            $cookieParams = session_get_cookie_params();
+            foreach ($cookieParams as $key => $value) {
+                if (is_scalar($value)) {
+                    $span->setAttribute("session.cookie.$key", $value);
```
brettmc: could this leak sensitive data?

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
```diff
@@ -0,0 +1,134 @@
...
+                }
+            }
+        } elseif ($function === 'session_destroy') {
+            $span->setAttribute('session.destroy.success', $exception === null);
```
brettmc: I think it's more correct to use the span status to determine this

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: updated.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: updated.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: updated

src/Instrumentation/Session/README.md
HeenaBansal20: Agree!
Updated accordingly.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: you are right. I kept the cookie ket in the span and redacted the value from span attributes.
Does this sound okay ?

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
```diff
@@ -0,0 +1,161 @@
...
+            $isSessionActive = !empty(session_id());
+            $span->setAttribute('php.session.id', session_id());
+            $span->setAttribute('php.session.name', session_name());
+            $span->setAttribute('php.session.status', $isSessionActive ? 'active' : 'inactive');
+
+            // Add session cookie parameters
+            $cookieParams = session_get_cookie_params();
+            foreach ($cookieParams as $key => $value) {
+                if (is_scalar($value)) {
+                    $span->setAttribute("php.session.cookie.$key", '<redacted>');
```
brettmc: if the values will always be redacted, maybe you should just store the keys as an array in `php.session.cookie.keys` ?

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
```diff
@@ -0,0 +1,161 @@
...
+    {
+        $scope = Context::storage()->scope();
+        if (!$scope) {
+            return;
+        }
+        $scope->detach();
+        $span = Span::fromContext($scope->context());
+        
+        // Add session-specific attributes in the post hook
+        if ($function === 'session_start') {
```
brettmc: this big if-else could be modernised to a switch statement?

HeenaBansal20: Agree! I'll update.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: yeah agree.. that would be ideal.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: I updated the span attributes to store only session keys,.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: I updated attributes to store only keys.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: done.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: @brettmc , yes I think it is possible and one of the edge case when headers are already sent before session_start and can be checked with session_status === [PHP_SESSION_ACTIVE](https://www.php.net/manual/en/session.constants.php#constant.php-session-active).

However , from the testcase , I am unable to make the session_status to return PHP_SESSION_ACTIVE since in CLI sessions behave differently it seems it always return PHP_SESSION_NONE.

brettmc: Could you inspect the return value of [session_start](https://www.php.net/manual/en/function.session-start.php#refsect1-function.session-start-returnvalues) and add an error and/or set the status based on that? You would have access to the return value in post hook.

HeenaBansal20: @brettmc , Thank you. That's a great suggestion and I implemented the suggestion in my latest commit. 
Please take a look.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
```diff
@@ -0,0 +1,164 @@
...
+            pre: static function ($object, ?array $params, ?string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation, $name) {
+                /** @psalm-suppress ArgumentTypeCoercion */
+                $builder = self::makeBuilder($instrumentation, $name, $function);
+                self::addParams($builder, $function, $params);
+                $parent = Context::getCurrent();
+                $span = $builder->startSpan();
+                Context::storage()->attach($span->storeInContext($parent));
+            },
+            post: static function ($object, ?array $params, mixed $return, ?Throwable $exception) use ($instrumentation, $name, $function) {
+                $builder = self::makeBuilder($instrumentation, $name, $function);
```
brettmc: I'm a bit confused about this. You create a builder (as if a new span was going to be started), but then don't do anything with it. Normally I'd expect a `pre` hook to start a span, and the `post` hook to end the active span, but not create any new spans...

src/Instrumentation/Session/tests/Integration/PhpSessionInstrumentationTest.php
```diff
@@ -0,0 +1,244 @@
...
+            'read_and_close' => true,
+            'cookie_lifetime' => 3600,
+        ];
+    
+        $return = session_start($options);
+        
+        // Verify the span was created
+        $this->assertCount(1, $this->storage);
+        $span = $this->storage[0];
+        echo session_status();
```
brettmc:
```suggestion
```

src/Instrumentation/Session/tests/Integration/PhpSessionInstrumentationTest.php
```diff
@@ -0,0 +1,244 @@
...
+        $this->assertNotNull($attributes->get('php.session.name'));
+        
+        // Check status
+        $this->assertEquals(StatusCode::STATUS_OK, $span->getStatus()->getCode());
+    }
+    
+    /**
+     * @runInSeparateProcess
+     */
+    public function test_session_start_exception_simulation(): void
```
brettmc: the documentation for session_start doesn't mention that it can throw an exception.

HeenaBansal20: Actually yes , I checked all instrumented session function and none of them throws exception. 
I just added the return status now and set the span status according to return status.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: yeah , I overlooked it , I dont want to create a new span in post hook. 
I corrected the post hook.

HeenaBansal20:
@brettmc , Thanks for reviewing it. 
DO you mind doing final check on this PR. I appreciate your time and efforts on reviewing this PR patiently . 
Thank you.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
```diff
@@ -0,0 +1,159 @@
...
+                $sessionStartSuccess = $return === true;
+                $span->setAttribute('php.session.status', $sessionStartSuccess ? 'active' : 'inactive');
+
+                // Add session cookie parameters
+                $cookieParams = session_get_cookie_params();
+                $index = 0;
+                ksort($cookieParams);
+                foreach ($cookieParams as $key => $value) {
+                    if (is_scalar($value)) {
+                        $span->setAttribute("php.session.cookie.key[$index]", $key);
```
brettmc: could this be an array of keys, eg `php.srssion.cookie.keys = [key1, key2, key3]` ?

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
```diff
@@ -0,0 +1,159 @@
...
+            case 'session_destroy':
+                break;
+        }
+        
+        // Set span status based on return value
+        $Success = $return === true;
+        if (!$Success) {
+            $span->setStatus(StatusCode::STATUS_ERROR, "$function failed with return code $return");
+        } else {
+            $span->setStatus(StatusCode::STATUS_OK);
```
brettmc: https://opentelemetry.io/docs/specs/otel/trace/api/#set-status says:

> Generally, Instrumentation Libraries SHOULD NOT set the status code to Ok, unless explicitly configured to do so. Instrumentation Libraries SHOULD leave the status code as Unset unless there is an error, as described above.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: Sure. Since we are just storing keys we can have array attribute holding all keys.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
HeenaBansal20: I see ..
Thanks @brettmc for sharing this info. I was not aware of this one.
So,  in this case I think we can keep the status to be default i.e STATUS::UNSET  unless $return value is !success which sets the status to STATUS_ERROR.

src/Instrumentation/Session/src/PhpSessionInstrumentation.php
```diff
@@ -0,0 +1,156 @@
...
+                // Use the return value to determine if session was successfully started
+                $sessionStartSuccess = $return === true;
+                $span->setAttribute('php.session.status', $sessionStartSuccess ? 'active' : 'inactive');
+
+                // Add session cookie parameters
+                $cookieParams = session_get_cookie_params();
+                $cookieKeys = [];
+                ksort($cookieParams);
+                foreach ($cookieParams as $key => $value) {
+                    if (is_scalar($value)) {
```
brettmc: if cookie keys can be mixed types, I think you should cast them all to string since arrays must be homogeneous, per spec: https://opentelemetry.io/docs/specs/otel/common/#attribute

HeenaBansal20: @brettmc , I dont think  session_get_cookie_params will return keys other than string type. it will only return  below predefined set of array.
https://www.php.net/manual/en/function.session-get-cookie-params.php#refsect1-function.session-get-cookie-params-returnvalues.
the keys will be always string type.

brettmc: why this check, if you're storing the keys? what's special about non-scalar values that makes you not want to store the associated key?

HeenaBansal20: Agree ... my bad , this can be removed since now w e just storing the keys , no check is required on value.

HeenaBansal20:
@brettmc , I would really appreciate if you can spend some time to review.
Thanks.

brettmc (approved):

PR #437, opened 2025-09-14
Add auto-instrumentation for PHP output buffering control functions 

HeenaBansal20:
This PR adds auto instrumentation for pHP output buffering control functions 

```
ob_start()
ob_flush()
ob_clean()
flush()
```


brettmc (approved):

brettmc:
> Should I remove curl_exec and curl_init from IO package

Yes, I agree with you. Can you submit it as a separate PR?

HeenaBansal20:
> > Should I remove curl_exec and curl_init from IO package
> 
> Yes, I agree with you. Can you submit it as a separate PR?

Sure I'll take it in a separate PR. 
Thanks

PR #442, opened 2025-09-17
feat: sqlcommenter for PDO, MySqli, PostgreSql (pdo_mysql, pdo_pgsql)

jerrytfleung:
It is a followup / replacement of https://github.com/open-telemetry/opentelemetry-php-contrib/pull/406

## Description:
Currently, there isn't any sqlcommenter for the database instrumentation library that allows customers to correlate their span with the database query. This PR adds the implementation for sqlcommenter according to https://github.com/open-telemetry/semantic-conventions/pull/2495

 - Decoupled sqlcommenter logic to a separate package, proposed `open-telemetry/opentelemetry-sqlcommenter`
 - Updated PDO (pdo_mysql, pdo_pgsql) & MySqli & PostgreSql instrumentation to add sqlcommenter feature

### Unit Testing:
- Wrote unit tests for sql comment for correctness and they're all passing.

### Manual Testing:
Once requiring `open-telemetry/opentelemetry-sqlcommenter`, the sqlcommenter feature is enabled. 

The following extracted the span attributes.
PDO
```
"attributes": {
            "code.function.name": "PDO::query",
            "db.query.text": "SELECT code, name, region from country LIMIT 1\/*traceparent='00-03d0617d3d102d4db6d345e1e6a10714-660de8b2725490e0-01'*\/;",
            "server.address": "44.203.164.2",
            "db.namespace": "world-db",
            "db.system.name": "postgresql"
        }
```

MySqli
```
"attributes": {
            "code.function.name": "mysqli::query",
            "db.query.text": "SELECT Code, Name, Region from country LIMIT 1\/*traceparent='00-79c1c1c373c34772ac73ff8575c3ee4f-0656f3d396b84beb-01'*\/;",
            "db.operation.name": "SELECT",
            "db.system.name": "mysql",
            "server.address": "34.239.101.164",
            "server.port": "3306",
            "db.namespace": "world"
        }
```

PostgreSql
```
"attributes": {
            "code.function.name": "PDO::query",
            "db.query.text": "SELECT code, name, region from country LIMIT 1\/*traceparent='00-0a7a38884b2dfc3b514376fc7bf4033a-81c6031fe6a0e18b-01'*\/;",
            "server.address": "44.203.164.2",
            "db.namespace": "world-db",
            "db.system.name": "postgresql"
        }
```


src/SqlCommenter/src/ContextPropagatorFactory.php
```diff
@@ -0,0 +1,73 @@
...
+use OpenTelemetry\Context\Propagation\MultiTextMapPropagator;
+use OpenTelemetry\Context\Propagation\NoopTextMapPropagator;
+use OpenTelemetry\Context\Propagation\TextMapPropagatorInterface;
+use OpenTelemetry\SDK\Registry;
+
+class ContextPropagatorFactory
+{
+    use LogsMessagesTrait;
+
+    public function create(): ?TextMapPropagatorInterface
```
brettmc, line 17: I'm not understanding the point of this (actually, the entire `src/ContextPropagator`)... isn't that what https://github.com/open-telemetry/opentelemetry-php/pull/1712 does?

jerrytfleung: They are not the same.

The response propagator in https://github.com/open-telemetry/opentelemetry-php/pull/1712 supports propagating context in the http response. 
(e.g. Propagate server-timing and traceresponse in the http response of slim instrumentation). 
```
< HTTP/1.1 200 OK
< Host: localhost:8080
< Date: Thu, 11 Sep 2025 23:28:27 GMT
< Connection: close
< X-Powered-By: PHP/8.3.6
< server-timing: traceparent;desc=00-73dbf0f78c24b08faf273fee7ae5d77b-76c077b8629b08d1-01
< traceresponse: 00-73dbf0f78c24b08faf273fee7ae5d77b-76c077b8629b08d1-01
< Content-type: text/html; charset=UTF-8
```

The context propagator in this PR works quite similar to the propagator `OTEL_PROPAGATORS` in the SDK. However, they are not the same. 
`OTEL_PROPAGATORS=baggage,tracecontext` propagates baggage and tracecontext from one service to another service.
The context propagator propagates context from one service to the database (via sqlcommenter & query statement). The [spec](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#sql-commenter) mentioned the instrumentation should allow user to pass a propagator to overwrite the global propagator. This context propagator maintains a list of propagator to serve that purpose. 

e.g. 
`OTEL_EXPERIMENTAL_RESPONSE_PROPAGATORS=servertiming,traceresponse OTEL_PROPAGATORS=b3,baggage,tracecontext OTEL_PHP_INSTRUMENTATION_CONTEXT_PROPAGATORS=tracecontext` means 
1. Propagating `server-timing` & `traceresponse` in http instrumentation library (e.g. slim...)
2. Propagating `b3`, `baggage` & `tracecontext` across service
3. Propagating `tracecontext` to database

`OTEL_EXPERIMENTAL_RESPONSE_PROPAGATORS=servertiming,traceresponse OTEL_PROPAGATORS=b3,baggage,tracecontext` means
1. Propagating `server-timing` & `traceresponse` in http instrumentation library (e.g. slim...)
2. Propagating `b3`, `baggage` & `tracecontext` across service and database 

Let me know if it is making sense to you.

cheempz: In the previous PR #406 `ContextPropagator` was under the PDO Instrumentation and thus naming made some sense. If we move it out like proposed in this PR, it should be renamed to something like `SqlCommenterContext...`. Also, any reason this class can't just be part of the SqlCommenter i.e under the proposed `src/SqlCommenter` directory?

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
```diff
@@ -440,11 +448,52 @@ private static function constructPostHook(int $paramsOffset, CachedInstrumentati
...
+            $query = self::UNDEFINED;
+        }
+        $span->setAttributes([
+            TraceAttributes::DB_QUERY_TEXT => $query,
+            TraceAttributes::DB_OPERATION_NAME => self::extractQueryCommand($query),
+        ]);
+
         self::addTransactionLink($tracker, $span, $mysqli);
+
+        if (class_exists('OpenTelemetry\Contrib\SqlCommenter\SqlCommenter') && $query !== self::UNDEFINED) {
```
cheempz, line 461: In the previous PR #406 the SQL Commenter feature was enabled via an explicit configuration. This PR proposes instead the customers opt-in by installing the `sqlcommenter` package, correct? Can the READMEs for each database instrumentation be updated to provide usage / configuration info?

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
```diff
@@ -440,11 +448,52 @@ private static function constructPostHook(int $paramsOffset, CachedInstrumentati
...
+            $comments = [];
+            if ($contextPropagator !== null) {
+                $contextPropagator->inject($comments);
+            } else {
+                Globals::propagator()->inject($comments);
+            }
+            if (class_exists('OpenTelemetry\Contrib\SqlCommenter\SqlCommenter')) {
+                $query = \OpenTelemetry\Contrib\SqlCommenter\SqlCommenter::inject($query, $comments);
+            }
+            if (class_exists('OpenTelemetry\Contrib\ContextPropagator\ContextPropagation') && \OpenTelemetry\Contrib\ContextPropagator\ContextPropagation::isAttributeEnabled()) {
```
cheempz: nit and personal opinion, this should be a SqlCommenter (not ContextPropagation) configuration option, whether the db query text attribute reflects value pre-or-post comment injection.

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
```diff
@@ -440,11 +448,52 @@ private static function constructPostHook(int $paramsOffset, CachedInstrumentati
...
+                $span->setAttributes([
+                    TraceAttributes::DB_QUERY_TEXT => (string) $query,
+                    TraceAttributes::DB_OPERATION_NAME => self::extractQueryCommand($query),
```
cheempz: Why does the `DB_OPERATION_NAME` need to be set again?

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
```diff
@@ -466,6 +512,14 @@ private static function queryPostHook(CachedInstrumentation $instrumentation, My
 
     }
 
+    /** @param non-empty-string $spanName */
+    private static function multiQueryPreHook(string $spanName, CachedInstrumentation $instrumentation, MySqliTracker $tracker, $obj, array $params, ?string $class, string $function, ?string $filename, ?int $lineno): void
+    {
+        $span = self::startSpan($spanName, $instrumentation, $class, $function, $filename, $lineno, []);
+        $mysqli = $obj ? $obj : $params[0];
+        self::addTransactionLink($tracker, $span, $mysqli);
+    }
```
cheempz, line 514: Just curious why this prehook was added?

src/SqlCommenter/src/ContextPropagatorFactory.php
jerrytfleung: Beside sqlcommenter, there are other ways to propagate the context to database. e.g. https://opentelemetry.io/docs/specs/semconv/database/sql-server/#set-context_info & https://github.com/open-telemetry/semantic-conventions/pull/2610 

So I think it would be better to decouple the context propagator logic.

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
jerrytfleung: Yes, the opt-in is by installing the sqlcommenter & context-propagator. 
Updated the README.md per each database instrumentation library.

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
jerrytfleung: There could be other ways (e.g. SET_CONTEXT_INFO / [V$SESSION.ACTION](https://docs.oracle.com/en/database/oracle/oracle-database/23/refrn/V-SESSION.html)) to inject comments into query but we have one way to manage context propagator. 

I think it would be better to do that in context propagator package.

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
jerrytfleung: DB_QUERY_TEXT and DB_OPERATION_NAME are set in pair. It is to make sure the DB_OPERATION_NAME is extracted from the updated DB_QUERY_TEXT.

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
jerrytfleung: It is for `multi_query` scenario where it works so differently when compared to `query` and `real_query`. 
In order to reflect the db query text attribute for pre-or-post comment injection, I need to move the assignment of DB_QUERY_TEXT from post hook to pre hook. `query` and `real_query` works in this way. 

For `multi_query`, the DB_QUERY_TEXT is set at the post hook and also the `next_result` post hook. There is no easy way to maintain the same `QueryPrehook` for `multi_query`. So I added a pre hook.

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
cheempz: According to the [spec](https://opentelemetry.io/docs/specs/semconv/database/database-spans/), `db.operation.name` is supposed to be the low cardinality operation such as `SELECT` extracted from query text. Does the `extractQueryCommand` actually include comments!? if so it really shouldn't. On assumption it doesn't, why do this possibly expensive work twice?

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
cheempz: Still not seeing what this added code is doing for multi query, can you explain.

src/SqlCommenter/src/ContextPropagatorFactory.php
cheempz: [SET_CONTEXT_INFO](https://opentelemetry.io/docs/specs/semconv/database/sql-server/#set-context_info) and [V$SESSION.ACTION](https://github.com/open-telemetry/semantic-conventions/pull/2610) work pretty differently from SQL Commenting and are prescriptive in the value type/size that can be injected, meaning i don't think they'd take a composite list of propagators. The example implementations i can find so far do not introduce a generic "database context propagator" concept:

* https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2709
* https://github.com/signalfx/splunk-otel-java/pull/2405

Or maybe i misunderstand your intention with this `ContextPropagator` package--how do you envision using it for SET_CONTEXT_INFO / V$SESSION.ACTION propagation?

src/Instrumentation/PDO/README.md
```diff
@@ -33,4 +33,12 @@ otel.instrumentation.pdo.distribute_statement_to_linked_spans = true
...
 ```shell
 OTEL_PHP_INSTRUMENTATION_PDO_DISTRIBUTE_STATEMENT_TO_LINKED_SPANS=true
+```
+
+## Database Context Propagation
+
+Enable context propagation for database queries by installing the following packages:
+```shell
+composer require open-telemetry/opentelemetry-context-propagator
+composer require open-telemetry/opentelemetry-sqlcommenter
```
cheempz, line 42: Right now it would be enabled for both `postgresql` or `mysql` under PDO, can we add a config to enable only specific driver?

src/SqlCommenter/src/ContextPropagatorFactory.php
jerrytfleung: I envision ContextPropagator acts as the source of the database context and put the information into a list, and SqlCommenter / SET_CONTEXT_INFO / V$SESSION.ACTION propagation can retrieve the required information from the list and propagate.

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
jerrytfleung: The reason is the `$query` is updated, `DB_OPERATION_NAME` should sync with the updated `$query` again.
It could be argued that `$query` is just updated by adding a sql comment and `DB_OPERATION_NAME` shouldn't have any impact. I think it is better to extract query command from the updated `$query` again. 
Also, I don't think it is expensive to call `extractQueryCommand` again.

@open-telemetry/php-maintainers Can we get any insight from you?

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
jerrytfleung: Added comments to the function to explain. 

In short, multi_query shared the same pre-hook function `QueryPreHook` as query and real_query. I need to create a new pre hook to keep multi_query works the same logic as before.

src/Instrumentation/MySqli/src/MySqliInstrumentation.php
brettmc: I don't think the operation should be re-extracted, since adding comments should not be mutating the actual query in any way.

jerrytfleung: Done removing the query command re-extraction logic.

src/SqlCommenter/src/ContextPropagatorFactory.php
brettmc: > ... mentioned the instrumentation should allow user to pass a propagator to overwrite the global propagator

So can we describe this component as a "per-service context propagator override", and give it a more meaningful name? (It's too similar to the core context propagation concept, IMO, and will cause confusion).
The reason I call out "per-service" (or per system/component/something), is to allow for a scenario where an application connects to multiple databases, and wants to propagate context to each of them in a different way. That doesn't seem unreasonable.

jerrytfleung: Yes, "per-service / per-system / per-component / something context propagator override" seems reasonable to me. 

To clarify more and make sure we are on the same page. 
Says we have an application using PDO to connect multiple databases
- MySqli (`pdo_mysql`)
- PostgreSQL (`pdo_pgsql`)
- Oracle Database (`pdo_oci`)
- SQL Server (`pdo_sqlsrv`)
- IBM DB2 (`pdo_ibm`)
- Firebird (`pdo_firebird`)

#### Ways of context propagation
1. As all above database supports C-style comments, the context could be propagated via [sqlcommenter](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#sql-commenter)
e.g.
```
SELECT * FROM songs /*traceparent='00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01',tracestate='congo%3Dt61rcWkgMzE%2Crojo%3D00f067aa0ba902b7'*/
```
2. SQL Server could use [SET CONTEXT_INFO](https://opentelemetry.io/docs/specs/semconv/database/sql-server/#set-context_info) way to propagate traceparent (55 bytes) only (128 bytes length limitation)
e.g.
```
DECLARE @traceparent varbinary(55);
SET @traceparent = CAST('00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01' AS varbinary(55));
SET CONTEXT_INFO @traceparent; 
``` 
Then run the query
4. Oracle database opts for [V$SESSION.ACTION](https://github.com/open-telemetry/semantic-conventions/pull/2610) way to propagate traceparent
e.g.
```
BEGIN
    DBMS_APPLICATION_INFO.SET_ACTION('00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01');
END;
``` 
Then run the query

## Questions
1. Does it mean the "context propagator overrides" in this case should be `per-component`?
e.g. I tried to mock up the names for illustrations. 
`OTEL_DATABASE_CONTEXT_PROPAGATORS_FOR_SQLCOMMENTER=tracecontext`
`OTEL_DATABASE_CONTEXT_PROPAGATORS_FOR_SET_CONTEXT_INFO=tracecontext`
`OTEL_DATABASE_CONTEXT_PROPAGATORS_FOR_SET_ACTION=tracecontext`
2. Should Oracle database (pdo_oci) opt for `V$SESSION.ACTION`(highest priority) > `SET CONTEXT_INFO` > `sqlcommenter` (least priority )?
3. Should MySqli (pdo_mysql) opt for `sqlcommenter` only as it doesn't support other propagation ways?

jerrytfleung: @brettmc I moved the propagator logic into SqlCommenter namespace so as to describe that as a "per sqlcommenter (component) context propagator override". Please take a look to see if I get your point correctly. Thanks!

src/Instrumentation/PDO/README.md
jerrytfleung: @brettmc / @open-telemetry/php-maintainers  Do you think it is a good idea to create a config to enable only specific driver?
e.g.
`OTEL_PHP_SQLCOMMENTER_OPT_IN_DATABASE` where default value = mysql and valid values are { mysql, postgresql }

jerrytfleung: For more information, the hardcoding logic https://github.com/jerrytfleung/opentelemetry-php-contrib/blob/service_name_propagator/src/Instrumentation/PDO/src/PDOInstrumentation.php#L135-L137 was approved in #406

brettmc: I'd be happy to wait for somebody to ask for it, this is a `0.1` release

brettmc:
plz add a .gitsplit.yaml entry for the new sql commenter package

brettmc (approved):

jerrytfleung:
> plz add a .gitsplit.yaml entry for the new sql commenter package

@brettmc 
```
  - prefix: "src/SqlCommenter"
    target: "https://${GH_TOKEN}@github.com/opentelemetry-php/contrib-sqlcommenter.git"
```
was added in `.gitsplit.yml` in this PR. 
Is there any other entry I need to aware?

PR #444, opened 2025-10-02
adding apache, fpm, k8s resource detectors

brettmc:
this ports the detectors from https://github.com/open-telemetry/opentelemetry-php/pull/1628 into contrib

NB that I haven't yet ported across the sdk-config parts yet, which can be a follow-up.

Replaces: https://github.com/open-telemetry/opentelemetry-php/pull/1628


src/ResourceDetectors/Kubernetes/src/Kubernetes.php
```diff
@@ -0,0 +1,538 @@
...
+     */
+    private function getStableInstanceId(string $podUid): string
+    {
+        // Pod UID is already a UUID, but we'll use our standard UUID v5 pattern for consistency
+        $components = [
+            'k8s',
+            $podUid,
+            $this->getPodName() ?? 'unknown-pod',
+            $this->getNamespace() ?? 'default',
+        ];
```
Nevay, line 68: While the following spec note is about the Collector, it seems to apply here too - the code currently generates the same instance id for all containers in a pod:

https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/#service-instance-id
> It’s not recommended for a Collector to set `service.instance.id` if it can’t unambiguously determine the service instance that is generating that telemetry. For instance, creating an UUID based on `pod.name` will likely be wrong, as the Collector might not know from which container within that pod the telemetry originated.

src/ResourceDetectors/Fpm/src/Fpm.php
```diff
@@ -0,0 +1,93 @@
...
+    private function getStableInstanceId(): string
+    {
+        $components = [
+            'fpm',
+            $this->getFpmPoolName() ?? 'default',
+            gethostname() ?: 'localhost',
+        ];
+
+        // Create a stable UUID v5 using a namespace UUID and deterministic name
+        $namespace = Uuid::fromString('4d63009a-8d0f-11ee-aad7-4c796ed8e320'); // DNS namespace UUID
```
Nevay, line 63:
```suggestion
        $namespace = Uuid::fromString('4d63009a-8d0f-11ee-aad7-4c796ed8e320');
```
Same comment needs to be removed in the kubernetes detector.

src/ResourceDetectors/Apache/src/Apache.php
```diff
@@ -0,0 +1,129 @@
...
+use OpenTelemetry\SemConv\Incubating\Attributes\ServiceIncubatingAttributes;
+use OpenTelemetry\SemConv\Version;
+use function php_sapi_name;
+use Ramsey\Uuid\Uuid;
+
+/**
+ * Apache resource detector that provides stable service instance IDs to avoid high cardinality issues.
+ *
+ * For Apache mod_php environments, generates a stable instance ID based on the server name and hostname
+ * rather than using random UUIDs which cause cardinality explosion in metrics.
```
Nevay, line 23: This seems like it will violate the single writer principle when not setting `process.pid` / when using threads instead of processes?

https://opentelemetry.io/docs/specs/otel/metrics/data-model/#single-writer
> Aggregations of metric streams MUST only be written from a single logical source at any given point time.

> For example, if an SDK implementation fails to find uniquely identifying Resource attributes for a component, then all instances of that component could be reporting metrics as if they are from the same resource. In this case, metrics will be reported at inconsistent time intervals. For metrics like cumulative sums, this could cause issues where pairs of points appear to reset the cumulative sum leading to unusable metrics.

PR #468, opened 2025-11-27
Added pg_query_params to  PostgreSQL instrumentation

andrewnicols:
The README already documents that this function is instrumented, but it is not.

I have not instrumented the query parameters as this is not done for `pg_send_query_params`.
I've implemented this locally and it works with the SQL commenter.

### DIFF
--- a/src/Instrumentation/PostgreSql/tests/Integration/PostgreSqlInstrumentationTest.php
+++ b/src/Instrumentation/PostgreSql/tests/Integration/PostgreSqlInstrumentationTest.php
@@ -124,6 +124,30 @@ class PostgreSqlInstrumentationTest extends TestCase
         $this->assertDatabaseAttributesForAllSpans($offset);
     }
 
+    public function test_pg_query_params(): void
+    {
+        $conn = pg_connect('host=' . $this->pgsqlHost . ' dbname=' . $this->database . ' user=' . $this->user . ' password=' . $this->passwd);
+        $this->assertTrue($conn instanceof Connection);
+
+        $offset = 0;
+        $this->assertSame('pg_connect', $this->storage->offsetGet($offset)->getName());
+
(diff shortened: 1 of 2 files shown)


src/Instrumentation/PostgreSql/tests/Integration/PostgreSqlInstrumentationTest.php
```diff
@@ -124,6 +124,34 @@ public function test_pg_query(): void
...
+
+        $this->assertSame('pg_query_params', $this->storage->offsetGet($offset)->getName());
+        $this->assertAttributes($offset, [
+            TraceAttributes::DB_QUERY_TEXT => 'SELECT * FROM users WHERE id = $1',
+            TraceAttributes::DB_OPERATION_NAME => 'SELECT',
+        ]);
+
+        while ($row = pg_fetch_assoc($res)) {
+        }
+        $offset++;
```
ChrisLightfootWild: Should this increment happen within the loop?

andrewnicols: This test is largely copied from the test for `pg_query`, however `pg_fetch_assoc` is not observed, so I wouldn't expect there to be any increase in the number of items in storage.

It looks like the offset increase is purely to make the 0-indexed storage add up for the assertCount.

I've changed the offset to be added closer to the call that it relates to, and I've modified the `assertCount` to check `$offset + 1`.

ChrisLightfootWild:
Hey @andrewnicols - sorry this one went a little stale. I'm happy to review this one if you are still interested in getting this merged?

andrewnicols:
Yes - that woudl be good thanks. Do you need anything from me? I think the patch is still up-to-date and complete.

PR #491, opened 2025-12-17
Enable context-flattening to support context field searching

### LINKED ISSUE #1837: [Laravel] Add option to flatten log context attributes for better searchability
This feature is specific to the Laravel auto-instrumentation package and how it serializes PSR-3 log context for OTLP export. It does not require specification changes.

  **Is your feature request related to a problem?**

  The Laravel LogWatcher JSON-encodes the entire log context into a single context attribute:

  context: {"http":{"method":"GET","path":"/users","status":200},"user_id":"123"}

  This makes it difficult or impossible to search/filter logs by specific context values in observability backends like SigNoz, Jaeger, or Grafana. Users cannot query for http.method = "GET" or user_id = "123" because these values are buried inside an opaque JSON string rather than being individual, indexed attributes.

  **Describe the solution you'd like**

  A configuration option OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN that, when enabled, spreads log context as individual OTLP attributes with dot-notation for nested arrays:

  http.method: GET
  http.path: /users
  http.status: 200
  user_id: 123

  Default behavior (off) would remain unchanged for backward compatibility. An argument could be made for making this behavior on by default in future versions, but for now simply having the option could be very useful for users.

  **Describe alternatives you've considered**

  1. Using PSR-3 instrumentation's MODE_EXPORT: This already spreads context as individual attributes, but it causes duplicate log records when used alongside Laravel's LogWatcher since both emit logs independently with no coordination between them.
  2. Post-processing with a LogRecordProcessor: Not viable because attributes are immutable by the time processors receive the record.
  3. Parsing the JSON in the observability backend: Possible in some backends but defeats the purpose of structured logging and varies by vendor.

  **Additional context**

  I have a working implementation ready as a PR that:
  - Uses ConfigurationResolver consistent with other OTEL PHP configuration patterns
  - Recursively flattens nested arrays with dot notation
  - Handles Stringable objects, scalars, and edge cases appropriately
  - Preserves existing exception attribute handling
  - Includes comprehensive unit tests
  - Is fully backward compatible (off by default)

nickmarden:
Description

  Resolves open-telemetry/opentelemetry-php#1837

  This PR adds an optional configuration to flatten log context attributes into individual OTLP attributes, improving searchability in observability backends.

  Problem

  The Laravel LogWatcher currently JSON-encodes the entire log context into a single context attribute:

  context: {"http":{"method":"GET","path":"/users","status":200},"user_id":"123"}

  This makes it impossible to search/filter by specific values (e.g., http.method = "GET") in backends like SigNoz, Jaeger, or Grafana.

  Solution

  A new configuration option OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN that, when enabled, spreads context as individual attributes with dot-notation for nested arrays:

  http.method: GET
  http.path: /users
  http.status: 200
  user_id: 123

  Changes

  - src/Watchers/LogWatcher.php: Added flattening logic with ConfigurationResolver, dot-notation for nested arrays, and proper handling of Stringable objects
  - README.md: Documented the new configuration option with examples
  - tests/Unit/Watchers/LogWatcherTest.php: Added 11 unit tests covering default behavior, flattened mode, nested arrays, exceptions, and edge cases

  Backward Compatibility

  - Default behavior unchanged (off by default)
  - Existing tests continue to pass

  Testing

  ./vendor/bin/phpunit --testsuite=unit      # 18 tests, 42 assertions
  ./vendor/bin/phpunit --testsuite=integration  # 21 tests, 120 assertions
  
  Screenshot
  
  
Screenshot 2025-12-17 at 6 56 35 PM

### DIFF
--- a/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
+++ b/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
@@ -8,16 +8,28 @@ use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Log\Events\MessageLogged;
 use Illuminate\Log\LogManager;
 use OpenTelemetry\API\Instrumentation\CachedInstrumentation;
+use OpenTelemetry\API\Instrumentation\ConfigurationResolver;
 use OpenTelemetry\API\Logs\Severity;
+use Stringable;
 use Throwable;
 use TypeError;
 
 class LogWatcher extends Watcher
 {
+    /**
+     * When enabled, log context attributes are spread as individual OTLP attributes
+     * instead of being JSON-encoded into a single 'context' attribute.
+     * This improves searchability in observability backends like SigNoz.
+     */
+    public const OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN = 'OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN';
+
     private LogManager $logger;
+    private bool $flattenAttributes;
+
     public function __construct(
         private CachedInstrumentation $instrumentation,
     ) {
+        $this->flattenAttributes = $this->shouldFlattenAttributes();
     }
 
     /** @psalm-suppress UndefinedInterfaceMethod */
@@ -56,20 +68,28 @@ class LogWatcher extends Watcher
             ->logger()
             ->logRecordBuilder();
 
-        $context = array_filter($log->context, static fn ($value) => $value !== null);
+        $contextToProcess = array_filter($log->context, static fn ($value) => $value !== null);
         $exception = $this->getExceptionFromContext($log->context);
 
         if ($exception !== null) {
             $logBuilder->setException($exception);
 
-            unset($context['exception']);
+            unset($contextToProcess['exception']);
         }
 
         $logBuilder->setBody($log->message)
             ->setSeverityText($log->level)
-            ->setSeverityNumber(Severity::fromPsr3($log->level))
-            ->setAttribute('context', $context)
-            ->emit();
+            ->setSeverityNumber(Severity::fromPsr3($log->level));
+
+        if ($this->flattenAttributes) {
+            foreach ($this->buildFlattenedAttributes($contextToProcess) as $key => $value) {
+                $logBuilder->setAttribute($key, $value);
+            }
+        } else {
+            $logBuilder->setAttribute('context', json_encode($contextToProcess));
+        }
+
+        $logBuilder->emit();
     }
 
     private function getExceptionFromContext(array $context): ?Throwable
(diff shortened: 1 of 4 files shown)


brettmc (approved):

src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
```diff
@@ -56,20 +68,28 @@ public function recordLog(MessageLogged $log): void
...
 
         $logBuilder->setBody($log->message)
             ->setSeverityText($log->level)
-            ->setSeverityNumber(Severity::fromPsr3($log->level))
-            ->setAttribute('context', $context)
-            ->emit();
+            ->setSeverityNumber(Severity::fromPsr3($log->level));
+
+        if ($this->flattenAttributes) {
+            foreach ($this->buildFlattenedAttributes($contextToProcess) as $key => $value) {
```
ChrisLightfootWild, line 86: Could https://laravel.com/docs/13.x/helpers#method-array-dot work here instead? 🤔

arthurmartelli: The implementation is much cleaner with that array method :)

ChrisLightfootWild (approved):

ChrisLightfootWild:
Hey @nickmarden / @arthurmartelli - firstly, thanks for your contribution here.

It appears that I made an error in merging this though, as it has conflicted with behaviour that changed in a later PR #511 (which had already been merged ahead of this one).

This may have to be reverted to restore the previous behaviour - however, I think it is something that perhaps this PR was trying to achieve in the first instance.

Is there any additional behaviour that this PR provides in addition to the changes introduced in #511 which you were still hoping to accomplish?

nickmarden:
@ChrisLightfootWild lemme take a look to see if #511 accomplishes what we need.

nickmarden:
Thanks @ChrisLightfootWild  for bringing this to our attention.

After reviewing #511, I think it solves the JSON-encoding problem by passing context as a structured array, which is better, but it doesn't solve the faceted search use case in systems like SigNoz. SigNoz (and similar backends) require individual top-level OTLP attributes to support filtering/faceting. A single context attribute containing a nested array/kvlist is not decomposed into searchable facets.

The value proposition of #491 is the opt-in `OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN` mode, which spreads context fields as individual dot-notation OTLP attributes (e.g., `http.method` = GET, `user_id` = 123). That's what enables faceted log search.

The issue with the current merged state is that the non-flatten path in #491 regressed #511's improvement by reverting to json_encode() instead of keeping the structured array. A fix could be:

- Non-flatten default: keep #511's structured array (`setAttribute('context', $context)`)
- Flatten mode: individual dot-notation attributes as #491 implemented

If you'd like, @arthurmartelli or I can open a follow-up PR that corrects the non-flatten regression while preserving the (optional) flatten feature.

nickmarden:
@ChrisLightfootWild please have a look at #460 - it keeps the desired context-flattening while preserving the #511 structured array behavior as default.

PR #501, opened 2026-01-22
Add selective instrumentation configuration

sinsi404:
I've added a feature to selectively enable or disable Laravel auto-instrumentations via environment variables. This is required for me to disable db instrumentation as it has some weird bug. It will also help improve performance when the application doesn't require specific traces.

This selective instrumentation approach is inspired by [auto-instrumentations-node](https://www.npmjs.com/package/@opentelemetry/auto-instrumentations-node)

This is my first PR to this community, so please let me know if there's anything I should fix or improve.

### DIFF
--- a/src/Instrumentation/Laravel/src/InstrumentationConfig.php
+++ b/src/Instrumentation/Laravel/src/InstrumentationConfig.php
@@ -1,0 +1,126 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Contrib\Instrumentation\Laravel;
+
+use OpenTelemetry\SDK\Common\Configuration\Configuration;
+
+final class InstrumentationConfig
+{
+    public const HTTP = 'http';
+    public const CONSOLE = 'console';
+    public const QUEUE = 'queue';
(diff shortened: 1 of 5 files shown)


src/Instrumentation/Laravel/src/InstrumentationConfig.php
```diff
@@ -0,0 +1,126 @@
...
+        self::REDIS,
+    ];
+
+    private const ALIASES = [
+        'all' => self::ALL_INSTRUMENTATIONS,
+        'watchers' => self::GROUP_WATCHERS,
+    ];
+
+    private const ENV_ENABLED = 'OTEL_LARAVEL_ENABLED_INSTRUMENTATIONS';
+    private const ENV_DISABLED = 'OTEL_LARAVEL_DISABLED_INSTRUMENTATIONS';
```
brettmc, line 52: spec [recommends](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#language-specific-environment-variables) that language-specific env vars should begin with `OTEL_PHP`

PR #524, opened 2026-03-17
fix(AwsSdk): handle missing x-amz-request-id in metadata

adiafora:
Fixes https://github.com/open-telemetry/opentelemetry-php/issues/1897

### DIFF
--- a/src/Instrumentation/AwsSdk/src/AwsSdkInstrumentation.php
+++ b/src/Instrumentation/AwsSdk/src/AwsSdkInstrumentation.php
@@ -76,7 +76,12 @@ final class AwsSdkInstrumentation
                 if ($result instanceof ResultInterface && isset($result['@metadata'])) {
                     $span->setAttribute(TraceAttributes::HTTP_RESPONSE_STATUS_CODE, $result['@metadata']['statusCode']); // @phan-suppress-current-line PhanTypeMismatchDimFetch
 
-                    $awsRequestId = $result['@metadata']['headers']['x-amz-request-id'] ?? null;
+                    $headers = $result['@metadata']['headers'] ?? []; // @phan-suppress-current-line PhanTypeMismatchDimFetch
+
+                    $awsRequestId = $headers['x-amz-request-id']
+                        ?? $headers['x-amzn-requestid']
+                        ?? $headers['x-amzn-request-id']
+                        ?? null;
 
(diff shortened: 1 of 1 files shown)


src/Instrumentation/AwsSdk/src/AwsSdkInstrumentation.php
```diff
@@ -75,7 +75,12 @@ public static function register(): void
 
                 if ($result instanceof ResultInterface && isset($result['@metadata'])) {
                     $span->setAttribute(TraceAttributes::HTTP_RESPONSE_STATUS_CODE, $result['@metadata']['statusCode']); // @phan-suppress-current-line PhanTypeMismatchDimFetch
-                    $span->setAttribute(TraceAttributes::AWS_REQUEST_ID, $result['@metadata']['headers']['x-amz-request-id']); // @phan-suppress-current-line PhanTypeMismatchDimFetch
+
+                    $awsRequestId = $result['@metadata']['headers']['x-amz-request-id'] ?? null;
```
scprek: This will fix the bug. Should we address the other options later? I was not 100% sure but based on the AWS docs this information can have several headers 

https://opentelemetry.io/docs/specs/semconv/cloud-providers/aws-sdk/

adiafora: @scprek I think we should fix the bug in this PR. The use of other headers should be added in another PR, I think.

scprek (approved):

src/Instrumentation/AwsSdk/src/AwsSdkInstrumentation.php
ChrisLightfootWild: The additional header checks are fairly trivial, so no objections to adding them in this PR from me!

adiafora: Yes, of course. I added it))

ChrisLightfootWild: Amazing, thanks for your efforts!

ChrisLightfootWild (approved):

ChrisLightfootWild:
Available in https://github.com/opentelemetry-php/contrib-auto-aws-sdk/releases/tag/0.3.0

PR #553, opened 2026-04-23
Laravel: Improve DB query instrumentation

andrii-androshchuk:
PR to improve DB query instrumentation:

- Switch from `TraceAttributes` to typed `DbAttributes` and `ServerAttributes`;
- Add `server.address` and `server.port` attributes to DB query spans;
- Add a driver name mapping helper to normalize Laravel driver names (e.g. pgsql → postgresql, sqlsrv → microsoft_sql_server) to OTel semantic convention values.


src/Instrumentation/Laravel/src/Watchers/QueryWatcher.php
```diff
@@ -9,7 +9,8 @@
 use Illuminate\Support\Str;
 use OpenTelemetry\API\Instrumentation\CachedInstrumentation;
 use OpenTelemetry\API\Trace\SpanKind;
-use OpenTelemetry\SemConv\TraceAttributes;
+use OpenTelemetry\SemConv\Attributes\DbAttributes;
```
ChrisLightfootWild, line 12: This is not available in the minimum version of sem-conv that this instrumentation depends on currently.

Check https://github.com/opentelemetry-php/sem-conv/tree/1.36.0/Attributes

andrii-androshchuk: You're right, @ChrisLightfootWild. What would you suggest?

ChrisLightfootWild: It wouldn't be the first contrib package to hit `^1.36` ([ref](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-php-contrib+%22open-telemetry%2Fsem-conv%22&type=code)), so maybe just bump the dependency.

andrii-androshchuk: @ChrisLightfootWild, I bumped to `^1.38`, hope that's okay.

andrii-androshchuk:
Okay, now there is another problem:

```
src/Watchers/LogWatcher.php:110 PhanDeprecatedClass Using a deprecated class \OpenTelemetry\API\Instrumentation\ConfigurationResolver defined at vendor/open-telemetry/api/Instrumentation/ConfigurationResolver.php:12 (Deprecated because: environment variables should be accessed using {@link \OpenTelemetry\API\Configuration\ConfigEnv\EnvComponentLoader}s.)
```

It seems odd, and coming from the underlying API classes. @ChrisLightfootWild, please, advice.

andrii-androshchuk:
Now I see what happened, another (a year old) PR https://github.com/open-telemetry/opentelemetry-php-contrib/pull/491 was merged yesterday:


And, we're back to encoding context as a single JSON string:
```php
$logBuilder->setAttribute('context', json_encode($contextToProcess) ?: '{}');
```

---

I guess, this is one more reason, why logging should be separated from this package. It's a complete domain on its own, and there is plenty (at least two) distinct packages available for logging in Open Telemetry, which should be used instead.

ChrisLightfootWild:
> And, we're back to encoding context as a single JSON string:
> 
> ```
> $logBuilder->setAttribute('context', json_encode($contextToProcess) ?: '{}');
> ```

Were we already in this position or not? Changing this behaviour was not the intent, I was trying to close off some of our open PRs.

Still wanting to remove the LogWatcher entirely some time after I manage to land https://github.com/open-telemetry/opentelemetry-php-contrib/pull/269.

ChrisLightfootWild (approved):

PR #559, opened 2026-04-29
Revert "Enable context-flattening to support context field searching"

ChrisLightfootWild:
Reverts open-telemetry/opentelemetry-php-contrib#491

PR #560, opened 2026-04-30
[Laravel] Add opt-in context flattening for faceted log search

nickmarden:
## Description

This PR restores the context-flattening feature from #491 (which was reverted in #559 due to a conflict with #511), now correctly layered on top of the structured-array default that #511 introduced.

### Background

PR #491 identified a real observability problem: the Laravel `LogWatcher` was JSON-encoding the entire log context into a single `context` string attribute, making it impossible to filter by specific values in backends like SigNoz, Jaeger, or Grafana. For example, given context `{"http": {"method": "GET", "path": "/users"}, "user_id": "123"}`, there was no way to query `http.method = "GET"` as a facet.

PR #491 solved this with an opt-in `OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN` flag that spreads context fields as individual OTLP attributes using dot notation for nested keys:

```
http.method = GET
http.path   = /users
user_id     = 123
```

### What went wrong with #491

When #491 was merged, it conflicted with #511, which had already changed the default `context` handling from a JSON-encoded string to a structured array (`->setAttribute('context', $context)`). The #491 code regressed that improvement by reverting the non-flatten path back to `json_encode()`.

PR #559 reverted #491 entirely to restore the #511 behaviour, but that also lost the flattening feature.

### What this PR does

This PR re-introduces the flattening feature cleanly on top of #511:

- **Default behaviour (unchanged from #511):** context is stored as a structured array attribute, e.g. `->setAttribute('context', ['http' => ['method' => 'GET'], 'user_id' => '123'])`.
- **Opt-in flatten mode** (`OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN=true`): context fields are spread as individual top-level OTLP attributes using `Illuminate\Support\Arr::dot()`, enabling faceted search in observability backends.

The non-flatten default path is identical to what #511 shipped. The flatten path is additive and off by default.

### Changes

- `src/Watchers/LogWatcher.php`: adds `OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN` constant, reads it via `ConfigurationResolver` in the constructor, and branches on it in `recordLog()`. A `normalizeValue()` helper ensures OTLP-compatible scalar types (casting `Stringable` objects to string, JSON-encoding anything else).
- `tests/Unit/Watchers/LogWatcherTest.php`: 8 unit tests covering default structured-array mode, flatten mode, nested dot-notation expansion, exception extraction, `Stringable` normalisation, and null filtering.

### Backward compatibility

- Default behaviour is unchanged from #511.
- `OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN` is off by default; existing deployments are unaffected.

### DIFF
--- a/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
+++ b/src/Instrumentation/Laravel/src/Watchers/LogWatcher.php
@@ -7,17 +7,27 @@ namespace OpenTelemetry\Contrib\Instrumentation\Laravel\Watchers;
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Log\Events\MessageLogged;
 use Illuminate\Log\LogManager;
+use Illuminate\Support\Arr;
 use OpenTelemetry\API\Instrumentation\CachedInstrumentation;
+use OpenTelemetry\API\Instrumentation\ConfigurationResolver;
 use OpenTelemetry\API\Logs\Severity;
+use Stringable;
 use Throwable;
 use TypeError;
 
 class LogWatcher extends Watcher
 {
+    public const OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN = 'OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN';
+
     private LogManager $logger;
+    private bool $flattenAttributes;
+
     public function __construct(
         private CachedInstrumentation $instrumentation,
     ) {
+        $resolver = new ConfigurationResolver();
+        $this->flattenAttributes = $resolver->has(self::OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN)
+            && $resolver->getBoolean(self::OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN);
     }
 
     /** @psalm-suppress UndefinedInterfaceMethod */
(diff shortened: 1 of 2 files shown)


ChrisLightfootWild (commented):
Thanks for re-creating this with a comprehensive write-up 💪 

There's a failure in the pipeline, if you could take a look:
`1) tests/Unit/Watchers/LogWatcherTest.php (class_definition)`

nickmarden:
@ChrisLightfootWild I think I fixed the issue that you highlighted in your previous comment, but CI is generally red for reasons unrelated to this PR AFAICT.

Do you want me to take a stab at getting CI green (which would be a lot of scope creep), or is this PR OK as-is?

andrii-androshchuk:
@nickmarden, why do we need the flattening in the first place?

I understand the original problem with the entire context being stored as a single string. However, after [#511](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/511), the attributes are exposed as properties under the root `context` attribute (as it should), and everything inside it should already be available for faceted search.

**Because of that, I am wondering what additional benefit flattening provides here, if the faceted search is working now.**

ChrisLightfootWild:
> @ChrisLightfootWild I think I fixed the issue that you highlighted in your previous comment, but CI is generally red for reasons unrelated to this PR AFAICT.
> 
> Do you want me to take a stab at getting CI green (which would be a lot of scope creep), or is this PR OK as-is?

Apologies for my delayed response, I'm just back from vacation. Generally happy to delegate unrelated CI failures to separate PRs.

nickmarden:
> **Because of that, I am wondering what additional benefit flattening provides here, if the faceted search is working now.**

@andrii-androshchuk I wasn't under the impression that ALL components of the context would be surfaced as attributes in dotted notation such as `{"user_info", {"id": 12345}}` -> `user_info.id` but we're going to test it to see if that is true.

nickmarden:
> > **Because of that, I am wondering what additional benefit flattening provides here, if the faceted search is working now.**

@andrii-androshchuk Thanks for the question — @arthurmartelli and I looked into this carefully.

After #511, `setAttribute('context', ['user' => ['id' => 1234]])` is exported via OTLP as a single context attribute whose value is a nested kvlist AnyValue. In SigNoz's ClickHouse schema, log attributes are stored in flat typed maps (attributes_string, attributes_number, etc.). A kvlist value gets serialized as a JSON blob under the key context — there is no separate `user.id` or `context.user.id` entry in those maps.

So `user.id = 1234` is not available as a facet in SigNoz after #511. The context attribute is structurally better than the old JSON-encoded string, but SigNoz's faceted search still treats it as a single opaque value.

The flattening this PR adds explicitly emits `user.id = 1234` as a primitive-valued top-level attribute, which SigNoz does index as a searchable facet. That's the gap #511 doesn't close.

Attached are SigNoz screenshots showing

The surfaced top-level attribute `http.ip` in SigNoz (with `OTEL_PHP_LARAVEL_LOG_ATTRIBUTES_FLATTEN`): 

image

And the current `main` branch behavior (no flattening) in which `http.ip` is just part of the `context `attribute JSON ball (no flattening):

image (1)

My conclusion: I think we should still merge this to support the originally-intended use case.

ChrisLightfootWild (approved):

ChrisLightfootWild:
Thanks for contributing @nickmarden :+1:

PR #567, opened 2026-05-05
[Laravel] Add opt-in fastcgi_finish_request() hook in Kernel::terminate() to prevent blocking OTLP export from affecting PHP-FPM response time

arthurmartelli:
## Problem

Under PHP-FPM, the OTel SDK exports spans during the PHP process shutdown phase via a synchronous, blocking HTTP POST. This export runs *after* `Kernel::handle()` returns but *before* the FastCGI connection to the upstream web server (nginx, Apache) is closed. The sequence today is:

```text
nginx -> FPM worker -> handle() -> terminate() -> [PHP shutdown] -> OTLP export -> FastCGI close
                                                                                    |
                                                                                   upstream proxy waits until here
```

When the collector is slow or unreachable, the FPM worker is held for the full exporter timeout
(default: 10s). This causes two observable failures:

- **Upstream proxy timeouts** — the proxy is waiting for the FastCGI connection to close; if its own timeout (e.g. 2 s) fires first, the client receives a gateway error even though the application handled the request successfully.
- **FPM worker pool exhaustion** — all workers block on the export during a collector outage, causing incoming requests to queue or be rejected.

## Solution

PHP-FPM provides `fastcgi_finish_request()`, which flushes the response buffer and closes the FastCGI socket to the upstream server immediately, allowing the worker to continue executing PHP in the background. The upstream proxy and client see the request as complete; the OTel export runs without holding the connection open.

The correct place to call this in a Laravel application is the **post-hook of `Kernel::terminate()`**, after the response has been sent and all terminating middleware have run, but before PHP shutdown handlers (including `TracerProvider::shutdown()`) fire.

This PR adds that hook as an **opt-in feature**, disabled by default.

### `src/Hooks/Illuminate/Contracts/Http/Kernel.php`

Adds `hookTerminate()`, registered unconditionally from `instrument()` alongside the existing `hookHandle()`. The hook's post-callback delegates entirely to `FpmFinishRequest::handle()`, which evaluates the feature flag and runtime guards on every call. The hook registration itself is cheap (a single cached property read per request when the feature is disabled).

### `src/Support/FpmFinishRequest.php` *(new)*

The guard and dispatch class. Responsibilities:

- **Feature flag** — reads `OTEL_PHP_INSTRUMENTATION_LARAVEL_FPM_FINISH_REQUEST_ENABLED` once and caches it for the process lifetime.
- **Long-running runtime guard** — detects Laravel Octane (Swoole / RoadRunner) via:
  1. `LARAVEL_OCTANE` environment variable (boolean-validated; unrecognised values are treated as `true`, fail-closed, with a warning).
  2. Laravel container bindings (`octane`, `octane.client`, `octane.worker`) as a fallback for environments that set neither env variable. Fires a deduplicated warning and skips `fastcgi_finish_request()` in both cases. **Do not enable this feature in Octane / long-running worker deployments.**
- **No-op outside FPM** — `fastcgi_finish_request()` is a PHP-FPM C built-in and is not defined in CLI or test environments. The code checks `function_exists()` before calling it.
- **Testability** — `setFinishRequestCallback()` injects a callable that replaces the `fastcgi_finish_request()` call, allowing full unit-test coverage without requiring FPM. The override path wraps the callback in a `try/catch(Throwable)` and logs a warning on failure.
- **Warning deduplication** — long-running runtime warnings are keyed on message text and emitted at most once per distinct message per process lifetime, preventing log flooding during sustained outages.

### `tests/Unit/Support/FpmFinishRequestTest.php` *(new)*

Full PHPUnit coverage via the callback injection path:

| Test                                                              | What it verifies                                                                    |
| ----------------------------------------------------------------- | ----------------------------------------------------------------------------------- |
| `test_is_not_enabled_by_default` (data-driven)                    | `false`, `0`, `off`, `no`, unset, empty, and invalid values all disable the feature |
| `test_is_enabled_when_env_is_truthy` (data-driven)                | `true`, `1`, `yes`, `on` all enable the feature                                     |
| `test_is_enabled_when_server_signal_is_scalar`                    | Non-string `$_SERVER` values (e.g. integer `1`) are coerced correctly               |
| `test_is_enabled_result_is_cached`                                | Re-reading env after first call does not change the cached result                   |
| `test_finish_request_is_called_when_enabled`                      | Callback fires when the feature is enabled                                          |
| `test_finish_request_is_not_called_when_disabled`                 | Callback is never invoked when the feature is disabled                              |
| `test_finish_request_is_not_called_when_octane_env_signal_is_set` | `LARAVEL_OCTANE=1` suppresses the call                                              |
| `test_explicit_octane_false_allows_finish_request`                | `LARAVEL_OCTANE=false` does not suppress the call                                   |
| `test_explicit_octane_zero_allows_finish_request`                 | `LARAVEL_OCTANE=0` does not suppress the call                                       |
| `test_warning_is_logged_when_octane_guard_fires`                  | A warning is emitted via `LogsMessagesTrait` when Octane is detected                |
| `test_octane_warning_is_logged_only_once`                         | The Octane warning is deduplicated across multiple `handle()` calls                 |
| `test_distinct_long_running_warnings_are_both_logged`             | Distinct warning messages are deduplicated independently                            |
| `test_warning_is_logged_when_octane_signal_is_invalid`            | Unrecognised `LARAVEL_OCTANE` values emit a specific warning and fail-closed        |
| `test_handle_delegates_every_call`                                | `handle()` does not itself deduplicate; the callback fires on every call            |
| `test_warning_is_logged_when_finish_request_callback_throws`      | A throwing callback logs a warning and does not propagate                           |
| `test_reset_cache_clears_warning_dedup_state`                     | `resetCache()` clears all cached state including the warning dedup map              |

The `fastcgi_finish_request() === false` error path is behind `@codeCoverageIgnoreStart` because
the C built-in cannot be mocked in unit tests; the callback path covers all reachable branches.

### `README.md`

Documents the feature, the required companion timeout configuration, and the Octane constraint.

## Configuration

Enable for PHP-FPM workloads only:

```shell
OTEL_PHP_INSTRUMENTATION_LARAVEL_FPM_FINISH_REQUEST_ENABLED=true
```

**Also configure exporter timeouts** to bound background worker hold time during collector
outages:

```shell
OTEL_EXPORTER_OTLP_TIMEOUT=500
OTEL_EXPORTER_OTLP_TRACES_TIMEOUT=500
OTEL_BSP_EXPORT_TIMEOUT=1000
```

Without these, FPM workers are still held for the SDK's default 10 s timeout during a full
outage. With them, background hold time is bounded to ≈ 600 ms per request.

## Notes

- The feature is **opt-in and off by default**. Existing users are entirely unaffected until they set the environment variable.
- Outside FPM (CLI, test runners, Octane), the feature is a no-op or emits a single warning. There are no side effects on non-FPM runtimes.

### DIFF
--- a/src/Aws/composer.json
+++ b/src/Aws/composer.json
@@ -31,7 +31,7 @@
         "phpstan/phpstan": "^1.4",
         "phpstan/phpstan-phpunit": "^1.0",
         "phpunit/phpunit": "~9",
-        "psalm/plugin-phpunit": "^0.20.0",
+        "psalm/plugin-phpunit": "^0.19.2",
         "vimeo/psalm": "6.4.0"
     },
     "autoload": {
--- a/src/Aws/composer.json
+++ b/src/Aws/composer.json
@@ -32,7 +32,7 @@
         "phpstan/phpstan-phpunit": "^1.0",
         "phpunit/phpunit": "~9",
         "psalm/plugin-phpunit": "^0.20.0",
-        "vimeo/psalm": "6.16.1"
+        "vimeo/psalm": "6.4.0"
     },
     "autoload": {
         "psr-4": {
--- a/src/Context/Swoole/composer.json
+++ b/src/Context/Swoole/composer.json
@@ -28,7 +28,7 @@
     "php-http/mock-client": "*",
     "phpstan/phpstan": "^1.1",
     "phpstan/phpstan-phpunit": "^1.0",
-    "psalm/plugin-phpunit": "^0.20.0",
+    "psalm/plugin-phpunit": "^0.19.2",
     "open-telemetry/sdk": "^1.0",
     "phpunit/phpunit": "^9.5",
     "vimeo/psalm": "6.4.0"
--- a/src/Context/Swoole/composer.json
+++ b/src/Context/Swoole/composer.json
@@ -31,6 +31,6 @@
     "psalm/plugin-phpunit": "^0.20.0",
     "open-telemetry/sdk": "^1.0",
     "phpunit/phpunit": "^9.5",
-    "vimeo/psalm": "6.16.1"
+    "vimeo/psalm": "6.4.0"
   }
 }
--- a/src/Exporter/Instana/composer.json
+++ b/src/Exporter/Instana/composer.json
@@ -27,7 +27,7 @@
         "phan/phan": "^5.0",
         "phpstan/phpstan": "^1.1",
         "phpstan/phpstan-phpunit": "^1.0",
-        "psalm/plugin-phpunit": "^0.20.0",
+        "psalm/plugin-phpunit": "^0.19.2",
         "open-telemetry/sdk": "^1.0",
         "phpunit/phpunit": "^9.5",
         "vimeo/psalm": "^4|^5|^6",
--- a/src/Instrumentation/AwsSdk/composer.json
+++ b/src/Instrumentation/AwsSdk/composer.json
@@ -23,7 +23,7 @@
     "phpstan/phpstan-mockery": "^1.1.0",
     "phpstan/phpstan": "^1.1",
     "phpstan/phpstan-phpunit": "^1.0",
-    "psalm/plugin-phpunit": "^0.20.0",
+    "psalm/plugin-phpunit": "^0.19.2",
     "open-telemetry/sdk": "^1.0",
     "phpunit/phpunit": "^9.5",
     "vimeo/psalm": "6.4.0"
--- a/src/Instrumentation/AwsSdk/composer.json
+++ b/src/Instrumentation/AwsSdk/composer.json
@@ -26,7 +26,7 @@
     "psalm/plugin-phpunit": "^0.20.0",
     "open-telemetry/sdk": "^1.0",
     "phpunit/phpunit": "^9.5",
-    "vimeo/psalm": "6.16.1"
+    "vimeo/psalm": "6.4.0"
   },
   "autoload": {
     "psr-4": {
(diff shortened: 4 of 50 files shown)


arthurmartelli:
@ChrisLightfootWild seems like the commits [fecdca74c64b](https://github.com/open-telemetry/opentelemetry-php-contrib/commit/fecdca74c64b31677c2c6cb1d437b08419201039) and [0e17efeb451a](https://github.com/open-telemetry/opentelemetry-php-contrib/commit/0e17efeb451a955129491d8963881fb1a66ca795) broke the pipelines :/

bobstrecansky:
We'll get the pipelines fixed up next week, sorry about that.

This PR is pretty large.  Is there any way to break it into smaller pieces?

grzegorz-daszuta:
I struggle to see why this functionality is required – this functionality is already provided in a more comprehensive way in https://github.com/symfony/http-foundation/blob/8.1/Response.php#L405, so the only situation when the described blocking would happen would be if there was a direct output from the handler instead of Response contract.

Downgrading all the dependencies across all the unaffected packages is a bad form and reeks of misuse of unsupervised coding agent.

arthurmartelli:
Thanks for the feedback, this is fair.

You are right that in the standard Laravel flow, Response::send() already calls fastcgi_finish_request(), so calling it again in Kernel::terminate() is likely redundant. I will re-check the rationale and either narrow this to non-standard flows or remove it.

About this part:

> Downgrading all the dependencies across all the unaffected packages is a bad form and reeks of misuse of unsupervised coding agent.

Those are the 2 fixup! commits in the branch history. They were added only as a temporary workaround to unblock CI (composer install failures in unaffected packages), and can be dropped once the underlying dependency issue is fixed. They are not AI-generated changes, just a pipeline workaround.

PR #570, opened 2026-05-07
Adding magento2 instrumentation

jerrytfleung:
This pull request introduces initial support for Magento 2 instrumentation in the project. The main changes add the necessary configuration, registration, and repository split logic to treat Magento 2 as a first-class instrumentation target, alongside updates to CI and code quality tooling to support the new integration.

**Magento 2 Instrumentation Integration:**

* Added `src/Instrumentation/Magento2/` directory with configuration files for code style (`.php-cs-fixer.php`), static analysis (`.phan/config.php`), `.gitignore`, and `.gitattributes`. [[1]](diffhunk://#diff-13301fece7c32cdce2e361c0d1b80c68d613aba2881ea8527cc922baaab62978R1-R12) [[2]](diffhunk://#diff-ecd68a32ba7234a795127e7b66583e3292e0341cc8f647149c5b35649d88293eR1) [[3]](diffhunk://#diff-835fcb5d17844b19acb5cfdfa993408e5926f42cf158e4f46cd62ab873c62b7dR1-R371) [[4]](diffhunk://#diff-e424daa1fef086ce31d9f255ece930ec91871ba32311d81db275a60ff96b4a43R1-R43)

**Composer and Autoloading Updates:**

* Updated `composer.json` to:
  * Register the Magento 2 source and test namespaces for autoloading and test discovery. [[1]](diffhunk://#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R33) [[2]](diffhunk://#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R116)
  * Include the Magento 2 `_register.php` file for instrumentation registration.
  * Add `open-telemetry/opentelemetry-auto-magento2` as a dependency.

**Repository and CI Configuration:**

* Updated `.gitsplit.yml` to enable splitting the Magento 2 instrumentation into its own repository.
* Updated `.github/workflows/php.yml` to:
  * Include Magento 2 in the matrix for CI jobs.
  * Test Magento 2 instrumentation with PHP 8.1 and 8.2. [[1]](diffhunk://#diff-a73bb6555480a5ee79ae276a3f5d71a08fa316e09a4a8da7b643cf1e92c97df9R41) [[2]](diffhunk://#diff-a73bb6555480a5ee79ae276a3f5d71a08fa316e09a4a8da7b643cf1e92c97df9R92-R95)

Require the merge of https://github.com/open-telemetry/opentelemetry-php/pull/1952 to provide the necessary prerequisite for the GitHub action of magento instrumentation


src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,261 @@
...
+                    $currentSpan->recordException($exception);
+                    $currentSpan->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
+                }
+                $currentSpan->end();
+            }
+        );
+
+        hook(
+            Http::class,
+            'launch',
```
jerrytfleung, line 84: @open-telemetry/php-approvers I am not sure if the following is a real issue or not.

I am hooking into the `Http::launch` function and creating the server root span. However, from the screenshot and the [spans.txt](https://github.com/user-attachments/files/27492187/spans.txt). The magento interceptor plugin which is intercepting `Http::launch()` creates the server root span instead of the actual `Http::launch` function. When I check the Interceptor code, the interceptor class extends the `Http` and therefore it is also a Http object. I think this is why the prehook and posthook are called. 
I am not sure if it is a real issue or is there anything I could do to hook into the actual `Http::launch` function. Please help to advise. Thanks!
```
/**
 * Interceptor class for @see \Magento\Framework\App\Http
 */
class Interceptor extends \Magento\Framework\App\Http implements \Magento\Framework\Interception\InterceptorInterface
{
    use \Magento\Framework\Interception\Interceptor;

   ...

    /**
     * {@inheritdoc}
     */
    public function launch()
    {
        $pluginInfo = $this->pluginList->getNext($this->subjectType, 'launch');
        return $pluginInfo ? $this->___callPlugins('launch', func_get_args(), $pluginInfo) : parent::launch();
    }

    ...
}
```

Screenshot 2026-05-07 at 10 34 23 AM

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,261 @@
...
+                );
+                if ($span instanceof ReadableSpanInterface) {
+                    $spanAttrs = [
+                        HttpAttributes::HTTP_REQUEST_METHOD => $span->getAttribute(HttpAttributes::HTTP_REQUEST_METHOD),
+                        UrlAttributes::URL_SCHEME => $span->getAttribute(UrlAttributes::URL_SCHEME),
+                        NetworkAttributes::NETWORK_PROTOCOL_VERSION => $span->getAttribute(NetworkAttributes::NETWORK_PROTOCOL_VERSION),
+                    ];
+                    /** @psalm-suppress PossiblyInvalidArgument */
+                    $histogram->record($span->getDuration() / ClockInterface::NANOS_PER_SECOND, array_merge($spanAttrs, $responseMeta));
+                }
```
Nevay: Should not read values from the span, see https://opentelemetry.io/docs/specs/otel/trace/sdk/#additional-span-interfaces:
> The [API-level definition for Span’s interface](https://opentelemetry.io/docs/specs/otel/trace/api/#span-operations) only defines write-only access to the span. This is good because instrumentations and applications are not meant to use the data stored in a span for application logic.

Additionally this won't record metrics if the span is non-recording.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,261 @@
...
+                    ->setAttribute(HttpAttributes::HTTP_REQUEST_METHOD, $request->getMethod())
+                    ->setAttribute(NetworkAttributes::NETWORK_PROTOCOL_VERSION, $request->getProtocolVersion())
+                    ->setAttribute(UserAgentAttributes::USER_AGENT_ORIGINAL, $request->getHeaderLine('User-Agent'))
+                    ->setAttribute(HttpIncubatingAttributes::HTTP_REQUEST_BODY_SIZE, $request->getHeaderLine('Content-Length'))
+                    ->setAttribute(ServerAttributes::SERVER_ADDRESS, $request->getUri()->getHost())
+                    ->setAttribute(ServerAttributes::SERVER_PORT, $request->getUri()->getPort() ?? ($request->getUri()->getScheme() === 'https' ? 443 : 80));
+
+                /** @psalm-suppress MixedAssignment */
+                foreach ($request->getHeaders() as $key => $value) {
+                    $spanBuilder->setAttribute(HttpAttributes::HTTP_REQUEST_HEADER . '.' . $key, implode(',', $value));
```
Nevay: https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/#opt-in
> Instrumentation that doesn’t support configuration MUST NOT populate Opt-In attributes.

jerrytfleung: I see. Removed all opt-in attributes.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: I see. Changed to use the scope to save the `$requestMeta`.

Yes, the pre/post hook is only triggered when the span is recording. 
I am trying to create http.server.request.duration histogram metrics similar to the one in https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/ReactPHP/src/ReactPHPInstrumentation.php#L185

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,366 @@
...
+                if (!$scope) {
+                    return;
+                }
+                $scope->detach();
+                $currentSpan = Span::fromContext($scope->context());
+                if ($exception) {
+                    $currentSpan->recordException($exception);
+                    $currentSpan->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
+                }
+                $currentSpan->end();
```
Nevay:
```suggestion
                if ($exception) {
                    $span->recordException($exception);
                    $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
                }
                $span->end();
```

---

Should this be modelled as log event instead?

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
Nevay: You could check for `$class === Http::class` in the pre and post callbacks.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,366 @@
...
+                    ->spanBuilder(sprintf('%s %s', $request->getMethod(), self::getScriptNameFromRequest($request)))
+                    ->setParent($parent)
+                    ->setSpanKind(SpanKind::KIND_SERVER)
+                    ->setAttribute(CodeAttributes::CODE_FUNCTION_NAME, sprintf('%s::%s', $class, $function))
+                    ->setAttribute(CodeAttributes::CODE_FILE_PATH, $filename)
+                    ->setAttribute(CodeAttributes::CODE_LINE_NUMBER, $lineno)
+                    ->setAttribute(UrlAttributes::URL_FULL, (string) $request->getUri())
+                    ->setAttribute(UrlAttributes::URL_SCHEME, $request->getUri()->getScheme())
+                    ->setAttribute(UrlAttributes::URL_PATH, $request->getUri()->getPath())
+                    ->setAttribute(HttpAttributes::HTTP_REQUEST_METHOD, $request->getMethod())
```
Nevay: Must set unknown request methods to `_OTHER`.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,366 @@
...
+                $parent = Globals::propagator()->extract($request->getHeaders());
+
+                $spanBuilder = $instrumentation->tracer()
+                    ->spanBuilder(sprintf('%s %s', $request->getMethod(), self::getScriptNameFromRequest($request)))
+                    ->setParent($parent)
+                    ->setSpanKind(SpanKind::KIND_SERVER)
+                    ->setAttribute(CodeAttributes::CODE_FUNCTION_NAME, sprintf('%s::%s', $class, $function))
+                    ->setAttribute(CodeAttributes::CODE_FILE_PATH, $filename)
+                    ->setAttribute(CodeAttributes::CODE_LINE_NUMBER, $lineno)
+                    ->setAttribute(UrlAttributes::URL_FULL, (string) $request->getUri())
```
Nevay:
```suggestion
```

Only defined for client spans.
Should instead set `url.query` if available.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,366 @@
...
+                    ->setAttribute(UrlAttributes::URL_SCHEME, $request->getUri()->getScheme())
+                    ->setAttribute(UrlAttributes::URL_PATH, $request->getUri()->getPath())
+                    ->setAttribute(HttpAttributes::HTTP_REQUEST_METHOD, $request->getMethod())
+                    ->setAttribute(NetworkAttributes::NETWORK_PROTOCOL_VERSION, $request->getProtocolVersion())
+                    ->setAttribute(UserAgentAttributes::USER_AGENT_ORIGINAL, $request->getHeaderLine('User-Agent'))
+                    ->setAttribute(ServerAttributes::SERVER_ADDRESS, $request->getUri()->getHost())
+                    ->setAttribute(ServerAttributes::SERVER_PORT, $request->getUri()->getPort() ?? ($request->getUri()->getScheme() === 'https' ? 443 : 80));
```
Nevay: Should attempt to set the server address and port according to https://opentelemetry.io/docs/specs/semconv/http/http-spans/#setting-serveraddress-and-serverport-attributes.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,366 @@
...
+                $currentSpan->end();
+            }
+        );
+
+        hook(
+            Http::class,
+            'launch',
+            pre: static function (Http $http, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                $factory = new Psr17Factory();
+                $request = (new ServerRequestCreator($factory, $factory, $factory, $factory))->fromGlobals();
```
Nevay, line 90: Could this reuse / share the implementation with the existing PSR-15 instrumentation? The code doesn't seem to contain Magento specific parts.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,366 @@
...
+                    'Duration of HTTP server requests.',
+                    ['ExplicitBucketBoundaries' => [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10]]
+                );
+
+                $requestMeta = $scope->offsetGet('requestMeta');
+                /** @var int $requestStart */
+                $requestStart = $scope->offsetGet('requestStart');
+                /** @psalm-suppress PossiblyInvalidArgument */
+                $histogram->record((float) (($requestEnd - $requestStart) / ClockInterface::NANOS_PER_SECOND), array_merge((array) $requestMeta, $responseMeta));
+                $span->end();
```
Nevay:
```suggestion
                $span->end($requestEnd);
```

For consistency with the pre hook (`$spanBuilder->setStartTimestamp($requestStart)`).

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,366 @@
...
+                $responseMeta = [];
+
+                if ($exception) {
+                    $span->recordException($exception);
+                    $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
+                    $responseMeta[ErrorAttributes::ERROR_TYPE] = (string) $exception->getCode();
+                }
+                if ($response instanceof HttpResponse) {
+                    $span->setAttribute(HttpAttributes::HTTP_RESPONSE_STATUS_CODE, $response->getStatusCode());
+                    $responseMeta[HttpAttributes::HTTP_RESPONSE_STATUS_CODE] = $response->getStatusCode();
```
Nevay: Should set `error.type` if status code is `5xx`.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Sure. Done

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Sure. Added `UrlAttributes::URL_QUERY` if present

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: I am not sure how to model as log event. Do you have an example?

src/Instrumentation/Magento2/composer.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "open-telemetry/opentelemetry-auto-magento2",
+  "description": "OpenTelemetry auto-instrumentation for Magento2",
+  "keywords": ["opentelemetry", "otel", "open-telemetry", "tracing", "magento2", "instrumentation"],
+  "type": "library",
+  "homepage": "https://opentelemetry.io/docs/languages/php/",
+  "readme": "./README.md",
+  "license": "Apache-2.0",
+  "require": {
+    "php": "^8.3",
```
bobstrecansky, line 10: **Blocker — PHP version conflict with CI matrix.** This package requires `^8.3`, but `.github/workflows/php.yml` (lines 92–95) adds matrix entries pinning Magento2 to PHP 8.1 and 8.2. Composer install will fail on both rows, so those CI jobs cannot pass as written.

Every other instrumentation in this repo uses `^8.1`. Either:
- Relax to `"php": "^8.1"` to match the rest of the repo, **or**
- Drop the 8.1/8.2 `include` entries in `php.yml` (the default matrix already covers 8.1–8.4, so you only need `include:` entries if you want to *exclude* versions).

Also, if you stay on `^8.3`, update `.phan/config.php` `target_php_version` accordingly. If you relax to `^8.1`, also drop the `#[Override]` attributes (they're PHP 8.3-only).

src/Instrumentation/Magento2/composer.json
```diff
@@ -0,0 +1,54 @@
...
+    "ext-opentelemetry": "*",
+    "ext-json": "*",
+    "open-telemetry/api": "^1.6",
```
bobstrecansky, line 12: **Blocker — sem-conv version incompatible with the rest of the repo.** Every other instrumentation pins `open-telemetry/sem-conv: ^1.30`. The `Attributes\\*` split namespace that this instrumentation uses heavily was introduced in sem-conv 1.32+, so requiring `^1.36` here will cause autoload failures for consumers of the root contrib package that resolve to `^1.30`.

Either:
- Drop to `^1.30` and switch to the flat `TraceAttributes` constants (matches every other instrumentation), or
- Coordinate a repo-wide bump of sem-conv before merging this PR.

The PR description mentions a prerequisite PR on `opentelemetry-php` — please make the dependency relationship explicit, and confirm whether a coordinated bump across the contrib monorepo is in scope here.

src/Instrumentation/Magento2/composer.json
```diff
@@ -0,0 +1,54 @@
...
+    "php-http/mock-client": "*",
+    "phpstan/phpstan-mockery": "^1.1.0",
+    "phpstan/phpstan": "^1.1",
+    "phpstan/phpstan-phpunit": "^1.0",
+    "psalm/plugin-phpunit": "^0.20.0",
+    "open-telemetry/sdk": "^1.8",
+    "phpunit/phpunit": "^9.5",
+    "vimeo/psalm": "^7.0",
+    "magento/community-edition": "dev-2.4-develop"
+  },
```
bobstrecansky, line 31: **Should fix — `magento/community-edition: dev-2.4-develop` in `require-dev` is very heavy.** This is a 200+ MB metapackage that pulls in the entire Magento storefront, admin, modules, and front-end build chain. CI install time will be dominated by this.

The unit tests in this PR all use `getMockBuilder()` / `TestFramework\Unit\Helper\ObjectManager` and don't appear to need a full Magento app fixture. If `magento/framework` (already in `require`) is sufficient to run the unit tests, please drop this. If it's needed for one specific fixture, narrow the dep to the specific module.

At minimum, add a comment explaining why the full edition is required, given the install cost.

.github/workflows/php.yml
```diff
@@ -88,6 +89,10 @@ jobs:
             php-version: 8.1
           - project: 'Instrumentation/Session'
             php-version: 8.1
+          - project: 'Instrumentation/Magento2'
+            php-version: 8.1
+          - project: 'Instrumentation/Magento2'
+            php-version: 8.2
```
bobstrecansky, line 95: These matrix entries pin Magento2 to PHP 8.1 and 8.2, but `src/Instrumentation/Magento2/composer.json` requires `"php": "^8.3"` — composer install will fail on both rows. See the related comment on `composer.json:10`. Decide whether to relax the package constraint to `^8.1` or drop these matrix entries.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,361 @@
...
+                $spanBuilder = $instrumentation->tracer()
+                    ->spanBuilder(sprintf('%s %s', $request->getMethod(), self::getScriptNameFromRequest($request)))
+                    ->setParent($parent)
+                    ->setSpanKind(SpanKind::KIND_SERVER)
+                    ->setAttribute(CodeAttributes::CODE_FUNCTION_NAME, sprintf('%s::%s', $class, $function))
+                    ->setAttribute(CodeAttributes::CODE_FILE_PATH, $filename)
+                    ->setAttribute(CodeAttributes::CODE_LINE_NUMBER, $lineno)
+                    ->setAttribute(UrlAttributes::URL_SCHEME, $request->getUri()->getScheme())
+                    ->setAttribute(UrlAttributes::URL_PATH, $request->getUri()->getPath())
+                    ->setAttribute(HttpAttributes::HTTP_REQUEST_METHOD, strlen($request->getMethod()) > 0 ? $request->getMethod() : '_OTHER')
```
bobstrecansky: **Blocker — `_OTHER` semantics are wrong.** The [HTTP semconv](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-request-method) says `_OTHER` must be used when the method is **not in the standard set** (GET/POST/PUT/DELETE/HEAD/OPTIONS/PATCH/TRACE/CONNECT), and the original method must be set on `http.request.method_original`. This `strlen() > 0` check only handles empty strings — an arbitrary `FOO` method will be sent verbatim as the attribute value, producing high-cardinality data.

Suggested fix:
```php
$knownMethods = ['GET','POST','PUT','DELETE','HEAD','OPTIONS','PATCH','TRACE','CONNECT'];
$rawMethod = $request->getMethod();
$method = in_array($rawMethod, $knownMethods, true) ? $rawMethod : '_OTHER';
$spanBuilder->setAttribute(HttpAttributes::HTTP_REQUEST_METHOD, $method);
if ($method === '_OTHER') {
    $spanBuilder->setAttribute(HttpAttributes::HTTP_REQUEST_METHOD_ORIGINAL, $rawMethod);
}
```
IMPORTANT: the bucketed `$method` (not the raw value) must also flow into `$requestMeta` used by the duration histogram on line ~133, otherwise the metric retains the high-cardinality original. `http.request.method_original` must NOT appear on the metric (the server histogram has a fixed low-cardinality attribute set per semconv).

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,361 @@
...
+                }
+                $scope->detach();
+                $span = Span::fromContext($scope->context());
+
+                $responseMeta = [];
+
+                if ($exception) {
+                    $span->recordException($exception);
+                    $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
+                    $responseMeta[ErrorAttributes::ERROR_TYPE] = (string) $exception->getCode();
```
bobstrecansky: **Blocker — `error.type` is misused.** Per the [semantic conventions](https://opentelemetry.io/docs/specs/semconv/attributes-registry/error/#error-type), `error.type` is a *low-cardinality identifier* — typically the exception class name. `$exception->getCode()` returns the exception's integer code, which is `0` for the vast majority of PHP/Magento exceptions, producing useless `"0"` metric labels and effectively breaking the request-duration histogram for the most common error case.

Fix:
```php
$responseMeta[ErrorAttributes::ERROR_TYPE] = $exception::class;
```

Also: reviewer feedback earlier in the thread asked you to set `error.type` on `5xx` responses too — that branch is still missing. Add an else-if for `$statusCode >= 500` that sets `error.type` to the HTTP status (per the spec's HTTP error semantics).

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,361 @@
...
+        hook(
+            Http::class,
+            'launch',
+            pre: static function (Http $http, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                $factory = new Psr17Factory();
+                $request = (new ServerRequestCreator($factory, $factory, $factory, $factory))->fromGlobals();
+                $parent = Globals::propagator()->extract($request->getHeaders());
+
+                $spanBuilder = $instrumentation->tracer()
+                    ->spanBuilder(sprintf('%s %s', $request->getMethod(), self::getScriptNameFromRequest($request)))
```
bobstrecansky: **Blocker — `Http::launch` will double-fire on every request due to Magento's Interceptor pattern.** Magento generates an `Http\Interceptor extends Http` class for every plugin-able class, so this hook fires once for `Http::launch()` and once for `Http\Interceptor::launch()` — producing **duplicate server root spans** and **double-counting** the `http.server.request.duration` histogram.

Nevay flagged this earlier in the thread with a screenshot; the suggested guard is not in the current diff. Add to both pre and post callbacks at the very top:
```php
if ($class !== Http::class) {
    return;
}
```
(Without this guard on `post:`, you'll detach the wrong scope on the duplicate run.)

Please also verify against a real Magento install that none of the other hooks here double-fire — most target interfaces/abstract classes where the interceptor model behaves differently, but it's worth confirming.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,361 @@
...
+use Magento\Framework\App\Http;
+use Magento\Framework\App\Request\Http as HttpRequest;
+use Magento\Framework\App\Response\Http as HttpResponse;
+use Magento\Framework\App\ResponseInterface;
+use Magento\Framework\App\View;
+use Magento\Framework\Controller\ResultInterface;
+use Magento\Framework\Event\InvokerInterface;
+use Magento\Framework\Event\Manager;
+use Magento\Framework\View\Element\Template;
+use Nyholm\Psr7Server\ServerRequestCreator;
```
bobstrecansky, line 21: **Blocker — undeclared runtime dependencies on `nyholm/psr7` and `nyholm/psr7-server`.** This file imports `Nyholm\Psr7Server\ServerRequestCreator` (here) and `Http\Discovery\Psr17Factory` (line 7), but neither package is declared in `composer.json`. Compare to the Wordpress instrumentation which explicitly declares:
```json
"nyholm/psr7": "^1",
"nyholm/psr7-server": "^1"
```

Also recommend switching from `Http\Discovery\Psr17Factory` to `Nyholm\Psr7\Factory\Psr17Factory` (which Wordpress uses): the discovery variant adds a `php-http/discovery` runtime dep with no functional benefit, and your `composer.json` already disables `php-http/discovery` in `allow-plugins` (which means it can't auto-register anyway).

Without these deps declared, the package will only work on hosts that happen to install them transitively.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,361 @@
...
+            }
+        );
+
+        hook(
+            Manager::class,
+            'dispatch',
+            pre: static function (Manager $manager, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                $eventName = (string) ($params[0]);
+                $span = $instrumentation->tracer()
+                    ->spanBuilder('EVENT: ' . $eventName)
```
bobstrecansky: **Should fix — span naming style.** `EVENT: ` (and `OBSERVER: ` below) use uppercase ALL-CAPS prefixes that diverge from the lowercase namespacing every other OTel instrumentation uses (`db.query`, `http.request`, etc.). This makes Magento spans visually inconsistent in trace UIs.

Consider `event.dispatch {eventName}` and `observer.dispatch {observerName}`, or drop the prefix entirely and use a dedicated attribute like `magento.event.name` for filtering.

Also: `$eventName = (string) ($params[0])` on line ~244 — if a third-party module passes a non-string, the cast silently produces `"Array"` or worse. Wrap with the same `is_string && !== ''` guard you use in `InvokerInterface::dispatch`.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,361 @@
...
+                    ->setAttribute(CodeAttributes::CODE_FUNCTION_NAME, sprintf('%s::%s', $class, $function))
+                    ->setAttribute(CodeAttributes::CODE_FILE_PATH, $filename)
+                    ->setAttribute(CodeAttributes::CODE_LINE_NUMBER, $lineno)
+                    ->startSpan();
+                Context::storage()->attach($span->storeInContext(Context::getCurrent()));
+            },
+            post: static function (Manager $manager, array $params) {
```
bobstrecansky: **Should fix — post-closure swallows exceptions.** This signature is `post: static function (Manager $manager, array $params)` — it doesn't accept the `?Throwable $exception` parameter, so if an event listener throws, the span ends without `recordException()` or `STATUS_ERROR`. The `InvokerInterface::dispatch` post-closure below has the same issue.

Every other post-hook in this file handles exceptions; please make these two consistent:
```php
post: static function (Manager $manager, array $params, mixed $return, ?Throwable $exception) {
    $scope = Context::storage()->scope();
    if (!$scope) { return; }
    $scope->detach();
    $span = Span::fromContext($scope->context());
    if ($exception) {
        $span->recordException($exception);
        $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
    }
    $span->end();
}
```

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,361 @@
...
+                    : 'unknown';
+                $span = $instrumentation->tracer()
+                    ->spanBuilder('OBSERVER: ' . $observerName)
+                    ->setAttribute(CodeAttributes::CODE_FUNCTION_NAME, sprintf('%s::%s', $class, $function))
+                    ->setAttribute(CodeAttributes::CODE_FILE_PATH, $filename)
+                    ->setAttribute(CodeAttributes::CODE_LINE_NUMBER, $lineno)
+                    ->startSpan();
+                Context::storage()->attach($span->storeInContext(Context::getCurrent()));
+            },
+            post: static function (InvokerInterface $invokerInterface, array $params) {
```
bobstrecansky: Same as the `Manager::dispatch` post-hook above — this signature is missing the `?Throwable $exception` parameter, so observer exceptions go unrecorded. Please mirror the exception-handling pattern used in `Http::launch`, `Action::dispatch`, etc.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,361 @@
...
+            'https://opentelemetry.io/schemas/1.32.0',
+        );
+
+        hook(
+            Bootstrap::class,
+            'terminate',
+            pre: static function (Bootstrap $bootstrap, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                $exception = $params[0] instanceof Throwable ? $params[0] : null;
+                $span = $instrumentation->tracer()
+                    ->spanBuilder('Bootstrap::terminate')
```
bobstrecansky: **Should fix — `Bootstrap::terminate` creates a span even on successful requests.** `Bootstrap::terminate(Throwable $e)` is called on every request termination in Magento, not only on errors. The current code always creates a span even when `$params[0]` is null/non-throwable, producing a useless `Bootstrap::terminate` span sibling to `Http::launch` for every successful request.

Either gate span creation on `$exception !== null`, or consider whether this hook is needed at all — `Http::launch` already covers the full request lifecycle, including exceptions.

src/Instrumentation/Magento2/app/etc/vendor_path.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Path to Composer vendor directory
+ */
+return './vendor';
```
bobstrecansky: **Should fix — this file shouldn't ship in the package.** `app/etc/vendor_path.php` is a Magento installation bootstrap file. Shipping it inside an OTel instrumentation package means anyone who installs `opentelemetry-auto-magento2` gets an unexpected `app/etc/` directory inside their vendor tree.

If you genuinely need this for tests, move it under `tests/` and add `tests` to `.gitattributes export-ignore` (which it already is). Otherwise, please remove the file. `.gitattributes` currently doesn't `export-ignore` `app/`, so this **will end up in the published archive**.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,361 @@
...
+                    $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
+                }
+                $span->end();
+            }
+        );
+    }
+
+    private static function getScriptNameFromRequest(ServerRequestInterface $request): string
+    {
+        $scriptName = $request->getServerParams()['SCRIPT_NAME'] ?? '/';
```
bobstrecansky, line 401: **Consider — span naming has poor cardinality.** Falling back to `/` and using `SCRIPT_NAME` (typically `/index.php` for every Magento request) means most spans end up named `GET /index.php`. Two suggestions:
- Update the span name to include `URL_PATH` for better differentiation, or
- Update the root span name to `http.route` after `FrontController::dispatch` resolves the route (this is the spec-recommended naming for HTTP server spans, see [HTTP semconv §HTTP server semantics](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name)).

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,361 @@
...
+                $span = Span::fromContext($scope->context());
+                if ($exception) {
+                    $span->recordException($exception);
+                    $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
+                }
+                $span->end();
+            }
+        );
+
+        /** @psalm-suppress DeprecatedClass */
```
bobstrecansky, line 213: **Consider — reuse PSR-15 / shared HTTP server instrumentation.** Nevay's earlier comment asked whether this could share implementation with the existing PSR-15 instrumentation, because the `Http::launch` hook is mostly generic HTTP server bookkeeping (propagator extract, request meta, status code mapping, duration histogram). The same logic is duplicated across Slim, Laravel, Wordpress, and now Magento2.

Not a blocker for this PR, but worth filing a follow-up issue for a shared `OpenTelemetry\Contrib\Instrumentation\Common\PsrHttpServerInstrumentation` — would be a real maintenance win.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,361 @@
...
+        hook(
+            Action::class,
```
bobstrecansky, line 215: **Should fix — add explicit `SpanKind::KIND_INTERNAL` for child spans.** Only `Http::launch` sets `KIND_SERVER`; every other hook in this file implicitly defaults to `KIND_INTERNAL`. The convention in this repo (see `SlimInstrumentation.php`, `WordpressInstrumentation.php`) is to set the span kind explicitly with `->setSpanKind(SpanKind::KIND_INTERNAL)` — it makes intent obvious and avoids any surprise if SDK defaults ever change. Please apply to `FrontController::dispatch`, `Action::dispatch`, `ActionInterface::execute`, `Manager::dispatch`, `InvokerInterface::dispatch`, `Template::fetchView`, `View::renderLayout`, and `Bootstrap::terminate`.

bobstrecansky:
## Review summary — request changes

Thanks for putting this together — adding Magento2 to the contrib monorepo is real work and the test coverage here is thorough. Below is an overall verdict; specific findings are posted as inline comments on the relevant lines.

### Must fix (blockers)
- **PHP version conflict**: `composer.json` requires `^8.3` but CI matrix pins 8.1/8.2 — composer install will fail. (inline on `composer.json:10`, `php.yml:95`)
- **`error.type` misuse**: uses `$exception->getCode()` (almost always `0`); should be `$exception::class`. Also still missing for 5xx. (inline on `Magento2Instrumentation.php:132`)
- **`_OTHER` semantic violation**: only emitted for empty methods; spec says emit for any non-standard method and set `http.request.method_original`. Histogram attrs also need the bucketed value. (inline on `Magento2Instrumentation.php:95`)
- **`Http::launch` double-fires** through Magento's Interceptor pattern (Nevay's earlier flag is unaddressed). Need `if ($class !== Http::class) return;` guard. (inline on `Magento2Instrumentation.php:87`)
- **`open-telemetry/sem-conv: ^1.36`** is incompatible with the repo baseline of `^1.30`. (inline on `composer.json:13`)
- **Missing runtime deps**: `nyholm/psr7`, `nyholm/psr7-server`. (inline on `Magento2Instrumentation.php:21`)
- **`magento/community-edition` in require-dev** is a 200MB+ metapackage; unit tests use mocks and likely don't need it. (inline on `composer.json:30`)

### Should fix
- `Manager::dispatch` and `InvokerInterface::dispatch` post-closures don't accept `?Throwable $exception` — listener errors are silently swallowed. (inline)
- `Bootstrap::terminate` creates a span on every request, not just on exception. (inline)
- `app/etc/vendor_path.php` shouldn't ship in the package; `.gitattributes` doesn't `export-ignore` `app/`. (inline)
- `EVENT:`/`OBSERVER:` span name prefixes diverge from repo convention. (inline)
- Add explicit `SpanKind::KIND_INTERNAL` on child spans. (inline)

### Consider
- Worth filing a follow-up to factor the generic HTTP server bookkeeping into shared code (as Nevay suggested) — the same logic now exists in Slim, Laravel, Wordpress, and Magento2.
- Improve span-name cardinality (`http.route` after FrontController dispatch).

### Risk
**MEDIUM** — purely additive, but CI as written won't pass, the interceptor issue will produce duplicate spans in production, and two HTTP semconv violations affect metric quality.

Happy to re-review once the blockers are addressed.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Sure. Done

.github/workflows/php.yml
jerrytfleung: `magento/community-edition@dev-2.4-develop` requires ~8.3.0 || ~8.4.0 || ~8.5.0.
I put these lines in the matrix so that the action will exclude 8.1 & 8.2 for Instrumentation/Magento2.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Thanks! Added.

src/Instrumentation/Magento2/composer.json
jerrytfleung: `magento/community-edition@dev-2.4-develop` supports 8.3+. So I set the PHP require version to be ^8.3 and also set the exclude 8.1 & 8.2 in php.yml.

src/Instrumentation/Magento2/composer.json
jerrytfleung: Different instrumentation libraries are using different sem-conv version now. I am not sure why it is a blocker. 

^1.38 - https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/Laravel/composer.json#L24
^1.36 - https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/PDO/composer.json#L16, https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/Slim/composer.json#L16, https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/MongoDB/composer.json#L15
^1.32 - https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/Psr15/composer.json#L15, https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/Psr6/composer.json#L15, https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/ReactPHP/composer.json#L16
^1.30 - https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/PostgreSql/composer.json#L24

The prerequisite PR in opentelemetry-php is just the extension required in the `opentelemetry-php-base` for Magento framework. It is already merged and the image is published successfully.

src/Instrumentation/Magento2/composer.json
jerrytfleung: Yes, it is heavy but necessary.
magento/framework is an internal library inside magento/community-edition and not public in packagist.org.
I put @Nevay's suggestion from this slack [message](https://cloud-native.slack.com/archives/C01NFPCV44V/p1777459570295929?thread_ts=1777411768.623579&cid=C01NFPCV44V)

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Sure. Switched to `Nyholm\Psr7\Factory\Psr17Factory`.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: I see. I referred to https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/ReactPHP/src/ReactPHPInstrumentation.php#L289 to fix the `HTTP_REQUEST_METHOD` and added `HTTP_REQUEST_METHOD_ORIGINAL`

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: I see. Fixed the error type in the $responseMeta and also fixed for the 5xx response.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Sure. Fixed the error.type if the status code is 5xx.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Thanks! Added the `Http::launch` guard in the pre/post hook.

In fact, other classes such as FrontController, Action, ActionInterface, Manager... also have this double-fire behavior due to the magento interceptor plugin. From magento's interceptor plugin [limitation](https://developer.adobe.com/commerce/php/development/components/plugins#limitations), it basically intercepts many classes. 

The guard works well for concrete class like `Http::launch`, but it doesn't work well with abstract classes / interface like `Action` and `ActionInterface` because they are not the actual class calling in the framework. 

For those class, as they are internal spans so I am not super worried. I think recommending user to disable interceptor plugin could be a way to handle. 

@bobstrecansky & @Nevay What do you think?

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Sure. Done.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Thanks! I am not aware how to put the third parameter if the function returning `void`. 

Fixed.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Fixed. Thanks!

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: `Boostrap::terminate` will be called on exception in https://github.com/magento/magento2/blob/2.4-develop/lib/internal/Magento/Framework/App/Bootstrap.php#L244 & https://github.com/magento/magento2/blob/2.4-develop/lib/internal/Magento/Framework/App/Bootstrap.php#L275. I also verified there is no `Boostrap::terminate` span found on successful requests. 

Can you help to point me where `Bootstrap::terminate` is called on every request termination in Magento? Thanks!

src/Instrumentation/Magento2/app/etc/vendor_path.php
jerrytfleung: @bobstrecansky I need to install the magento so as to fulfill the composer's dependency requirement.

If I remove the file `app/etc/vendor_path.php` and keep the `app/etc` directory, the `vendor_path.php` will be auto-generated when I run `PROJECT=Instrumentation/Magento2 PHP_VERSION=8.4 make all`

If I remove the file `app/etc/vendor_path.php` and the `app/etc` directory, the composer will complain 
```
In Plugin.php line 323:
                                                                                                                                            
  file_put_contents(/usr/src/myapp/src/Instrumentation/Magento2/app/etc/vendor_path.php): Failed to open stream: No such file or directory 
```
instead.

I removed the file `app/etc/vendor_path.php` and added `app/etc/.gitkeep` to keep the directories.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Sure. Applied the 1st suggestion.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Reusing the logic is great!

I am happy to help but I am not so sure what would a shared `OpenTelemetry\Contrib\Instrumentation\Common\PsrHttpServerInstrumentation` have. 

As I know so far, the instrumentation libraries will be git split for individual composer library, they don't know each other. 

@bobstrecansky @Nevay What do you think?

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Sure. Added.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: I am not sure how to reuse / share those implementation where instrumentation libraries will be finally git split into individual library in packagist.org. 
Can you please help to elaborate more on this point? Thanks!

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Applied.

jerrytfleung:
> ## Review summary — request changes
> Thanks for putting this together — adding Magento2 to the contrib monorepo is real work and the test coverage here is thorough. Below is an overall verdict; specific findings are posted as inline comments on the relevant lines.
> 
> ### Must fix (blockers)
> * **PHP version conflict**: `composer.json` requires `^8.3` but CI matrix pins 8.1/8.2 — composer install will fail. (inline on `composer.json:10`, `php.yml:95`)
> * **`error.type` misuse**: uses `$exception->getCode()` (almost always `0`); should be `$exception::class`. Also still missing for 5xx. (inline on `Magento2Instrumentation.php:132`)
> * **`_OTHER` semantic violation**: only emitted for empty methods; spec says emit for any non-standard method and set `http.request.method_original`. Histogram attrs also need the bucketed value. (inline on `Magento2Instrumentation.php:95`)
> * **`Http::launch` double-fires** through Magento's Interceptor pattern (Nevay's earlier flag is unaddressed). Need `if ($class !== Http::class) return;` guard. (inline on `Magento2Instrumentation.php:87`)
> * **`open-telemetry/sem-conv: ^1.36`** is incompatible with the repo baseline of `^1.30`. (inline on `composer.json:13`)
> * **Missing runtime deps**: `nyholm/psr7`, `nyholm/psr7-server`. (inline on `Magento2Instrumentation.php:21`)
> * **`magento/community-edition` in require-dev** is a 200MB+ metapackage; unit tests use mocks and likely don't need it. (inline on `composer.json:30`)
> 
> ### Should fix
> * `Manager::dispatch` and `InvokerInterface::dispatch` post-closures don't accept `?Throwable $exception` — listener errors are silently swallowed. (inline)
> * `Bootstrap::terminate` creates a span on every request, not just on exception. (inline)
> * `app/etc/vendor_path.php` shouldn't ship in the package; `.gitattributes` doesn't `export-ignore` `app/`. (inline)
> * `EVENT:`/`OBSERVER:` span name prefixes diverge from repo convention. (inline)
> * Add explicit `SpanKind::KIND_INTERNAL` on child spans. (inline)
> 
> ### Consider
> * Worth filing a follow-up to factor the generic HTTP server bookkeeping into shared code (as Nevay suggested) — the same logic now exists in Slim, Laravel, Wordpress, and Magento2.
> * Improve span-name cardinality (`http.route` after FrontController dispatch).
> 
> ### Risk
> **MEDIUM** — purely additive, but CI as written won't pass, the interceptor issue will produce duplicate spans in production, and two HTTP semconv violations affect metric quality.
> 
> Happy to re-review once the blockers are addressed.

Please take another look when you've time. Thanks!

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,573 @@
...
+                $scope = Context::storage()->scope();
+                if (!$scope) {
+                    return;
+                }
+                $scope->detach();
+                $span = Span::fromContext($scope->context());
+
+                $responseMeta = [];
+
+                if ($exception) {
```
bobstrecansky, line 151: If a RuntimeException is thrown and the response ends up with a 503 status, `error.type` will be 503 instead of RuntimeException. The exception class should take priority per semconv.

Fix: Only set error.type from status code when there is no exception:
```
if ($statusCode >= 500 && !$exception) {
    $responseMeta[ErrorAttributes::ERROR_TYPE] = $statusCode;
}
```

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,573 @@
...
+                }
+                $span->end();
+            }
+        );
+
+        hook(
+            Template::class,
+            'fetchView',
+            pre: static function (Template $template, array $params, string $class, string $function, ?string $filename, ?int $lineno) use ($instrumentation) {
+                $filename = is_string($params[0]) ? $params[0] : null;
```
bobstrecansky: nit: we should probably follow the convention in this repo and use a separate filename parameter here?

```
$templateFile = is_string($params[0]) ? $params[0] : null;
$span = $instrumentation->tracer()
    ->spanBuilder('template ' . ($templateFile ?? 'unknown'))
    ...
    ->setAttribute(CodeAttributes::CODE_FILE_PATH, $filename)  // keep original
```

src/Instrumentation/Magento2/src/ResponsePropagationSetter.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        if (null === self::$instance) {
+            self::$instance = new self();
+        }
+
+        return self::$instance;
+    }
+
+    #[Override]
+    public function set(mixed &$carrier, string $key, string $value): void
```
bobstrecansky, line 26: With zend.assertions = -1 (production default), assert() is a no-op. The function signature accepts mixed &$carrier so there's no guard if a non-HttpResponse is passed. A TypeError will be thrown at setHeader() on a non-object, but with an opaque message.

Fix: Use an explicit guard:
```
if (!$carrier instanceof HttpResponse) {
    return;
}
```

bobstrecansky (commented):
A couple small last minute things.  Great work with this!

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: I see. Done.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Thanks! It is a great catch. Fixed.

src/Instrumentation/Magento2/src/ResponsePropagationSetter.php
jerrytfleung: I was following the assert practice from other instrumentation libraries. e.g. https://github.com/open-telemetry/opentelemetry-php-contrib/blob/main/src/Instrumentation/Curl/src/HeadersPropagator.php

But it is always better to use explicit guard.

Done.

bobstrecansky:
This LGTM.  @Nevay - are you good with this?

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
Nevay: The first suggestion violates semconv, URI path must not be used by default https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name
> HTTP span names SHOULD be `{method} {target}` if there is a (low-cardinality) `target` available. 
> Instrumentation MUST NOT default to using URI path as a `{target}`.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,573 @@
...
+                $parent = Globals::propagator()->extract($request->getHeaders());
+
+                $requestMeta = [
+                    HttpAttributes::HTTP_REQUEST_METHOD => self::canonizeMethod($request->getMethod()) ?? '_OTHER',
+                    UrlAttributes::URL_SCHEME => $request->getUri()->getScheme(),
+                    NetworkAttributes::NETWORK_PROTOCOL_VERSION => $request->getProtocolVersion(),
+                ];
+
+                $spanBuilder = $instrumentation->tracer()
+                    ->spanBuilder(sprintf('%s %s', $request->getMethod(), strlen($request->getUri()->getPath()) > 0 ? $request->getUri()->getPath() : self::getScriptNameFromRequest($request)))
```
Nevay: Must use the canonicalized method name, or `HTTP` if `null` https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name
> The {method} MUST be {http.request.method} if the method represents the original method known to the instrumentation. In other cases (when {http.request.method} is set to _OTHER), {method} MUST be HTTP.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,573 @@
...
+                }
+                if ($serverPort !== null) {
+                    $spanBuilder->setAttribute(ServerAttributes::SERVER_PORT, $serverPort);
+                }
+
+                $requestStart = Clock::getDefault()->now();
+                $span = $spanBuilder->setStartTimestamp($requestStart)->startSpan();
+
+                if (strlen($request->getUri()->getQuery()) > 0) {
+                    $span->setAttribute(UrlAttributes::URL_QUERY, $request->getUri()->getQuery());
```
Nevay: Should redact sensitive values, could utilize `open-telemetry/opentelemetry-config-http` `UriSanitizer`.

> `url.query`: Sensitive content provided in `url.query` SHOULD be scrubbed when instrumentations can identify it.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,573 @@
...
+                if (!$scope) {
+                    return;
+                }
+                $scope->detach();
+                $span = Span::fromContext($scope->context());
+
+                $responseMeta = [];
+
+                if ($exception) {
+                    $span->recordException($exception);
```
Nevay: Might want to record exceptions as [`http.server.request.exception` log event](https://opentelemetry.io/docs/specs/semconv/http/http-exceptions/#http-server-request-exception) instead.

https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans/
> Status: [Deprecated](https://opentelemetry.io/docs/specs/otel/document-status/)
Use [Semantic conventions for exceptions in logs](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/) instead.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Sure. Done

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: I see. Applied the `DefaultSanitizer` from the `open-telemetry/opentelemetry-config-http`

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: Sure. Done

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: I see. Applied the 2nd suggestion
I update the root span name after `Action::dispatch` where the route name is resolved.

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,581 @@
...
+                    $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
+                    $responseMeta[ErrorAttributes::ERROR_TYPE] = $exception::class;
+                }
+                if ($response instanceof HttpResponse) {
+                    $statusCode = $response->getStatusCode();
+                    $span->setAttribute(HttpAttributes::HTTP_RESPONSE_STATUS_CODE, $statusCode);
+                    $responseMeta[HttpAttributes::HTTP_RESPONSE_STATUS_CODE] = $statusCode;
+                    if ($statusCode >= 500 && !$exception) {
+                        $responseMeta[ErrorAttributes::ERROR_TYPE] = $statusCode;
+                    }
```
Nevay, line 163:
```suggestion
                    if ($statusCode >= 500 && !$exception) {
                        $span->setStatus(StatusCode::STATUS_ERROR);
                        $responseMeta[ErrorAttributes::ERROR_TYPE] = $statusCode;
                    }
```

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
```diff
@@ -0,0 +1,581 @@
...
+    }
+
+    /**
+     * @param string $method
+     * @return string|null
+     * @psalm-pure
+     */
+    private static function canonizeMethod(string $method): ?string
+    {
+        // RFC9110, RFC5789
```
Nevay: RFC10008 HTTP QUERY was added to the list of known methods https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/#http-request-method

src/Instrumentation/Magento2/composer.json
```diff
@@ -0,0 +1,54 @@
...
+  "description": "OpenTelemetry auto-instrumentation for Magento2",
+  "keywords": ["opentelemetry", "otel", "open-telemetry", "tracing", "magento2", "instrumentation"],
+  "type": "library",
+  "homepage": "https://opentelemetry.io/docs/languages/php/",
+  "readme": "./README.md",
+  "license": "Apache-2.0",
+  "require": {
+    "php": "^8.3",
+    "ext-opentelemetry": "*",
+    "ext-json": "*",
```
Nevay:
```suggestion
    "ext-ctype": "*",
```
Or replace usage of `ctype_digit` with regex/strspn/..

`ext-json` is a core PHP extension -> always enabled.

Nevay (approved):

src/Instrumentation/Magento2/composer.json
jerrytfleung: Done. 
1. Replaced usage of `ctype_digit` with `preg_match('/^\d+$/', $port)`
2. Removed the `"ext-json": "*"` in `composer.json`

src/Instrumentation/Magento2/src/Magento2Instrumentation.php
jerrytfleung: I see. Added QUERY to the list.

bobstrecansky (approved):

PR #581, opened 2026-05-20
added: peer service attribute to Guzzle

paulgrav:
The HttpClientInstrumentation sets the PEER_SERVICE attribute whilst GuzzleInstrumentation.php does not.

Grafana’s Traces Drilldown view uses the PEER_SERVICE attribute to make the spans more readable.

This PR aligns behaviour between the two instrumentations.

### DIFF
--- a/src/Instrumentation/Guzzle/src/GuzzleInstrumentation.php
+++ b/src/Instrumentation/Guzzle/src/GuzzleInstrumentation.php
@@ -54,6 +54,7 @@ class GuzzleInstrumentation
                     ->spanBuilder(sprintf('%s', $request->getMethod()))
                     ->setParent($parentContext)
                     ->setSpanKind(SpanKind::KIND_CLIENT)
+                    ->setAttribute(TraceAttributes::PEER_SERVICE, $request->getUri()->getHost())
                     ->setAttribute(TraceAttributes::URL_FULL, (string) $request->getUri())
                     ->setAttribute(TraceAttributes::HTTP_REQUEST_METHOD, $request->getMethod())
                     ->setAttribute(TraceAttributes::NETWORK_PROTOCOL_VERSION, $request->getProtocolVersion())


bobstrecansky:
@paulgrav - we are actively working on fixing CI - once your draft is ready for review, I'll review it and merge.

src/Instrumentation/Guzzle/src/GuzzleInstrumentation.php
```diff
@@ -54,6 +54,7 @@ public static function register(): void
                     ->spanBuilder(sprintf('%s', $request->getMethod()))
                     ->setParent($parentContext)
                     ->setSpanKind(SpanKind::KIND_CLIENT)
+                    ->setAttribute(TraceAttributes::PEER_SERVICE, $request->getUri()->getHost())
```
andrii-androshchuk, line 57: The schematic convention [says](https://github.com/opentelemetry-php/sem-conv/blob/main/TraceAttributes.php#L3561C8-L3561C179):
> The [`service.name`](/docs/resource/README.md#service) of the remote service. SHOULD be equal to the actual `service.name` resource attribute of the remote service if any.

Are we sure that the `$request->getUri()->getHost()` will return the `service.name` here?

paulgrav: In that case then perhaps neither HttpClientInstrumentation nor GuzzleInstrumentation should set `PEER_SERVICE`?

Only the client of those libraries can know the `PEER_SERVICE` name, but there’s no good way to set it unless you write your own hooks.

For where I work, the host name would make a good proxy for the service.name.

I’d like to know the thinking behind this: [line 65](https://github.com/open-telemetry/opentelemetry-php-contrib/blob/b7978539ac4027e20715282d6ed8f0f50165e50d/src/Instrumentation/Symfony/src/HttpClientInstrumentation.php#L65)

andrii-androshchuk: @paulgrav, I mean, I was expecting this, but at the same time, I don't think having something configured incorrectly elsewhere is a good justification for continuing the same (incorrect) pattern in new places. A much better approach would be to fix the existing incorrect behavior rather than following it.

paulgrav: I can’t disagree. Looking at [HttpClientInstrumentation](https://github.com/paulgrav/opentelemetry-php-contrib/blob/38f145c1e29b406110c94999772def3e0557bf52/src/Instrumentation/Symfony/src/HttpClientInstrumentation.php#L65), the `PEER_SERVICE` attribute should probably be `SERVICE_ADDRESS`.

andrii-androshchuk: Not sure what is `SERVICE_ADDRESS`, but I can only think of this attribute: https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/#network-peer-address

And, if we ALL think that's the right one, it should be updated everywhere.

andrii-androshchuk: Or this:
https://opentelemetry.io/docs/specs/semconv/registry/attributes/server/#server-address

paulgrav: Yep, the last one.

andrii-androshchuk: Indeed, the [server.address](https://opentelemetry.io/docs/specs/semconv/registry/attributes/server/#server-address) looks good for this case. Let's do this.

andrii-androshchuk: @bobstrecansky  / @ChrisLightfootWild, what do you say about the existing instrumentations that use `service.name` whereas `server.address` should have been used. Can we just change it?

@paulgrav provided such an example:
https://github.com/paulgrav/opentelemetry-php-contrib/blob/38f145c1e29b406110c94999772def3e0557bf52/src/Instrumentation/Symfony/src/HttpClientInstrumentation.php#L65

bobstrecansky: The way I'm reading: https://opentelemetry.io/docs/specs/semconv/general/attributes/

is:

service.name and server.address serve two completely different scopes. service.name identifies the logical application emitting the telemetry, while server.address is a network attribute identifying the specific host or IP of a target system being called

bobstrecansky: I'm wondering if the Github history will tell us anything here, or if it was just a mistake.

ChrisLightfootWild: Java appears to have peer service mapping functionality:
 - https://opentelemetry.io/docs/zero-code/java/agent/declarative-configuration/
 - https://github.com/open-telemetry/opentelemetry-java/blob/v1.63.0/api/incubator/src/main/java/io/opentelemetry/api/incubator/config/InstrumentationConfigUtil.java#L21-L32

ChrisLightfootWild: I think `server.address` is the correct way forward. With the lacking functionality to do service mapping ideally being added to in future 🤔

PR #607, opened 2026-06-09
fix(Laravel): fix incorrect URL path construction in `httpTarget()`

zigzagdev:
### Content
- Fixed a PHP operator precedence bug  in `httpTarget()` where `.` (concatenation) has higher precedence than `===` (strict equality). This caused incorrect path construction when appending query strings.
- Ensured correct leading slash handling by replacing `$request->path()` (which strips the leading slash in Laravel) with a robust combination of `$request->getBaseUrl() . $request->getPathInfo()`.
- Added integration tests covering both root and non-root paths, with and without query strings.
  
### Cause

The original code was evaluated incorrectly due to operator precedence:
```php
$question = $request->getBaseUrl() . $request->getPathInfo() === '/' ? '/?' : '?';
return $query ? $request->path() . $question . $query : $request->path();
```

This caused the following incorrect behaviors:
  - GET `/?foo=bar` → URL_PATH = `//?foo=bar` (double slash)
  - GET `/hello?foo=bar` → URL_PATH = `foo?foo=bar` (missing leading slash)

Fixed code:
```php
$path = $request->getBaseUrl() . $request->getPathInfo();

return $query ? $path . '?' . $query : $path;
```

### Test Results
screen shot 2026-06-09 20 13 09

### DIFF
--- a/src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Http/Kernel.php
+++ b/src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Http/Kernel.php
@@ -112,9 +112,9 @@ class Kernel implements LaravelHook
     private function httpTarget(Request $request): string
     {
         $query = $request->getQueryString();
-        $question = $request->getBaseUrl() . $request->getPathInfo() === '/' ? '/?' : '?';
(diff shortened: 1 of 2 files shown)


ChrisLightfootWild (approved):
Thanks for the fix @zigzagdev 👍

PR #608, opened 2026-06-11
Fix monolog handler hidden dependency

aleksandrpanshin-vaimo:
This pull request fix issue described in [#1943](https://github.com/open-telemetry/opentelemetry-php/issues/1943)

### DIFF
--- a/src/Logs/Monolog/composer.json
+++ b/src/Logs/Monolog/composer.json
@@ -11,7 +11,7 @@
   "require": {
     "php": "^8.1",
     "monolog/monolog": "^1.1|^2|^3",
-    "open-telemetry/api": "^1.0"
+    "open-telemetry/api": "^1.8"
   },
   "autoload": {
     "psr-4": {
--- a/src/Logs/Monolog/src/Handler.php
+++ b/src/Logs/Monolog/src/Handler.php
@@ -10,8 +10,6 @@ use Monolog\Formatter\NormalizerFormatter;
 use Monolog\Handler\AbstractProcessingHandler;
 use OpenTelemetry\API\Instrumentation\ConfigurationResolver;
 use OpenTelemetry\API\Logs as API;
-use OpenTelemetry\SDK\Common\Exception\StackTraceFormatter;
-use OpenTelemetry\SemConv\Attributes\ExceptionAttributes;
 
 use Throwable;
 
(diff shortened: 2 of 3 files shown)


ChrisLightfootWild:
Thanks @aleksandrpanshin-vaimo 👍

There's a failure in https://github.com/open-telemetry/opentelemetry-php-contrib/actions/runs/27341511999/job/80787097984 regarding an unused import, if you could fix that?

You can run all the tests against that package locally with:

```bash
make all PHP_VERSION=8.1 PROJECT=Logs/Monolog
```

aleksandrpanshin-vaimo:
Hey @ChrisLightfootWild!

I fixed the issue with styles, although there is still one unrelated error on Monolog package, which is coming from a deprecation notice that `\OpenTelemetry\API\Instrumentation\ConfigurationResolver` should not be used directly, and EnvComponentLoader should be used instead. 

I don't think this should be done in the same pull request, although I'd like to get an advice on how to implement it correctly:

* As far as I understand, I need to create an `EnvComponentLoader` and register it in SPI.
* This will give me an instance of, let's say, `Monolog\InstrumentationConfiguration`.
* Now the question: how do I get access to it from the handler? The only idea I have is to make this instrumentation config static, but I don't like. Probably, I miss some API from the OTel package that is intended exactly for this, can you point me?

bobstrecansky (approved):

ChrisLightfootWild:
> Hey @ChrisLightfootWild!
> 
> I fixed the issue with styles, although there is still one unrelated error on Monolog package, which is coming from a deprecation notice that `\OpenTelemetry\API\Instrumentation\ConfigurationResolver` should not be used directly, and EnvComponentLoader should be used instead.
> 
> I don't think this should be done in the same pull request, although I'd like to get an advice on how to implement it correctly:
> 
> * As far as I understand, I need to create an `EnvComponentLoader` and register it in SPI.
> * This will give me an instance of, let's say, `Monolog\InstrumentationConfiguration`.
> * Now the question: how do I get access to it from the handler? The only idea I have is to make this instrumentation config static, but I don't like. Probably, I miss some API from the OTel package that is intended exactly for this, can you point me?

Sorry for the delay!

I believe there's an example of this here https://github.com/ChrisLightfootWild/opentelemetry-php-contrib/pull/1

--

Once #634 lands, I will split the Monolog instrumentation into its own workflow so we can work towards a green pipeline again :+1:

PR #579, opened 2026-05-15
fix(deps): unblock composer resolution under Psalm 6.16.1

nickmarden:
## Summary

PR #543 bumped `vimeo/psalm` to 6.16.1 across 40 sub-projects but left two surrounding constraints incompatible with the new Psalm, which broke composer resolution and took the entire CI matrix red.

This PR is a draft to validate the fix on real CI. It only touches `composer.json` files (no lock files; they're gitignored here anyway).

### Why CI is currently red on main

- `psalm/plugin-phpunit ^0.20.0` pins `psalm-plugin-api ^0.1`, which conflicts with the `plugin-api ^0.2` that ships with Psalm 6.10+. The latest `plugin-phpunit` release that supports Psalm 6.10+ is `0.19.7`, so this constrains to `^0.19.7` across all sub-projects.
- `vimeo/psalm 6.16.1` requires `netresearch/jsonmapper ^5.0`, but `phan/phan 5.x` transitively pins it to `<=4` via `felixfbecker/advanced-json-rpc 3.2.1`. `phan/phan 6.0.5` drops that dep and accepts `jsonmapper ^5.0`, so this widens `phan/phan` to `^5 || ^6` (or `^5.4 || ^6` where the floor was 5.4).

Verified locally that `composer install --dry-run` resolves cleanly for all 45 sub-projects after these changes.

## Test plan

- [ ] PHP QA matrix turns mostly green (we still expect ~5 residual failures from pre-existing Phan/Psalm code-level issues that PR #497 has been tracking).
- [ ] No new failures versus pre-#543 baseline.

PR #613, opened 2026-06-11
fix: code-level cleanup surfaced during psalm/phan upgrade work

nickmarden:
## Summary

Small set of correctness fixes surfaced while investigating CI failures from the psalm/phan upgrade work (PR #579, now superseded by the revert in #600). These are clean code-level changes with no suppressions beyond one narrow phan annotation.

Each of the three projects touched here is **currently failing on `main`**, and these are the fixes for those failures.

### `Aws` — phan, all PHP versions

`main` reports:

```
src/AwsSdkInstrumentation.php:26 PhanDeprecatedInterface Using a deprecated interface \OpenTelemetry\API\Instrumentation\InstrumentationInterface
src/AwsSdkInstrumentation.php:26 PhanDeprecatedTrait Using a deprecated trait \OpenTelemetry\API\Instrumentation\InstrumentationTrait
```

`AwsSdkInstrumentation` already implemented every interface method itself, so the interface and trait are dropped. The two properties the trait contributed (`$propagator`, `$tracerProvider`) are now declared explicitly, and a constructor installs the same `NoopTextMapPropagator` / `NoopTracerProvider` defaults that the trait's constructor used to set, so an instance stays usable before `setPropagator()` / `setTracerProvider()` are called.

Also adds `@phan-suppress-next-line PhanTypeMismatchArgument` for the `TransportInterface` generic type narrowing on `OtlpExporter` construction in `AwsLambdaWrapper`.

### `Instrumentation/PDO` — phpstan, PHP 8.2 / 8.3 / 8.4

`main` reports the same error on every version:

```
Class Pdo\Sqlite referenced with incorrect case: PDO\Sqlite.
```

The cause is the `use PDO;` import in `PDOInstrumentationTest`: class aliases are case-insensitive, so the relative `@var Pdo\Sqlite` resolves through it to `PDO\Sqlite`. Fully qualifying it as `@var \Pdo\Sqlite` fixes it.

With that resolved, the `Pdo\Sqlite` ignore patterns are all obsolete — PHPStan ships PHP 8.4 stubs whatever version it runs on, so the class and its methods always resolve and the patterns went unmatched (which is itself an error under `reportUnmatchedIgnoredErrors`). Only `PDO::connect()` still needs ignoring below 8.4.

### `ResourceDetectors/Container` — phpunit, all PHP versions

`ContainerTest::test_valid_v1` asserts `ResourceAttributes::SCHEMA_URL` while the detector emits `Version::VERSION_1_38_0->url()`, so it fails on `main`.

An earlier revision of this PR resolved the mismatch by changing the detector. As @ChrisLightfootWild [pointed out in review](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/613#discussion_r3399618164), that silently downgrades the detector's semantic conventions claim from 1.38.0 to 1.32.0 via the deprecated `ResourceAttributes` class. The detector is now left alone and the test is corrected instead, in its own commit. The package already requires `open-telemetry/sem-conv: ^1.38`.

## Test plan

Full `make all-checks` (style, composer validate, phan, psalm, phpstan, phpunit) run locally in the project's Docker images:

| Project | PHP | Result |
| --- | --- | --- |
| `Aws` | 8.1, 8.4 | pass (81 tests) |
| `ResourceDetectors/Container` | 8.1, 8.4 | pass (9 tests) |
| `Instrumentation/PDO` | 8.2, 8.4 | pass (23 tests) |

On 8.4 the `Pdo\Sqlite` subclass test runs for real rather than skipping.

The remaining red jobs in this PR's matrix (`Symfony`, `Yii`, `Psr3`, `Monolog`, `Sampler/RuleBased`, `Laravel`, `ReactPHP`, `HttpConfig`, `CodeIgniter`, `ExtRdKafka`) are pre-existing failures on `main` and are untouched here.

### DIFF
--- a/src/Aws/src/AwsSdkInstrumentation.php
+++ b/src/Aws/src/AwsSdkInstrumentation.php
@@ -9,12 +9,12 @@ use Aws\Middleware;
 use Aws\ResultInterface;
 use Closure;
 use GuzzleHttp\Promise;
-use OpenTelemetry\API\Instrumentation\InstrumentationInterface;
-use OpenTelemetry\API\Instrumentation\InstrumentationTrait;
+use OpenTelemetry\API\Trace\NoopTracerProvider;
 use OpenTelemetry\API\Trace\SpanKind;
 use OpenTelemetry\API\Trace\StatusCode;
 use OpenTelemetry\API\Trace\TracerInterface;
 use OpenTelemetry\API\Trace\TracerProviderInterface;
+use OpenTelemetry\Context\Propagation\NoopTextMapPropagator;
 use OpenTelemetry\Context\Propagation\TextMapPropagatorInterface;
 use Psr\Http\Message\RequestInterface;
 use Stringable;
@@ -23,19 +23,30 @@ use Throwable;
 /**
  * @experimental
  */
-class AwsSdkInstrumentation implements InstrumentationInterface
+class AwsSdkInstrumentation
 {
-    use InstrumentationTrait;
-
     public const NAME = 'AWS SDK Instrumentation';
     public const VERSION = '0.0.1';
     public const SPAN_KIND = SpanKind::KIND_CLIENT;
 
+    private TextMapPropagatorInterface $propagator;
+    private TracerProviderInterface $tracerProvider;
     private array $clients = [];
 
     private array $instrumentedClients = [];
 
     private array $spanStorage = [];
+
+    /**
+     * Defaults match those previously supplied by the removed
+     * InstrumentationTrait constructor, so an instance is usable before
+     * setPropagator()/setTracerProvider() are called.
+     */
+    public function __construct()
+    {
+        $this->propagator = new NoopTextMapPropagator();
+        $this->tracerProvider = new NoopTracerProvider();
+    }
 
     public function getName(): string
     {
--- a/src/Aws/src/Lambda/AwsLambdaWrapper.php
+++ b/src/Aws/src/Lambda/AwsLambdaWrapper.php
@@ -64,6 +64,7 @@ class AwsLambdaWrapper
                 ?: self::DEFAULT_OTLP_EXPORTER_ENDPOINT,
             'application/x-protobuf'
         );
+        /** @phan-suppress-next-line PhanTypeMismatchArgument */
         $exporter = new OtlpExporter($transport);
 
         $spanProcessor = new SimpleSpanProcessor($exporter);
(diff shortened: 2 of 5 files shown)


src/ResourceDetectors/Container/src/Container.php
```diff
@@ -36,7 +35,7 @@ public function getResource(): ResourceInfo
             $attributes[ResourceAttributes::CONTAINER_ID] = $id;
         }
 
-        return ResourceInfo::create(Attributes::create($attributes), Version::VERSION_1_38_0->url());
+        return ResourceInfo::create(Attributes::create($attributes), ResourceAttributes::SCHEMA_URL);
```
ChrisLightfootWild: Prior to this change, the instrumentation claims to emit telemetry adhering to `1.38.0` of the semantic conventions.

Changing to the deprecated `ResourceAttributes` class  downgrades that claim to `1.32.0` (and then it may not be correct for that version).

I think this should be reverted.

ChrisLightfootWild (commented):
@nickmarden thanks for picking some of these changes out from your previous PR.

Happy to move the psalm dependencies forward etc if you can get the working combination that seemed to elude renovate-bot. 😬

src/ResourceDetectors/Container/src/Container.php
nickmarden: Agreed, and reverted. The detector keeps `Version::VERSION_1_38_0->url()`.

For context on why I touched it at all: `ContainerTest::test_valid_v1` asserts `ResourceAttributes::SCHEMA_URL`, so it disagrees with the detector and fails on `main` today, independently of this PR. I had fixed the mismatch from the wrong end. It now fixes the test instead, in a separate commit so it is easy to review on its own.

nickmarden:
@ChrisLightfootWild lemme know if you need anything more on this PR, thanks.

nickmarden:
@ChrisLightfootWild please let me know if you are still interested in this PR, otherwise I will close it. Thanks!

src/Aws/src/AwsSdkInstrumentation.php
ChrisLightfootWild, line 1: I think #705 covered these. Apologies, I appreciate you submitted yours first and it dropped off the radar slightly! 😢

ChrisLightfootWild (commented):
Can you rebase to main and see what's left still? @intuibase has been working through a bunch of errors in order to try and get the green CI again 💪

PR #622, opened 2026-06-18
fix(Laravel): use correct param index for queue name in `hookLater()`

zigzagdev:
### Content
`hookLater()` was passing the wrong argument to `getQueue()` when building the span name.

Laravel's method `later()` signature is:

  ```php
  later($delay, $job, $data = '', $queue = null)
   // $params[0] = delay
   // $params[1] = job
  // $params[2] = data  ← was incorrectly used as queue name
  // $params[3] = queue ← correct
  
  Compare with bulk() where $params[2] is correctly the queue:
  bulk($jobs, $data = '', $queue = null)
  // $params[0] = jobs
  // $params[1] = data
  // $params[2] = queue ← correct
  
`hookLater()` was copied from hookBulk() but the index was not updated, 
causing the job's data payload to be passed to `getQueue()` instead of the actual queue name.

  Fix: change $params[2] to $params[3] in `hookLater()`.
  ```
  
### Test Results
スクリーンショット 2026-06-18 18 56 54


ChrisLightfootWild (approved):

ChrisLightfootWild:
Thanks @zigzagdev 

This also closes https://github.com/opentelemetry-php/contrib-auto-laravel/pull/11 👍

PR #335, opened 2025-03-20
Fix post hook exception possibility in opentelemetry-auto-curl

vsimon:
Fixes open-telemetry/opentelemetry-php/issues/1531

PR #629, opened 2026-06-22
fix(guzzle): end span when Client::transfer() throws synchronously

andreicodrutonofrei-sketch:
## Description

`GuzzleHttp\Client::transfer()` can throw **synchronously** before returning a
promise — e.g. when invalid request options are supplied, since `applyOptions()`
runs *outside* `transfer()`'s try/catch. In that case the extension invokes the
`transfer` post hook with a `null` return value.

The post hook typed the promise as non-nullable (`PromiseInterface $promise`), so
the runtime signature check failed: the hook was skipped, the span leaked (never
ended/detached), and an `E_CORE_WARNING` was emitted:

GuzzleHttp\Client::transfer(): OpenTelemetry: post hook invalid signature,
class=GuzzleHttp\Client function=transfer

This surfaces as a fatal error in error trackers such as Sentry.

## Fix

- Make the post hook's promise parameter nullable (`?PromiseInterface $promise`).
- When the promise is `null` (synchronous throw), end the span — unless it was
  already ended via the exception branch — and return instead of dereferencing
  null.

## Test

Adds a regression test that triggers a synchronous throw via a numerically-indexed
`headers` option and asserts the span is still ended with an error status.

## Related

Same class of "post hook invalid signature / span leak on throw" issue previously
fixed for other instrumentations:
- auto-curl: #335 (issue open-telemetry/opentelemetry-php#1531)
- Symfony Messenger: #267

### DIFF
--- a/src/Instrumentation/Guzzle/tests/Integration/GuzzleInstrumentationTest.php
+++ b/src/Instrumentation/Guzzle/tests/Integration/GuzzleInstrumentationTest.php
@@ -252,6 +252,28 @@ class GuzzleInstrumentationTest extends TestCase
         }
     }
 
+    /**
+     * @link https://github.com/open-telemetry/opentelemetry-php/issues/1988
+     */
+    public function test_post_hook_when_transfer_throws_synchronously(): void
+    {
+        $this->assertCount(0, $this->storage);
+
+        // A numerically-indexed "headers" option makes Client::applyOptions()
+        // throw InvalidArgumentException before transfer() returns a promise.
+        try {
+            $this->client->get('/foo', ['headers' => ['invalid']]);
+            $this->fail('Expected InvalidArgumentException was not thrown');
+        } catch (\InvalidArgumentException $e) {
+            // expected
+        }
+
+        $this->assertCount(1, $this->storage);
+        $span = $this->storage->offsetGet(0);
+        assert($span instanceof ImmutableSpan);
+        $this->assertSame(StatusCode::STATUS_ERROR, $span->getStatus()->getCode());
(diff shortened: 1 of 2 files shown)


ChrisLightfootWild (approved):

src/Instrumentation/Guzzle/tests/Integration/GuzzleInstrumentationTest.php
```diff
@@ -252,6 +252,28 @@ public function test_exceptions_enabled_sets_response_attributes($response, ?int
...
+        // throw InvalidArgumentException before transfer() returns a promise.
+        try {
+            $this->client->get('/foo', ['headers' => ['invalid']]);
+            $this->fail('Expected InvalidArgumentException was not thrown');
+        } catch (\InvalidArgumentException $e) {
+            // expected
+        }
+
+        $this->assertCount(1, $this->storage);
+        $span = $this->storage->offsetGet(0);
```
ChrisLightfootWild, line 277: The tests are failing, possibly around here. Are you able to run them locally to fix?

andreicodrutonofrei-sketch: Hey @ChrisLightfootWild, yes, I will work on fixing them as soon as possible.

andreicodrutonofrei-sketch: can you point me in the direction of what tests should I be focused on? For example the CodeIgniter ones fail for different reasons.

PR #630, opened 2026-06-23
Apply SqlCommenter injection to PDO::prepare as well

Westacular:
In https://github.com/open-telemetry/opentelemetry-php-contrib/pull/442 which added SQLCommenter support for PDO, the SQLCommenter injection was added to hooks for `PDO::query` and `PDO::exec` but oddly not `PDO::prepare`.

This adds the same hook modification to `PDO::prepare` as that PR did for the other two PDO methods, such that prepared queries are also annotated.

While it's possible that a query might be prepared in advance, such that the attributes injected by SQLCommenter at the time of `prepare` do not match the 'correct' attributes for when the query is actually executed, I think in practice it's far more common that `prepare` is called for a query immediately before it is executed. And, really, any time after `prepare` is too late to modify the query anyway, so ... I think a best effort of doing the injection with `prepare` is better than nothing.

### DIFF
--- a/src/Instrumentation/PDO/src/PDOInstrumentation.php
+++ b/src/Instrumentation/PDO/src/PDOInstrumentation.php
@@ -226,8 +226,12 @@ class PDOInstrumentation
                 /** @psalm-suppress ArgumentTypeCoercion */
                 $builder = self::makeBuilder($instrumentation, 'PDO::prepare', $function, $class, $filename, $lineno)
                     ->setSpanKind(SpanKind::KIND_CLIENT);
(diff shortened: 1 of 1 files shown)


jerrytfleung:
@Westacular There is a reason that I didn't add the hook to `PDO::prepare`. 

I found that in mysql scenario, the prepare statement template is compiled and parameters are bind-ed on mysql server side. 
If you inject trace context during `prepare`, the trace context will be compiled on mysql server side and every `execute` function will use the same trace context.

PR #631, opened 2026-06-23
Do not add an extra % when escaping in SQLCommenter

Westacular:
The formatter for SQLCommenter comments was calling `urlencode` to perform percent-escaping, and then adding an _extra_ % to each escaped value with that idea that you need a second % to escape the percent symbols themselves for SQL.

But this isn't necessary, because it's inside a SQL comment (not a literal in a SQL query), and % has no special meaning there.

This was noted in the old Google SQLCommenter repo: https://github.com/google/sqlcommenter/issues/168

... which specifically called out the PHP implementation as out-of-spec.

And the spec also makes it clear that the extra % character is not required: https://google.github.io/sqlcommenter/spec/#key-value-format

I've removed the Util method that was doing the extra % escaping, and switched `formatComments` to just call `urlencode` directly.

### DIFF
--- a/src/SqlCommenter/src/Utils.php
+++ b/src/SqlCommenter/src/Utils.php
@@ -15,22 +15,10 @@ class Utils
         return '/*' . implode(
             ',',
             array_map(
-                static fn (string $value, string $key) => Utils::customUrlEncode($key) . "='" . Utils::customUrlEncode($value) . "'",
+                static fn (string $value, string $key) => urlencode($key) . "='" . urlencode($value) . "'",
                 $comments,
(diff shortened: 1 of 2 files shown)


ChrisLightfootWild:
@Westacular there's some tests that are failing after this change. Can you look to update those too please? :+1:

Westacular:
@ChrisLightfootWild Oh, yes, oops. Done.

ChrisLightfootWild:
@jerrytfleung would you mind taking a look at these changes please?

ChrisLightfootWild (approved):

PR #639, opened 2026-06-26
chore: improve renovate for docker

thompson-tomo:
This improves how renovate handles docker updates as:
- digest updates treated the same as patches ie grouped as opposed to individual arriving anytime
- Docker images pinned to patch version where possible to provide insight on pr’s if it is major, minor etc

in the end this should result in less pr’s while not reducing time to raise other than by a few days in worst case for major/minor docker updates and few hours for digest updates.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -7,7 +7,7 @@
   "packageRules": [
     {
       "groupName": "all patch versions",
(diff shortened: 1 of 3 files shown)


ChrisLightfootWild (approved):
Thanks!

PR #640, opened 2026-06-26
fix(laravel): set STATUS_ERROR for non-zero exit code in `Command::execute`

zigzagdev:
### Motivation

`Console\Kernel::handle` already marks its span as an error when the exit code is non-zero (`$exitCode !== Command::SUCCESS`), but `Illuminate\Console\Command::execute`'s post hook lacked the same guard. 
As a result, when an Artisan command failed with a non-zero exit code, its span stayed at `STATUS_UNSET` instead of `STATUS_ERROR`.

### What I have done

- Added the same `$exitCode !== Command::SUCCESS` check to `Command::execute`'s post hook in `src/Hooks/Illuminate/Console/Command.php`, setting `STATUS_ERROR` on the span when the command fails.
- Added a `FailingCommand` test fixture (`tests/Fixtures/Console/FailingCommand.php`) that always returns `Command::FAILURE`.
- Added `test_failing_command_sets_status_error` to `CommandTest.php`, asserting the span status is `STATUS_ERROR` when the command fails.

### Test Results
```bash
php@9388bd10fb4e:/usr/src/myapp/src/Instrumentation/Laravel$ vendor/bin/phpunit --filter=test_failing_command_sets_status_error tests/Integration/Console/CommandTest.php
PHPUnit 9.6.34 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.34
Configuration: /usr/src/myapp/src/Instrumentation/Laravel/phpunit.xml.dist

.                                                                   1 / 1 (100%)

Time: 00:02.157, Memory: 26.00 MB

OK (1 test, 4 assertions)
```

### DIFF
--- a/src/Instrumentation/Laravel/src/Hooks/Illuminate/Console/Command.php
+++ b/src/Instrumentation/Laravel/src/Hooks/Illuminate/Console/Command.php
@@ -6,6 +6,7 @@ namespace OpenTelemetry\Contrib\Instrumentation\Laravel\Hooks\Illuminate\Console
 
 use Illuminate\Console\Command as IlluminateCommand;
 use OpenTelemetry\API\Trace\Span;
+use OpenTelemetry\API\Trace\StatusCode;
 use OpenTelemetry\Context\Context;
 use OpenTelemetry\Contrib\Instrumentation\Laravel\Hooks\LaravelHook;
 use OpenTelemetry\Contrib\Instrumentation\Laravel\Hooks\LaravelHookTrait;
@@ -55,6 +56,10 @@ class Command implements LaravelHook
                 $span->addEvent('command finished', [
                     'exit-code' => $exitCode,
                 ]);
+
+                if ($exitCode !== IlluminateCommand::SUCCESS) {
+                    $span->setStatus(StatusCode::STATUS_ERROR);
+                }
 
                 $this->endSpan($exception);
             }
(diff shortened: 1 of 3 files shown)


ChrisLightfootWild:
@zigzagdev there are some failures in the pipeline - the test assertion differences might need looking into here, as the expectations do not seem to match.

zigzagdev:
@ChrisLightfootWild 

Thanks for taking a look. 
I checked the pipeline logs closely and don't see any actual assertion mismatches — all tests pass (Tests: 40, Assertions: 1170, 0 failures, 40/40 green). The non-zero exit codes come from two things unrelated to this PR's test expectations:

- Psalm: The only new-looking error is inside `vendor/laravel/framework/.../Command.php`:111 (RedundantCast) — that's Laravel framework's own source, not ours. `laravel/framework` is floated (^10.0 || ^11.0 || ^12.0 || ^13.0), so a newer Laravel patch release likely changed a property type and triggered this in their code.

- PHPUnit: exits 1 due to a runner warning (No filter is configured, code coverage will not be processed) + 2 deprecations, not a failed assertion.

The remaining Psalm findings (QueueTest.php, WorkerTest.php, TestCase.php, etc.) are all in files this PR doesn't touch at all — they're pre-existing on main and unrelated to this PR #640 .
If you're seeing a specific assertion mismatch on your end, let me know which job/line and I'll dig in. 
If these pre-existing Psalm/test issues turn out to need real fixes, I'm happy to open a separate follow-up PR to clean those up rather than bundling it into this one.

bobstrecansky (approved):

PR #645, opened 2026-06-29
feat(metrics): add PHP runtime metrics package

intuibase:
Adds open-telemetry/opentelemetry-metrics-runtime — a new contrib package that auto-registers OTel observable instruments for PHP runtime internals: memory usage/peak/limit, garbage collection stats (including PHP 8.3+ timing fields), OPcache memory and hit-rate, and CPU time via getrusage().

- OPcache and CPU metrics are skipped when underlying functions are unavailable
- Supports PHP 8.1–8.5


src/Metrics/Runtime/src/CpuMetrics.php
```diff
@@ -0,0 +1,64 @@
...
+                'process.runtime.php.cpu.involuntary_context_switches',
+                '{switches}',
+                'Number of times the process was preempted involuntarily',
+            )
+            ->observe(static function (ObserverInterface $observer): void {
+                /** @var array<string, int> $usage */
+                $usage = getrusage();
+                $observer->observe($usage['ru_nivcsw']);
+            });
+    }
```
Nevay: Can use `::batchObserve()` to call `getrusage()` only once per collection cycle, see also
https://github.com/Nevay/otel-instrumentation-runtime-metrics/blob/5d744d71224e1b9a5a22f26e03565eae65c5ba49/src/ProcessMetrics.php#L28-L67

Same applies to the callbacks that use `gc_status()` / `opcache_get_status()`.

src/Metrics/Runtime/src/MemoryMetrics.php
```diff
@@ -0,0 +1,66 @@
...
+                'By',
+                'Memory limit configured in php.ini (-1 means unlimited)',
+            )
+            ->observe(static function (ObserverInterface $observer): void {
+                $limit = ini_get('memory_limit');
+                $observer->observe(self::parseMemoryLimit((string) $limit));
+            });
+    }
+
+    public static function parseMemoryLimit(string $limit): int
```
Nevay, line 59: Should use `ini_parse_quantity()` to handle all valid values; can require `symfony/polyfill-php82` to support PHP 8.1.

src/Metrics/Runtime/src/CpuMetrics.php
```diff
@@ -0,0 +1,64 @@
...
+        $meter
+            ->createObservableCounter(
+                'process.runtime.php.cpu.voluntary_context_switches',
+                '{switches}',
+                'Number of times the process voluntarily yielded the CPU',
+            )
+            ->observe(static function (ObserverInterface $observer): void {
+                /** @var array<string, int> $usage */
+                $usage = getrusage();
+                $observer->observe($usage['ru_nvcsw']);
```
Nevay: `ru_nvcsw` and `ru_nivcsw` might be missing depending on the OS https://www.php.net/manual/en/function.getrusage.php#refsect1-function.getrusage-notes.

src/Metrics/Runtime/src/GarbageCollectionMetrics.php
```diff
@@ -0,0 +1,98 @@
...
+                'Total number of garbage collection cycles run',
+            )
+            ->observe(static function (ObserverInterface $observer): void {
+                $observer->observe(gc_status()['runs']);
+            });
+
+        $meter
+            ->createObservableCounter(
+                'process.runtime.php.gc.collected',
+                '{objects}',
```
Nevay:
```suggestion
                '{object}',
```

Same for all other annotations, https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units:
> All non-units that use curly braces to annotate a quantity need to match the grammatical number of the quantity it represent. For example if measuring the number of individual requests to a process the unit would be `{request}`, not `{requests}`.

src/Metrics/Runtime/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+```shell
+# Disable OPcache and CPU metrics only
+OTEL_PHP_DISABLED_METRICS=opcache,cpu
+
+# Disable GC metrics only
+OTEL_PHP_DISABLED_METRICS=gc
+```
+
+Available group names: `memory`, `gc`, `opcache`, `cpu`.
```
Nevay: Can instead use different meters `io.opentelemetry.contrib.php.runtime.{group}` to allow disabling them using meter configurators or views.

```yaml
meter_provider:
  meter_configurator/development:
    meters:
      - name: io.opentelemetry.contrib.php.runtime.opcache
        config:
          enabled: false
```
```yaml
meter_provider:
  views:
    - selector:
        meter_name: io.opentelemetry.contrib.php.runtime.opcache
      stream: 
        aggregation: 
          drop:

```

src/Metrics/Runtime/composer.json
```diff
@@ -0,0 +1,44 @@
...
+  "keywords": ["opentelemetry", "otel", "open-telemetry", "metrics", "runtime", "php"],
+  "type": "library",
+  "homepage": "https://opentelemetry.io/docs/languages/php/",
+  "readme": "./README.md",
+  "license": "Apache-2.0",
+  "minimum-stability": "dev",
+  "prefer-stable": true,
+  "require": {
+    "php": "^8.1",
+    "open-telemetry/sdk": "^1.0",
```
Nevay:
```suggestion
    "open-telemetry/api": "^1.1",
```

And move sdk dependency to `require-dev` if needed.

src/Metrics/Runtime/_register.php
```diff
@@ -0,0 +1,20 @@
...
+    return;
+}
+
+RuntimeMetrics::register(
+    Globals::meterProvider()->getMeter(
+        RuntimeMetrics::getInstrumentationName(),
+        null,
+        Version::VERSION_1_38_0->url(),
+    )
+);
```
Nevay: This file might be included before the `SdkAutoloader` initializes `Globals`; can instead register the metrics as [`Instrumentation`](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/API/Instrumentation/AutoInstrumentation/Instrumentation.php) to let the SDK load them during `SdkAutoloader::autoload()`.

src/Metrics/Runtime/README.md
```diff
@@ -0,0 +1,113 @@
...
+| `process.runtime.php.opcache.misses` | Counter | `{misses}` | Total cache misses. |
+| `process.runtime.php.opcache.hit_rate` | Gauge | `%` | Cache hit rate percentage. |
+| `process.runtime.php.opcache.cached_scripts` | Gauge | `{scripts}` | Number of scripts currently in cache. |
+| `process.runtime.php.opcache.interned_strings.memory_used` | UpDownCounter | `By` | Memory used by interned strings. |
+| `process.runtime.php.opcache.interned_strings.memory_free` | UpDownCounter | `By` | Free memory in the interned strings buffer. |
+| `process.runtime.php.opcache.interned_strings.count` | Gauge | `{strings}` | Number of interned strings currently stored. |
+
+### CPU (`cpu`)
+
+Registered only on platforms where `getrusage()` is available (Linux, macOS). Not available on Windows.
```
Nevay: `getrusage()` is available on Windows.

intuibase:
Thanks for the thorough review @Nevay! All points addressed.

src/Metrics/Runtime/src/CpuMetrics.php
```diff
@@ -0,0 +1,64 @@
...
+                $usage = getrusage();
+
+                $userSeconds = (float) $usage['ru_utime.tv_sec'] + (float) $usage['ru_utime.tv_usec'] / 1_000_000.0;
+                $systemSeconds = (float) $usage['ru_stime.tv_sec'] + (float) $usage['ru_stime.tv_usec'] / 1_000_000.0;
+
+                $cpuObserver->observe($userSeconds, ['cpu.mode' => 'user']);
+                $cpuObserver->observe($systemSeconds, ['cpu.mode' => 'system']);
+                // ru_nvcsw and ru_nivcsw may be absent on some operating systems (e.g. macOS)
+                $vcsObserver->observe($usage['ru_nvcsw'] ?? 0);
+                $ivcsObserver->observe($usage['ru_nivcsw'] ?? 0);
```
Nevay: Should not record these values if they are not available.

src/Metrics/Runtime/src/CpuMetrics.php
```diff
@@ -0,0 +1,64 @@
...
+            'process.runtime.php.cpu.voluntary_context_switches',
+            '{switch}',
+            'Number of times the process voluntarily yielded the CPU',
+        );
+
+        $involuntaryCS = $meter->createObservableCounter(
+            'process.runtime.php.cpu.involuntary_context_switches',
+            '{switch}',
+            'Number of times the process was preempted involuntarily',
+        );
```
Nevay: [`process.context_switches`](https://opentelemetry.io/docs/specs/semconv/system/process-metrics/#metric-processcontext_switches)

src/Metrics/Runtime/src/RuntimeMetrics.php
```diff
@@ -0,0 +1,74 @@
...
+
+    public static function getInstrumentationName(): string
+    {
+        return self::INSTRUMENTATION_NAME;
+    }
+
+    /** @return list<string> */
+    private static function disabledGroups(): array
+    {
+        $value = getenv(self::ENV_DISABLED_METRICS);
```
Nevay: If we want to keep the option to disable them via env: should IMO use the existing `OTEL_PHP_DISABLED_INSTRUMENTATIONS` env variable with `metrics-runtime-memory`/`metrics-runtime-gc`/.. instead of adding a new env variable.

---

As this is using the `Instrumentation` registration mechanism: can use `EnvComponentLoader` to load an `InstrumentationConfiguration` instead of using `getenv()` directly ([example](https://github.com/Nevay/otel-instrumentation-amphp-http-client/blob/main/src/ConfigEnv/InstrumentationConfigurationAmphpHttpClientConfig.php)). This ensures that:
1. values from `.env` files are available
2. we do not use the env-based configuration if file-based configuration is used

> When `OTEL_CONFIG_FILE` is set, all other environment variables besides those referenced in the configuration file for environment variable substitution MUST be ignored. Ignoring the environment variables is necessary because there is no intuitive way to merge the flat environment variable scheme with the structured file configuration scheme in all cases.

src/Metrics/Runtime/src/MemoryMetrics.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $meter
+            ->createObservableUpDownCounter(
+                'process.runtime.php.memory.usage',
+                'By',
+                'Current memory usage (real allocation from OS)',
+            )
+            ->observe(static function (ObserverInterface $observer): void {
+                $observer->observe(memory_get_usage(true), ['memory.type' => 'real']);
+                $observer->observe(memory_get_usage(false), ['memory.type' => 'emalloc']);
```
Nevay: Should record `memory_get_usage()` and `memory_get_usage(true) - memory_get_usage()` to still produce useful values if the `memory.type` dimension is dropped.


```yaml
meter_provider:
  views:
    - selector:
        instrument_name: process.runtime.php.memory.usage
      stream:
        attribute_keys:
          excluded:
            - memory.type
```

https://opentelemetry.io/docs/specs/semconv/general/metrics/#general-guidelines
> “As a rule of thumb, aggregations over all the attributes of a given metric SHOULD be meaningful,”

src/Metrics/Runtime/composer.json
```diff
@@ -0,0 +1,50 @@
...
+      "php-http/discovery": false,
+      "tbachert/spi": true
+    }
+  },
+  "extra": {
+    "spi": {
+      "OpenTelemetry\\API\\Instrumentation\\AutoInstrumentation\\Instrumentation": [
+        "OpenTelemetry\\Contrib\\Metrics\\Runtime\\RuntimeMetricsInstrumentation"
+      ]
+    }
```
ChrisLightfootWild: I think it might be worth keeping the `_register.php` to register instrumentation, and copying the [pattern here](https://github.com/Nevay/otel-instrumentation-runtime-metrics/blob/main/composer.json#L37-L39) which allows the removal of the file if the composer plugin runs to pre-compute the service discovery.

This allows the instrumentation to still work in the event that the plugin has not been allowed to run by the developer.

src/Metrics/Runtime/src/CpuMetrics.php
```diff
@@ -0,0 +1,64 @@
...
+class CpuMetrics
+{
+    public const GROUP = 'cpu';
+
+    public static function isAvailable(): bool
+    {
+        return function_exists('getrusage');
+    }
+
+    public static function register(MeterInterface $meter): void
```
Nevay, line 22: Mark class `internal` or accept `MeterProviderInterface` instead of `MeterInterface`

ChrisLightfootWild (commented):
Awesome stuff.

Could we add a `.gitattributes` also to this package? 💪

src/Metrics/Runtime/src/GarbageCollectionMetrics.php
```diff
@@ -0,0 +1,95 @@
...
+use OpenTelemetry\API\Metrics\ObserverInterface;
+
+class GarbageCollectionMetrics
+{
+    public const GROUP = 'gc';
+
+    public static function register(MeterInterface $meter): void
+    {
+        $runs = $meter->createObservableCounter(
+            'process.runtime.php.gc.runs',
```
Nevay: Should drop the `process.runtime.` prefix, see other language specific metrics https://opentelemetry.io/docs/specs/semconv/runtime/#metrics.

src/Metrics/Runtime/src/RuntimeMetrics.php
```diff
@@ -0,0 +1,74 @@
...
+     * functions are unavailable, regardless of this setting.
+     */
+    public static function register(MeterProviderInterface $meterProvider): void
+    {
+        $schemaUrl = Version::VERSION_1_38_0->url();
+        $disabled = self::disabledGroups();
+
+        if (!\in_array(MemoryMetrics::GROUP, $disabled, true)) {
+            MemoryMetrics::register(
+                $meterProvider->getMeter(self::INSTRUMENTATION_NAME . '.memory', null, $schemaUrl),
```
Nevay: Should specify instrumentation version.
```suggestion
                $meterProvider->getMeter(self::INSTRUMENTATION_NAME . '.memory', InstalledVersions::getPrettyVersion('open-telemetry/opentelemetry-metrics-runtime'), $schemaUrl),
```

src/Metrics/Runtime/src/MemoryMetrics.php
```diff
@@ -0,0 +1,59 @@
...
+        $meter
+            ->createObservableUpDownCounter(
+                'php.memory.usage',
+                'By',
+                'Current memory usage (real allocation from OS)',
+            )
+            ->observe(static function (ObserverInterface $observer): void {
+                $emalloc = memory_get_usage(false);
+                $observer->observe($emalloc, ['memory.type' => 'emalloc']);
+                $observer->observe(memory_get_usage(true) - $emalloc, ['memory.type' => 'overhead']);
```
Nevay: Should call `memory_get_usage(true)` immediately after `memory_get_usage(false)`, calling `$observer->observe()` might allocate additional memory.

src/Metrics/Runtime/src/MemoryMetrics.php
```diff
@@ -0,0 +1,59 @@
...
+            });
+
+        $meter
+            ->createObservableGauge(
+                'php.memory.limit',
+                'By',
+                'Memory limit configured in php.ini (-1 means unlimited)',
+            )
+            ->observe(static function (ObserverInterface $observer): void {
+                $limit = ini_get('memory_limit') ?: '-1';
```
Nevay: I'm learning towards not reporting this value if unlimited to keep avg/min/... aggregations useful.

src/Metrics/Runtime/src/RuntimeMetricsInstrumentation.php
```diff
@@ -13,6 +13,9 @@ class RuntimeMetricsInstrumentation implements Instrumentation
 {
     public function register(HookManagerInterface $hookManager, ConfigProperties $configuration, Context $context): void
     {
-        RuntimeMetrics::register($context->meterProvider);
+        $config = $configuration->get(RuntimeMetricsConfig::class);
+        $disabled = $config instanceof RuntimeMetricsConfig ? $config->disabled : [];
+
+        RuntimeMetrics::register($context->meterProvider, $disabled);
```
ChrisLightfootWild: Based on [a PR](https://github.com/ChrisLightfootWild/opentelemetry-php-contrib/pull/1) from @Nevay previously, perhaps build a config object and pass that in.

It should be easier in future to access existing config that way.

```suggestion
$config = $configuration->get(RuntimeMetricsConfig::class) ?? new RuntimeMetricsConfig();

RuntimeMetrics::register($context->meterProvider, $config);
```

ChrisLightfootWild: @intuibase sorry, I should have noted that `RuntimeMetrics::register()` will need to accept `RuntimeMetricsConfig` also, and then reference the `->disabled` instrumentations where they are needed there.

src/Metrics/Runtime/_register.php
```diff
@@ -0,0 +1,14 @@
...
+declare(strict_types=1);
+
+use Nevay\SPI\ServiceLoader;
+use OpenTelemetry\API\Configuration\ConfigEnv\EnvComponentLoader;
+use OpenTelemetry\API\Instrumentation\AutoInstrumentation\Instrumentation;
+use OpenTelemetry\Contrib\Metrics\Runtime\InstrumentationConfigurationRuntimeMetricsConfig;
+use OpenTelemetry\Contrib\Metrics\Runtime\RuntimeMetricsInstrumentation;
+
+// @phan-suppress-next-line PhanTemplateTypeConstraintViolation
+ServiceLoader::register(Instrumentation::class, RuntimeMetricsInstrumentation::class);
```
ChrisLightfootWild, line 12: If an older SDK is used (pre-`1.1.0`), this dependency may not exist. So we should probably promote it to the `requires` in this package as an immediate dependency.

ChrisLightfootWild (approved):
Looking great! Sorry, I left one final thought on here.

It can always be added as a follow-up PR though, especially as we are only going to be tagging an initial `0.1` release etc.

src/Metrics/Runtime/_register.php
Nevay:
```php
if (!class_exists(ServiceLoader::class)) {
    return;
}
```

src/Metrics/Runtime/src/InstrumentationConfigurationRuntimeMetricsConfig.php
```diff
@@ -0,0 +1,54 @@
...
+        GarbageCollectionMetrics::GROUP,
+        OpcacheMetrics::GROUP,
+        CpuMetrics::GROUP,
+    ];
+
+    public function load(EnvResolver $env, EnvComponentLoaderRegistry $registry, Context $context): InstrumentationConfiguration
+    {
+        $disabledInstrumentations = $env->list('OTEL_PHP_DISABLED_INSTRUMENTATIONS') ?? [];
+
+        if (in_array('all', $disabledInstrumentations, true) || in_array(RuntimeMetrics::NAME, $disabledInstrumentations, true)) {
```
Nevay:
```suggestion
        if (disabledInstrumentations === ['all'] || in_array(RuntimeMetrics::NAME, $disabledInstrumentations, true)) {
```
To match current semantics of `Sdk::isInstrumentationDisabled()`.

Nevay (approved):

src/Metrics/Runtime/_register.php
intuibase: I've added tbachert/spi instead of SDK.

ChrisLightfootWild:
https://github.com/opentelemetry-php/contrib-metrics-runtime/releases/tag/0.1.0

PR #650, opened 2026-07-07
fix(laravel): don't let Kernel::handle hook throw on hostile method override

hendrikheil:
## Which problem is this PR solving?

`Illuminate\Http\Request::method()`/`getMethod()` throws a `SuspiciousOperationException` when the request carries a malformed HTTP method override — e.g. `_method=__construct` or a `X-HTTP-METHOD-OVERRIDE: __construct` header, which is a fairly common vulnerability-scanner probe. `getHost()` has the same failure mode for a `Host` header that fails validation, and it's reached both directly (`host()`/`getHost()`) and indirectly via `fullUrl()` (used for the `URL_FULL` attribute).

The `Kernel::handle` `pre` hook calls all of these unguarded while building the span. Because this is a `pre` hook, it runs *before* the intercepted `Kernel::handle()` method body executes — i.e. before Laravel's own `try { ... } catch (Throwable $e)` around request handling is even reached. An uncaught throw here isn't caught by the application at all; the extension reports it as a raw PHP warning outside the normal response lifecycle (`OpenTelemetry: pre hook threw exception, ... message=Invalid HTTP method override.`), and the span for the request is silently dropped.

In a real deployment this surfaced as a `FatalError: Cannot modify header information - headers already sent` — the warning text got written to the response body before any headers were sent, so a later attempt to set a header (e.g. by the framework's own exception rendering) fatals.

## Short description of the changes

- Extract `$request->method()` access into a new `httpMethod()` helper that catches `Throwable` and falls back to `'unknown'`, used in both the `pre` and `post` hooks (also removes a redundant second call to `method()` that existed in the `pre` hook).
- Add a `httpFullUrl()` helper guarding `$request->fullUrl()` the same way, since it independently calls `getHost()` and runs *before* `httpHostName()` in the attribute chain — so a hostile `Host` header could still crash the hook even with `httpHostName()` guarded.
- Guard the one reachable `getHost()` call site inside `httpHostName()`. The `getHost()` fallback branch there is dead code (`Illuminate\Http\Request` always defines `host()`) and was already uncovered on `main`, so it's intentionally left unwrapped/untouched rather than dragging pre-existing untested lines into this diff.
- Add regression tests: a malformed `X-HTTP-METHOD-OVERRIDE` header case and a malformed `Host` header case (the latter constructs the request manually, since `MakesHttpRequests::call()`'s `Request::create()` always overwrites `HTTP_HOST` from the parsed URI). Plus focused unit tests invoking `httpMethod()`/`httpFullUrl()`/`httpHostName()` directly via reflection so each guard's catch branch is exercised independent of call order.

Reverting the `Kernel.php` guards locally reproduces the exact warning and the dropped span for both the method-override and Host-header cases, confirming this is the same failure mode in both spots.

### DIFF
--- a/src/Instrumentation/Laravel/tests/Unit/Hooks/Illuminate/Contracts/Http/KernelTest.php
+++ b/src/Instrumentation/Laravel/tests/Unit/Hooks/Illuminate/Contracts/Http/KernelTest.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Tests\Contrib\Instrumentation\Laravel\Unit\Hooks\Illuminate\Contracts\Http;
+
+use Illuminate\Http\Request;
+use OpenTelemetry\Contrib\Instrumentation\Laravel\Hooks\Illuminate\Contracts\Http\Kernel;
+use PHPUnit\Framework\TestCase;
+use ReflectionClass;
+use ReflectionMethod;
+use Symfony\Component\HttpFoundation\Exception\SuspiciousOperationException;
+
+class KernelTest extends TestCase
+{
+    public function test_http_method_swallows_suspicious_operation_exception(): void
+    {
+        $request = new class() extends Request {
+            public function method(): string
+            {
+                throw new SuspiciousOperationException('Invalid method override.');
+            }
+        };
+
+        $this->assertSame('unknown', $this->invokeGuard('httpMethod', $request));
+    }
+
+    public function test_http_full_url_swallows_suspicious_operation_exception(): void
+    {
+        $request = new class() extends Request {
+            public function fullUrl(): string
+            {
+                throw new SuspiciousOperationException('Invalid Host.');
+            }
+        };
+
+        $this->assertSame('', $this->invokeGuard('httpFullUrl', $request));
+    }
+
+    public function test_http_host_name_swallows_suspicious_operation_exception(): void
+    {
+        $request = new class() extends Request {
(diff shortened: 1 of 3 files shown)


src/Instrumentation/Laravel/tests/Unit/Hooks/Illuminate/Contracts/Http/KernelTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+        $this->assertSame('', $this->invokeGuard('httpHostName', $request));
+    }
+
+    private function invokeGuard(string $method, Request $request): string
+    {
+        $kernel = (new ReflectionClass(Kernel::class))->newInstanceWithoutConstructor();
+
+        $reflectionMethod = new ReflectionMethod(Kernel::class, $method);
+        $reflectionMethod->setAccessible(true);
```
PuvaanRaaj, line 57: This line currently fails the Laravel quality matrix with Psalm UnusedMethodCall. On the supported PHP 8.1+ runtimes, reflection members can be invoked without calling setAccessible(), so please remove this call (or otherwise adjust the helper) to restore the required quality checks.

PuvaanRaaj (commented):
Reviewed the current Laravel hardening diff and the relevant CI failures. The guarded request accessors and regression coverage are directionally useful, but the branch still has a directly introduced quality-check failure.

ChrisLightfootWild (approved):
Thanks @hendrikheil 👍

PR #652, opened 2026-07-08
feat(symfony): emit CONSUMER span when a received Messenger message is dispatched

### LINKED ISSUE #1314: [Symfony auto-instrumentation] Messenger instrumentation should use SpanKind::KIND_PRODUCER, SpanKind::KIND_CONSUMER
**Steps to reproduce**

Enable Symfony auto-instrumentation with default setup.

Spans when creating the message should be marked with `SpanKind::KIND_PRODUCER`.

Spans when **handling** the message should be marked with `SpanKind::KIND_CONSUMER`.

https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-kind

Everything is marked `KIND_INTERNAL` and it seems the consumer is currently not instrumented at all (I might be wrong here, please feel free to correct me).

Additionally, it seems to me the names of the spans are also not correct, as shown in the table in the spec linked above?

PuvaanRaaj:
## Problem

Fixes open-telemetry/opentelemetry-php#1314.

The Symfony Messenger instrumentation only emitted `KIND_PRODUCER` spans (`DISPATCH` / `SEND`). The **consumer** side of messaging was not represented at all — as noted in the issue, "the consumer is currently not instrumented" and spans were not marked `KIND_CONSUMER` per the [messaging span-kind conventions](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#span-kind).

## Fix

When a Symfony worker pulls a message off a transport and re-dispatches it onto the bus, the `Envelope` carries a `ReceivedStamp`. The `MessageBusInterface::dispatch` pre-hook now:

- detects the `ReceivedStamp` and emits a `KIND_CONSUMER` span named `CONSUME <message>` (instead of the producer `DISPATCH` span); and
- for an `Envelope`, reports the **wrapped** message class as `symfony.messenger.message` and in the span name, rather than `Symfony\...\Envelope`.

Plain (producer-side) dispatches are unchanged — still `DISPATCH` / `KIND_PRODUCER`.

This is intentionally scoped to span kind + naming; trace-context propagation through transports remains out of scope (as the class docblock already notes).

## Tests

Added `test_dispatch_of_received_message_is_a_consumer_span`: dispatching an `Envelope` stamped with `ReceivedStamp` produces a `CONSUME …` / `KIND_CONSUMER` span with the unwrapped message class. Fails before this change, passes after. Existing producer tests remain green (22 tests).

> Note: `psalm`/`phpstan` report 2 pre-existing errors in `SymfonyInstrumentation.php` / `SymfonyInstrumentationTest.php` (a missing `TraceResponsePropagator` due to local Symfony version skew) that also occur on a clean checkout and are unrelated to this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Instrumentation/Symfony/src/MessengerInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/MessengerInstrumentation.php
@@ -14,6 +14,7 @@ use OpenTelemetry\SemConv\TraceAttributes;
 use OpenTelemetry\SemConv\Version;
 use Symfony\Component\Messenger\Envelope;
 use Symfony\Component\Messenger\MessageBusInterface;
+use Symfony\Component\Messenger\Stamp\ReceivedStamp;
 use Symfony\Component\Messenger\Transport\Sender\SenderInterface;
 
 /**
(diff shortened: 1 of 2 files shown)


PuvaanRaaj:
@open-telemetry/php-approvers @brettmc adds the missing KIND_CONSUMER span for Symfony Messenger (#1314) — worker-received messages (ReceivedStamp) now emit a CONSUME/consumer span instead of a producer DISPATCH. Test included. Review welcome. 🙏

bobstrecansky (approved):

PR #653, opened 2026-07-08
Add `process.paging.faults` and `process.uptime` runtime metrics

Nevay:
Adds additional recommended process metrics:
- [`process.paging.faults`](https://opentelemetry.io/docs/specs/semconv/system/process-metrics/#metric-processpagingfaults)
- [`process.uptime`](https://opentelemetry.io/docs/specs/semconv/system/process-metrics/#metric-processuptime)

Follow-up to #645, cc @intuibase

### DIFF
--- a/src/Metrics/Runtime/README.md
+++ b/src/Metrics/Runtime/README.md
@@ -95,7 +95,7 @@ Registered only on platforms where `getrusage()` is available (Linux, macOS, Win
 |--------|------|------|-------------|
 | `process.cpu.time` | Counter | `s` | CPU time consumed. Reported for `user` and `system` modes via the `cpu.mode` attribute. |
 | `process.context_switches` | Counter | `{context_switch}` | Number of times the process has been context switched. Reported for `voluntary` and `involuntary` switches via the `process.context_switch.type` attribute. |
-| `process.paging.faults` | Counter | `{fault}` | Number of page faults the process has made. Reported for `minor` and `major` faults via the `system.paging.fault.ype` attribute. |
+| `process.paging.faults` | Counter | `{fault}` | Number of page faults the process has made. Reported for `minor` and `major` faults via the `system.paging.fault.type` attribute. |
 
 ## Configuration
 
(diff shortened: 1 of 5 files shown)


src/Metrics/Runtime/src/GarbageCollectionMetrics.php
```diff
@@ -36,63 +37,66 @@ public static function register(MeterInterface $meter): void
...
+            's',
+            'Cumulative time spent running destructors during GC',
+        );
+        $freeTime = $meter->createObservableCounter(
+            'php.gc.free_time',
+            's',
+            'Cumulative time spent freeing memory during GC',
+        );
+        $processUptime = $meter->createObservableGauge(
+            'process.uptime',
```
Nevay, line 57:
```suggestion
            'php.uptime',
```

intuibase: I'm not sure what the specification suggests here, but uptime seems to be more process-related than technology-related.

Do you think it might be useful to keep it for the FPM and Apache SAPIs as well, so that we can collect metrics on worker lifetime?

If you decide to go with php.uptime, don't forget to update it in README.md.

ChrisLightfootWild: @Nevay any thoughts on this? :+1:

Nevay: > keep it for the FPM and Apache SAPIs as well, so that we can collect metrics on worker lifetime?

`REQUEST_TIME_FLOAT`/`gc_status()['application_time']` are reset on each request, we cannot use it to track the lifetime of the underlying worker process.

Uptime seems mostly relevant for long-running worker processes -> keeping it as CLI-only `process.uptime`.

src/Metrics/Runtime/README.md
```diff
@@ -68,6 +68,7 @@ RuntimeMetrics::register($meterProvider, new RuntimeMetricsConfig(disabled: ['op
 | `php.gc.collector_time` | Counter | `s` | Cumulative time spent in the GC collector. **PHP 8.3+** |
 | `php.gc.destructor_time` | Counter | `s` | Cumulative time spent running destructors during GC. **PHP 8.3+** |
 | `php.gc.free_time` | Counter | `s` | Cumulative time spent freeing memory during GC. **PHP 8.3+** |
+| `process.uptime` | Gauge | `s` | The time the process has been running. **PHP 8.3+** |
```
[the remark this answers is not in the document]
Nevay: Added note that `process.uptime` is only observed for CLI SAPI.

PR #655, opened 2026-07-11
fix(curl): deduplicate propagation headers on outgoing request

### LINKED ISSUE #1898: Duplicate propagation headers when auto-guzzle and auto-curl are both enabled
**Describe your environment**

- PHP 8.4
- `open-telemetry/opentelemetry-auto-guzzle` 1.2.1
- `open-telemetry/opentelemetry-auto-curl` 0.0.4
- `guzzlehttp/guzzle` 7.10.0 (default cURL handler)

**Steps to reproduce**

Install both `opentelemetry-auto-guzzle` and `opentelemetry-auto-curl`. Make any Guzzle HTTP request using the default handler stack and capture the outgoing request headers.

A single set of propagation headers (traceparent, tracestate, etc.) on the outgoing request.

All propagation headers appear twice. It looks like the following is happening:

1. The Guzzle hook fires on `Client::transfer()` and calls `inject()` on the PSR-7 `RequestInterface`, adding propagation headers with span A's context
2. `CurlFactory::applyHeaders()` copies those headers into `CURLOPT_HTTPHEADER`
3. The cURL hook fires on `curl_exec()` and calls `inject()` on `CurlHandleMetadata`, adding propagation headers again with span B's context
4. `getRequestHeadersToSend()` merges `headersToPropagate` with existing `headers` via `array_merge()`, which appears to produce duplicates of each propagation header
5. Both sets seem to be sent on the wire

In our case, we had a third-party API reject requests outright because of the duplicate headers.

**Additional context**
The same behavior may apply with `auto-psr18` + `auto-curl` since PSR-18 also injects into `RequestInterface` before the request reaches cURL.

Since the cURL span is the innermost span (closest to the wire), it might make sense for `getRequestHeadersToSend()` to strip existing propagation field keys from `$this->headers` before merging in `$this->headersToPropagate`, so the cURL instrumentation's context takes precedence. `propagator->fields()` already provides the list of keys that could be matched against.

We are also working on addressing leakage of the propagation headers to external entities. While that may solve most of our problems, this seems like a real issue that may cause undefined behaviours in some systems. We plan to use a custom propagator that decorates the configured propagators, but if you have any advice on better options that would be much appreciated.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PuvaanRaaj:
## Problem

With both `open-telemetry/opentelemetry-auto-guzzle` and `open-telemetry/opentelemetry-auto-curl` enabled, every outgoing Guzzle request (which uses cURL as its default handler) ends up carrying `traceparent`/`tracestate` **twice** on the wire. Some third-party APIs reject requests with duplicate propagation headers.

## Root cause

`src/Instrumentation/Curl/src/CurlHandleMetadata.php`, `getRequestHeadersToSend()` (line 57 on `main`):

```php
$headers = array_merge($this->headersToPropagate, $this->headers);
```

Sequence of events:
1. The Guzzle hook fires on `Client::transfer()` and calls `inject()` on the PSR-7 request, adding `traceparent`/`tracestate` to the request headers.
2. Guzzle's `CurlFactory::applyHeaders()` copies those headers into `CURLOPT_HTTPHEADER`. The curl instrumentation's `updateFromCurlOption()` picks this list up and stores it in `CurlHandleMetadata::$headers`.
3. The curl hook fires on `curl_exec()` and calls `inject()` again, this time on `CurlHandleMetadata` via `HeadersPropagator`, which appends a **second** set of propagation headers into `$headersToPropagate`.
4. `getRequestHeadersToSend()` merges `$headersToPropagate` and `$headers` with `array_merge()`. Since `CURLOPT_HTTPHEADER` is a plain numerically-indexed list of `"Name: value"` strings (not an associative map keyed by header name), `array_merge()` just concatenates the two lists — both `traceparent` entries survive and both go out with `CURLOPT_HTTPHEADER`.

Same class of bug applies to `auto-psr18` + `auto-curl` (any instrumentation that injects into the PSR-7/PSR-18 request before curl's own hook runs).

## Fix

Deduplicate the final header list by header name (case-insensitive) in `getRequestHeadersToSend()`, building an associative map keyed by lowercased header name from `$this->headers` first, then overlaying `$this->headersToPropagate` (so the freshly-injected, most-recent propagation context wins for any header name collision), then flattening back to the `"Name: value"` list form cURL expects. Non-propagation headers and their original casing/values are left untouched.

Invariant satisfied: exactly one `traceparent`/`tracestate`/`baggage` header reaches the wire, regardless of how many instrumentations injected into the request beforehand.

## Tests

Added `src/Instrumentation/Curl/tests/Unit/CurlHandleMetadataTest.php`:
- `test_get_request_headers_to_send_deduplicates_already_propagated_header`: seeds `CURLOPT_HTTPHEADER` with a `traceparent`/`tracestate` pair (simulating an upstream Guzzle/PSR-18 injection), then calls `setHeaderToPropagate()` again for the same header names (simulating curl's own inject), and asserts the resulting list has exactly one `traceparent` and one `tracestate`, with the freshly-injected value, plus the untouched `Content-Type` header.
- `test_get_request_headers_to_send_returns_null_when_nothing_to_propagate`: unchanged existing behavior.

**Red/green check**: reverted `CurlHandleMetadata.php` to the `array_merge()` version on `main` and confirmed the new test fails (`Failed asserting that actual size 5 matches expected size 3`, i.e. duplicate headers present). Restored the fix and confirmed it passes.

## Verification (Docker, `PHP_VERSION=8.2`)

- `make test` (Instrumentation/Curl): 19 tests, 131 assertions, all green (existing integration tests unaffected).
- `make style` (php-cs-fixer): clean, 0 files changed.
- `make psalm`: no errors.
- `make phpstan`: no errors.

No public API changes.

Fixes open-telemetry/opentelemetry-php#1898

### DIFF
--- a/src/Instrumentation/Curl/src/CurlHandleMetadata.php
+++ b/src/Instrumentation/Curl/src/CurlHandleMetadata.php
@@ -54,10 +54,22 @@ class CurlHandleMetadata
         if (count($this->headersToPropagate) == 0) {
             return null;
         }
-        $headers = array_merge($this->headersToPropagate, $this->headers);
+
+        // Deduplicate by header name (case-insensitive), so that a header already present on the
+        // handle (e.g. injected upstream by another instrumentation such as auto-guzzle/auto-psr18)
+        // does not end up duplicated alongside the value we're about to inject here.
+        $headers = [];
+        foreach ($this->headers as $header) {
+            $name = strtolower(trim(explode(':', $header, 2)[0]));
+            $headers[$name] = $header;
+        }
+        foreach ($this->headersToPropagate as $header) {
+            $name = strtolower(trim(explode(':', $header, 2)[0]));
+            $headers[$name] = $header;
+        }
         $this->headersToPropagate = [];
 
-        return $headers;
+        return array_values($headers);
     }
 
     public function getCapturedResponseHeaders(): array
--- a/src/Instrumentation/Curl/src/CurlHandleMetadata.php
+++ b/src/Instrumentation/Curl/src/CurlHandleMetadata.php
@@ -55,21 +55,21 @@ class CurlHandleMetadata
             return null;
         }
 
-        // Deduplicate by header name (case-insensitive), so that a header already present on the
-        // handle (e.g. injected upstream by another instrumentation such as auto-guzzle/auto-psr18)
-        // does not end up duplicated alongside the value we're about to inject here.
-        $headers = [];
-        foreach ($this->headers as $header) {
-            $name = strtolower(trim(explode(':', $header, 2)[0]));
-            $headers[$name] = $header;
-        }
+        // Replace only headers which are being propagated. Repeated unrelated headers may be
+        // intentional, and CURLOPT_HTTPHEADER permits them.
+        $headersToPropagate = [];
         foreach ($this->headersToPropagate as $header) {
             $name = strtolower(trim(explode(':', $header, 2)[0]));
-            $headers[$name] = $header;
+            $headersToPropagate[$name] = $header;
         }
+
+        $headers = array_values(array_filter(
+            $this->headers,
+            static fn (string $header): bool => !isset($headersToPropagate[strtolower(trim(explode(':', $header, 2)[0]))]),
+        ));
         $this->headersToPropagate = [];
 
-        return array_values($headers);
+        return [...$headers, ...array_values($headersToPropagate)];
     }
 
     public function getCapturedResponseHeaders(): array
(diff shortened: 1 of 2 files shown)


PuvaanRaaj:
cc @open-telemetry/php-approvers @brettmc for review 🙏

intuibase (approved):

intuibase (approved):

intuibase (approved):

PR #656, opened 2026-07-11
feat(laravel): inject trace context into outbound `HTTP` client requests

zigzagdev:
### Motivation

`ClientRequestWatcher` records spans for outbound `Illuminate\Http\Client` requests, but never injected　`traceparent`/`tracestate` headers into them. 
Since the watcher only listens on the `RequestSending` event, which carries an immutable PSR-7 request that never reaches the request Guzzle actually sends, distributed　tracing was completely broken for any downstream service called via Laravel's HTTP client — there was no way to link the receiving service's trace back to the calling trace.

### What I have done

- Added `RequestPropagationSetter` (`src/Propagators/RequestPropagationSetter.php`), a `PropagationSetterInterface`implementation for PSR-7 `RequestInterface`, following the same pattern as the existing `ResponsePropagationSetter` and the other `PropagationSetterInterface` implementations across this monorepo.                                                                                   
 - Added a hook on `Illuminate\Http\Client\PendingRequest::runBeforeSendingCallbacks()` (`src/Hooks/Illuminate/Http/Client/PendingRequest.php`), the point where Laravel finalizes the PSR-7 request just                                                  
    before handing it to Guzzle. Injects the active span's trace context into the request headers there, since that's                                                  
    the last point the actual outgoing request can still be mutated.                                                                                                   
- Registered the new hook in `LaravelInstrumentation.php`.                                                                                                           
- Added `test_it_injects_trace_context_into_outbound_requests` to `ClientTest.php`, asserting the injected                                                           
    `traceparent` header matches an active parent span's trace/span id.

### Test Results

```bash                                                                                                                                                          
vendor/bin/phpunit tests/Integration/Http/ClientTest.php
PHPUnit 9.6.34 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.34
Configuration: /usr/src/myapp/src/Instrumentation/Laravel/phpunit.xml.dist

...                                                                 3 / 3 (100%)

Time: 00:01.066, Memory: 24.00 MB
OK (3 tests, 17 assertions)
```

### DIFF
--- a/src/Instrumentation/Laravel/src/Hooks/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Instrumentation/Laravel/src/Hooks/Illuminate/Http/Client/PendingRequest.php
@@ -50,6 +50,7 @@ class PendingRequest implements LaravelHook
                     return $params;
                 }
 
+                /** @phan-suppress-next-line PhanAccessMethodInternal */
                 TraceContextPropagator::getInstance()->inject($request, RequestPropagationSetter::instance(), Context::getCurrent());
                 $params[0] = $request;
 
--- a/src/Instrumentation/Laravel/src/Hooks/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Instrumentation/Laravel/src/Hooks/Illuminate/Http/Client/PendingRequest.php
@@ -9,6 +9,7 @@ use OpenTelemetry\API\Trace\Propagation\TraceContextPropagator;
 use OpenTelemetry\Context\Context;
 use OpenTelemetry\Contrib\Instrumentation\Laravel\Hooks\LaravelHook;
 use OpenTelemetry\Contrib\Instrumentation\Laravel\Hooks\LaravelHookTrait;
+use OpenTelemetry\Contrib\Instrumentation\Laravel\LaravelInstrumentation;
 use OpenTelemetry\Contrib\Instrumentation\Laravel\Propagators\RequestPropagationSetter;
 use function OpenTelemetry\Instrumentation\hook;
 use Psr\Http\Message\RequestInterface;
@@ -28,6 +29,10 @@ class PendingRequest implements LaravelHook
      * is actually sent (the `RequestSending` event ClientRequestWatcher listens on can't do this
      * itself, since it only receives an immutable copy that never reaches the real request).
      *
+     * Gated behind an opt-in flag: the target of an outbound HTTP client request may be a
+     * third-party service outside the application's control, which should not receive internal
+     * trace (and, in future, baggage) data unless the developer explicitly asks for it.
+     *
      * @psalm-suppress PossiblyUnusedReturnValue
      */
     protected function hookRunBeforeSendingCallbacks(): bool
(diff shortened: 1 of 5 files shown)


ChrisLightfootWild (commented):
We should opt into this behaviour rather than it be the default. There is no mechanism to prevent leaking to external endpoints (we do not know what a developer may be sending in baggage etc)

zigzagdev:
@ChrisLightfootWild 

I [fixed](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/656/commits/1038204573014a3e41b9d1f744b345ba2794a0a3) it.

src/Instrumentation/Laravel/src/Hooks/Illuminate/Http/Client/PendingRequest.php
```diff
@@ -0,0 +1,60 @@
...
+                if (!LaravelInstrumentation::shouldPropagateHttpClientTraceContext()) {
+                    return $params;
+                }
+
+                $request = $params[0];
+                if (!$request instanceof RequestInterface) {
+                    return $params;
+                }
+
+                TraceContextPropagator::getInstance()->inject($request, RequestPropagationSetter::instance(), Context::getCurrent());
```
PuvaanRaaj, line 54: This call currently fails the Laravel quality jobs with PhanAccessMethodInternal: RequestPropagationSetter is marked @internal and is being called from the Hooks subnamespace. Please adjust the annotation/scope or use the project’s established suppression pattern so the required quality checks pass.

PuvaanRaaj (commented):
Reviewed the updated opt-in propagation implementation and its Laravel CI results. The opt-in addresses the earlier leakage concern, but the current head has a directly introduced Phan failure that blocks the quality matrix.

src/Instrumentation/Laravel/src/Hooks/Illuminate/Http/Client/PendingRequest.php
zigzagdev: @PuvaanRaaj 

Suppressed via the same @phan-suppress-next-line PhanAccessMethodInternal pattern used in Kernel.php, rather than removing @internal from

▎ RequestPropagationSetter (it's intentionally internal). phan passes locally now.

[fix commit](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/656/commits/fb70805c4b13d77811c3b7863ad72fd06d748363)

bobstrecansky:
@zigzagdev - are you still working on this PR?

zigzagdev:
@bobstrecansky 

Thanks for checking in! I'm not working in this PR — it's ready for review/merge.

Summary of what's been addressed:

1. Initial implementation: added trace context injection for outbound HTTP client requests
2. @ChrisLightfootWild's feedback: made the propagation opt-in (default off) to prevent leaking to external endpoints
3. @PuvaanRaaj's feedback: fixed the Phan `PhanAccessMethodInternal` warning using the project's established suppression pattern (same as Kernel.php)

All tests pass locally. Happy to make further changes if needed, but as far as I can tell, all raised concerns have been addressed.

intuibase:
@zigzagdev 
Could you please document new config option (`OTEL_PHP_INSTRUMENTATION_LARAVEL_HTTP_CLIENT_PROPAGATION_ENABLED`) in README.md?

zigzagdev:
@intuibase 

Thank you for reviewing!
I've added documentation for OTEL_PHP_INSTRUMENTATION_LARAVEL_HTTP_CLIENT_PROPAGATION_ENABLED to README.md in [this commit](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/656/commits/09270268ca23bf4fe05077719881ede2a73ce94c), with a clear explanation of when and why to enable it.

Please let me know if the documentation looks correct.

andrii-androshchuk (approved):

PR #660, opened 2026-07-17
chore: swtch to jaeger v2

thompson-tomo:
The jaegar tracing image has been switched as existing image is eol jaegertracing/jaeger#6321


ChrisLightfootWild (approved):
Thanks 👍

PR #661, opened 2026-07-18
feat(symfony): create root span for console command execution

### LINKED ISSUE #1774: [opentelemetry-auto-symfony] missing console default root span creation
I am in the process of setting up open telemetry for a Symfony 6 project. Everything is working nicely so far. However I noticed that only for the HTTP kernel a default root span is being created. I have created a listener for the ConsoleCommandEvent, ConsoleErrorEvent, ... which handles the default root span myself for now. 

Was this left out on purpose? I'm wondering if I should not be doing this.

For reference, the packages I'm currently depending on:
```
        "open-telemetry/exporter-otlp": "^1.3",
        "open-telemetry/opentelemetry-auto-doctrine": "^0.3.1",
        "open-telemetry/opentelemetry-auto-symfony": "^1.1",
        "open-telemetry/opentelemetry-logger-monolog": "^1.1",
        "open-telemetry/sdk": "^1.9",
        "open-telemetry/transport-grpc": "^1.1",
```

Thanks in advance.

PuvaanRaaj:
## The gap

The Symfony auto-instrumentation creates a default root span for HTTP requests (via the `HttpKernel::handle` hook), but nothing equivalent existed for **console commands**. When a Symfony CLI command runs, no root span was created, so console executions produced no trace unless the user wired up their own `ConsoleEvents` listener by hand.

This was reported in open-telemetry/opentelemetry-php#1774.

## The fix

Adds a new `ConsoleInstrumentation` that hooks `Symfony\Component\Console\Command\Command::run()` and creates an `INTERNAL` root span for the command execution, mirroring the existing HTTP kernel instrumentation idiom (same `CachedInstrumentation` / `hook()` pre+post / `Context::storage()` scope handling):

- **Span name** taken from the command name (falls back to `command` when unnamed).
- **`code.*` attributes** (`code.function.name`, `code.file.path`, `code.line.number`) set like the other hooks.
- **Exit code** recorded as `symfony.console.exit_code`.
- **Error status**: sets `StatusCode::STATUS_ERROR` and records the exception on a thrown `Throwable`, and sets `STATUS_ERROR` for a non-success (non-zero) return value.

Wired into `_register.php` alongside the existing registrations, and `symfony/console` added to `require-dev` with the same `^5.4||^6.0` constraint used for `symfony/messenger` and `symfony/http-client`.

Fixes open-telemetry/opentelemetry-php#1774

## Testing

Added `ConsoleInstrumentationTest` (integration) covering: a successful command produces one INTERNAL root span with the expected name and exit-code attribute; a non-zero exit code sets ERROR status; a throwing command records the exception and sets ERROR status with the exception message as description.

Verified locally with the `opentelemetry` extension loaded: full package suite passes (`vendor/bin/phpunit` — 25 tests, 136 assertions) and `php-cs-fixer --dry-run` reports no style violations.

### DIFF
--- a/src/Instrumentation/Symfony/_register.php
+++ b/src/Instrumentation/Symfony/_register.php
@@ -2,6 +2,7 @@
 
 declare(strict_types=1);
 
+use OpenTelemetry\Contrib\Instrumentation\Symfony\ConsoleInstrumentation;
 use OpenTelemetry\Contrib\Instrumentation\Symfony\HttpClientInstrumentation;
 use OpenTelemetry\Contrib\Instrumentation\Symfony\MessengerInstrumentation;
 use OpenTelemetry\Contrib\Instrumentation\Symfony\SymfonyInstrumentation;
@@ -20,3 +21,4 @@ if (extension_loaded('opentelemetry') === false) {
 SymfonyInstrumentation::register();
 MessengerInstrumentation::register();
 HttpClientInstrumentation::register();
+ConsoleInstrumentation::register();
--- a/src/Instrumentation/Symfony/composer.json
+++ b/src/Instrumentation/Symfony/composer.json
@@ -31,7 +31,8 @@
     "phpunit/phpunit": "^9.5",
     "vimeo/psalm": "6.4.0",
     "symfony/http-client": "^5.4||^6.0",
-    "symfony/messenger": "^5.4||^6.0"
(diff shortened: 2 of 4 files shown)


bobstrecansky:
@PuvaanRaaj - are you still working on this PR?

PR #662, opened 2026-07-18
fix(symfony): clean up sub-request scope and record exceptions on spans

### LINKED ISSUE #1844: [opentelemetry-php-contrib] [Symfony auto-instrumentation] Exception events not recorded on spans when Symfony handles errors via sub-request
## Summary

When an exception is thrown in a Symfony controller, the exception event is not visible in the exported trace, even though the span shows `Status: Error`. This is because the `terminate` hook retrieves the wrong span (the error_controller sub-request span) instead of the original main request span.

## Environment

- PHP 8.4
- Symfony 8
- open-telemetry/opentelemetry-auto-symfony ^1.1.1

## Reproduction

Test endpoint: `GET /v1/menu/test/error/division`

```php

#[Route('/test/error/division', name: 'test_division_error', methods: ['GET'])]
public function testDivisionError(): JsonResponse
{
    $a = 10;
    $b = 0;
    $result = intdiv($a, $b); // DivisionByZeroError
    
    return $this->json(['result' => $result]);
}
```

## Problem Description

When an exception occurs in a Symfony application:

1. The exception is thrown in the controller
2. `handleThrowable` is called and records the exception on the **main request span**
3. Symfony creates a **sub-request** to `error_controller` to render the error page
4. The sub-request creates a **new span** that gets pushed onto the context storage stack
5. `terminate` is called and retrieves the span from `Context::storage()->scope()`
6. **Bug**: This returns the sub-request span, not the main request span
7. The main request span (with the exception event) is never properly ended/exported

The span names confirm this in debug:
- `handleThrowable`: span name is "GET"
- `terminate`: span name is "GET error_controller"

### BEFORE: Trace output (missing exception event)

```
Span #0
    Trace ID       : d500694f4e175ce558a22ef2692c9dd4
    Parent ID      : 763b808e5ea7cb22
    ID             : 11f365f73b482185
    Name           : GET /v1/menu/test/error/division
    Kind           : Internal
    Start time     : 2025-12-30 14:21:16.917804532 +0000 UTC
    End time       : 2025-12-30 14:21:18.603553128 +0000 UTC
    Status code    : Error
    Status message :                          <-- Empty!
    DroppedAttributesCount: 0
    DroppedEventsCount: 0                     <-- No events!
    DroppedLinksCount: 0
Attributes:
     -> http.response.status_code: Int(500)
     -> http.route: Str(/v1/menu/test/error/division)
     ...
                                              <-- No Events section!
```

**Note**: The span shows `Status: Error` but has no exception event and no status message.

### Issue Sub-request spans never ended

The `handle` POST hook only processes when there's an exception:

```php
post: static function (...) {
    // ...
    if (null === $scope || null === $exception) {
        return;  // <-- Sub-request completes successfully, returns early
    }
    // Span never detached or ended!
}
```

The error_controller sub-request completes **successfully** (it renders the error page), so its span is created but never ended.

### Issue terminate() gets wrong span

```php
// In terminate hook:
$scope = Context::storage()->scope();  // Returns top of stack = sub-request span
$span = Span::fromContext($scope->context());  // Wrong span!
```

The context storage is a stack. The sub-request span was pushed on top but never popped, so `terminate` gets the wrong span.
### Fix: Handle sub-requests in `handle` POST hook

Always end sub-request spans in the `handle` POST hook since `terminate()` only handles the main request:

```php
post: static function (...) {
    $scope = Context::storage()->scope();
    if (null === $scope) {
        return;
    }

    $type = $params[1] ?? HttpKernelInterface::MAIN_REQUEST;
    $isSubRequest = ($type === HttpKernelInterface::SUB_REQUEST);

    // MAIN_REQUEST without exception: let terminate() handle everything
    if (!$isSubRequest && null === $exception) {
        return;
    }

    $span = Span::fromContext($scope->context());
    $scope->detach();

    // Record exception if present
    if (null !== $exception) {
        $span->recordException($exception, [
            TraceAttributes::EXCEPTION_ESCAPED => true,
        ]);
        $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
    }

    // SUB_REQUEST: end span here since terminate() only handles MAIN_REQUEST
    if ($isSubRequest) {
        if (null !== $response) {
            $span->setAttribute(TraceAttributes::HTTP_RESPONSE_STATUS_CODE, $response->getStatusCode());
        }
        $span->end();
    }
}
```

### AFTER: Trace output (with fix applied)

Now we get **two spans** with proper exception recording:

**Span #0 - Sub-request (error_controller)**
```
Span #0
    Trace ID       : f392848bd679c0a6e2f38ca888cc2fbf
    Parent ID      : f7c3f8b3f8ecd552
    ID             : ...
    Name           : GET error_controller
    Kind           : Internal
    Status code    : Unset
    -> http.response.status_code: Int(500)
```

**Span #1 - Main request (with exception event!)**
```
Span #1
    Trace ID       : f392848bd679c0a6e2f38ca888cc2fbf
    Parent ID      : 
    ID             : f7c3f8b3f8ecd552
    Name           : GET /v1/menu/test/error/division
    Kind           : Server
    Start time     : 2025-12-30 14:23:42.004143064 +0000 UTC
    End time       : 2025-12-30 14:23:44.928306556 +0000 UTC
    Status code    : Error
    Status message : 
Attributes:
     -> http.response.status_code: Int(500)
     -> http.route: Str(/v1/menu/test/error/division)
     -> network.protocol.version: Str(1.1)
     -> http.response.body.size: Int(92985)

Events:
SpanEvent #0
     -> Name: exception
     -> Timestamp: 2025-12-30 14:23:43.154858485 +0000 UTC
     -> Attributes::
          -> exception.type: Str(DivisionByZeroError)
          -> exception.message: Str(Division by zero)
          -> exception.stacktrace: Str(DivisionByZeroError: Division by zero
            at intdiv(MenuController.php:429)
            at App.Controller.MenuController.testDivisionError(MenuController.php:429)
            at Symfony.Component.HttpKernel.HttpKernel.handleRaw(HttpKernel.php:183)
            at Symfony.Component.HttpKernel.HttpKernel.handle(HttpKernel.php:76)
            at Symfony.Component.HttpKernel.Kernel.handle(Kernel.php:191)
            ...)
```

### LINKED ISSUE #1905: [contrib-auto-symfony] SUB_REQUEST scope not cleaned up in handle post-hook, causing root HTTP spans to never be exported
> Filing here as `opentelemetry-php-contrib` has issues disabled. This concerns `open-telemetry/opentelemetry-auto-symfony` `1.1.1`.

## Summary

When Symfony handles a request that results in a non-200 response, `HttpKernel::handle()` is called **twice** — once for the main request and once internally as a `SUB_REQUEST` to render the error page via `ExceptionController`. The `handle` post-hook returns early for both calls (because `$exception` is `null` for successful sub-requests), leaving two scopes on the OTEL context stack. When `terminate()` fires it pops only the topmost scope (the sub-request), ending that span. The main request scope (the root `SERVER` span) is never popped or ended — and therefore never exported.

## Environment

- `open-telemetry/opentelemetry-auto-symfony`: `1.1.1`
- Symfony: `6.4`
- PHP: `8.2`
- OTEL exporter: OTLP/gRPC → Tempo

## Steps to Reproduce

Any request that triggers Symfony's error handling internally (e.g. a 404, 403, or any response that goes through `ExceptionController`) will reproduce this:

1. Make a request that returns a non-200 response (e.g. hit a route that returns a 404)
2. Check your trace backend (Jaeger, Tempo, etc.)

**Result:** Child spans (e.g. Doctrine queries) appear with `rootServiceName: <root span not yet received>`. The root HTTP span is never received.

## Root Cause

After both `handle()` calls complete, the OTEL context stack looks like this:

```
[top]    scope 1 → sub-request span  (KIND_INTERNAL, child of scope 2)
[bottom] scope 2 → main request span (KIND_SERVER, root, no parent)
```

The `handle` post-hook:

```php
post: static function (...) {
    $scope = Context::storage()->scope();
    if (null === $scope || null === $exception) {
        return;  // ← returns early for all successful requests, including sub-requests
    }
    // ...
}
```

Both the SUB_REQUEST post-hook (no exception) and the MAIN_REQUEST post-hook (no exception when error page renders successfully) return early. Both scopes remain on the stack.

`terminate()` then pops **one** scope (scope 1, the sub-request), ends it, and returns. Scope 2 (the main HTTP span) stays on the stack forever — never ended, never exported.

Note: this is distinct from the exception case fixed in PR #317. The bug occurs on **normal** request handling whenever Symfony renders an error response internally.

## Expected Behavior

The root HTTP span (`KIND_SERVER`) should be ended and exported for every request, including those that produce error responses.

## Proposed Fix

In the `handle` post-hook, detect `SUB_REQUEST` and end those spans immediately, since sub-requests never receive a `terminate()` call:

```php
post: static function (
    HttpKernel $kernel,
    array $params,
    ?Response $response,
    ?\Throwable $exception
): void {
    $type = $params[1] ?? HttpKernelInterface::MAIN_REQUEST;
    $scope = Context::storage()->scope();

    if (null === $scope) {
        return;
    }

    // Main request with no exception: terminate() handles span ending
    if ($type === HttpKernelInterface::MAIN_REQUEST && null === $exception) {
        return;
    }

    $span = Span::fromContext($scope->context());
    $scope->detach();

    if (null !== $exception) {
        $span->recordException($exception, [TraceAttributes::EXCEPTION_ESCAPED => true]);
        if (null !== $response && $response->getStatusCode() >= Response::HTTP_INTERNAL_SERVER_ERROR) {
            $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
        }
    }

    // Sub-requests don't get a terminate() call — end the span here
    if ($type === HttpKernelInterface::SUB_REQUEST) {
        $span->end();
    }
},
```

This ensures the sub-request scope is cleaned up immediately, so when `terminate()` fires only the main request scope remains on the stack and is properly ended and exported.

We are currently working around this with `cweagans/composer-patches` but would love to see this fixed upstream. Happy to open a PR if this direction looks good.

PuvaanRaaj:
## Summary

Fixes two related bugs in the Symfony auto-instrumentation `HttpKernel::handle()` hook lifecycle, both rooted in the same sub-request handling code path.

## Root cause

Symfony calls `HttpKernel::handle()` twice when it renders an error response internally: once for the `MAIN_REQUEST` and again as a nested `SUB_REQUEST` dispatched to the error controller. The old `handle` post-hook returned early whenever there was no exception:

```php
$scope = Context::storage()->scope();
if (null === $scope || null === $exception) {
    return;
}
```

The error-controller sub-request renders successfully (no exception), so its span was never ended and its scope was never detached. This left the sub-request scope leaked on top of the context stack.

- **#1905** — With the sub-request scope leaked, `terminate()` popped only that topmost (sub-request) scope. The main-request root `SERVER` span was never popped, ended, or exported, so child spans showed `rootServiceName: <root span not yet received>`.
- **#1844** — `handleThrowable` records the exception on the currently-active span (the main span) before the error sub-request is dispatched. Because the leaked sub-request scope displaced the main scope, `terminate()` ended the wrong span and the main span — the one carrying the exception event and `Status: Error` — was never exported.

## The fix

In the `handle` post-hook:

- A `MAIN_REQUEST` that returned normally still returns early and is ended by the `terminate` hook (unchanged behaviour).
- Every other path — sub-requests, and any request where an exception propagated out of `handle()` — now detaches its scope and ends its span in the post-hook, since neither is ever passed to `terminate()`. This ensures every attached scope is detached exactly once and no scope leaks onto the context stack.
- Exceptions continue to be recorded (`recordException` + `STATUS_ERROR`) on the active span when present.

With the sub-request scope cleaned up immediately, `terminate()` sees the main-request scope again and ends/exports the root span with its exception event intact.

## Testing

- Added `test_http_kernel_sub_request_scope_does_not_leak_main_span`: a main request followed by an internal sub-request leaves no leaked scope; the sub-request span exports immediately and the main root `SERVER` span is ended and exported by `terminate()`.
- Added `test_http_kernel_records_exception_on_main_span_during_sub_request_error_handling`: an exception recorded on the active span during error handling ends up on the exported main span.
- Updated the existing sub-request tests to stop calling `terminate()` after a `SUB_REQUEST` handle (Symfony never does this for sub-requests; the span is now ended by the post-hook).
- Full suite passes locally: `vendor/bin/phpunit` → 24 tests, 132 assertions, OK. `php-cs-fixer --dry-run` clean.

Fixes open-telemetry/opentelemetry-php#1905
Fixes open-telemetry/opentelemetry-php#1844

### DIFF
--- a/src/Instrumentation/Symfony/composer.json
+++ b/src/Instrumentation/Symfony/composer.json
@@ -27,6 +27,7 @@
     "phpstan/phpstan": "^1.1",
     "phpstan/phpstan-phpunit": "^1.0",
     "psalm/plugin-phpunit": "^0.19.2",
+    "open-telemetry/opentelemetry-propagation-traceresponse": "^0.3",
     "open-telemetry/sdk": "^1.8",
     "phpunit/phpunit": "^9.5",
     "vimeo/psalm": "6.4.0",
--- a/src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
@@ -94,18 +94,41 @@ final class SymfonyInstrumentation
                 ?\Throwable $exception
             ): void {
                 $scope = Context::storage()->scope();
-                if (null === $scope || null === $exception) {
+                if (null === $scope) {
+                    return;
+                }
+
+                $type = $params[1] ?? HttpKernelInterface::MAIN_REQUEST;
+                $isSubRequest = ($type === HttpKernelInterface::SUB_REQUEST);
+
+                // A MAIN_REQUEST that didn't throw is ended by the `terminate` hook.
+                if (!$isSubRequest && null === $exception) {
                     return;
                 }
 
                 $span = Span::fromContext($scope->context());
                 $scope->detach();
-                $span->recordException($exception);
-                $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
+
+                if (null !== $exception) {
+                    $span->recordException($exception);
+                    $span->setStatus(StatusCode::STATUS_ERROR, $exception->getMessage());
+                }
+
+                if ($isSubRequest && null !== $response) {
+                    $span->setAttribute(TraceAttributes::HTTP_RESPONSE_STATUS_CODE, $response->getStatusCode());
+                }
+
+                // We only reach this point for sub-requests or when an exception propagated
+                // out of handle(); neither is ever passed to `terminate()`, so the span must
+                // be ended here. Leaving a sub-request span open leaks its scope onto the
+                // context stack and prevents the main-request root span from being ended and
+                // exported (#1905). Only a MAIN_REQUEST that returned normally is left for
+                // `terminate()`, and that path already returned above.
                 $span->end();
             }
         );
 
+        /** @psalm-suppress UnusedFunctionCall */
         hook(
             HttpKernel::class,
             'terminate',
--- a/src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
+++ b/src/Instrumentation/Symfony/src/SymfonyInstrumentation.php
@@ -116,6 +116,9 @@ final class SymfonyInstrumentation
 
                 if ($isSubRequest && null !== $response) {
                     $span->setAttribute(TraceAttributes::HTTP_RESPONSE_STATUS_CODE, $response->getStatusCode());
+                    if ($response->isServerError()) {
+                        $span->setStatus(StatusCode::STATUS_ERROR);
+                    }
                 }
 
                 // We only reach this point for sub-requests or when an exception propagated
--- a/src/Instrumentation/Symfony/tests/Integration/SymfonyInstrumentationTest.php
+++ b/src/Instrumentation/Symfony/tests/Integration/SymfonyInstrumentationTest.php
@@ -6,6 +6,7 @@ namespace OpenTelemetry\Tests\Instrumentation\Symfony\tests\Integration;
 
 use OpenTelemetry\API\Trace\SpanKind;
 use OpenTelemetry\API\Trace\StatusCode;
+use OpenTelemetry\Contrib\Propagation\TraceResponse\TraceResponsePropagator;
 use OpenTelemetry\SemConv\TraceAttributes;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
@@ -143,8 +144,9 @@ class SymfonyInstrumentationTest extends AbstractTest
         $request = new Request();
         $request->attributes->set('_controller', 'ErrorController');
 
-        $response = $kernel->handle($request, HttpKernelInterface::SUB_REQUEST);
-        $kernel->terminate($request, $response);
+        // Sub-requests are never passed to terminate() by Symfony; their span is ended
+        // by the handle post-hook, so it is exported as soon as handle() returns.
+        $kernel->handle($request, HttpKernelInterface::SUB_REQUEST);
 
         $this->assertCount(1, $this->storage);
 
@@ -160,8 +162,7 @@ class SymfonyInstrumentationTest extends AbstractTest
         // String controller
         $request = new Request();
         $request->attributes->set('_controller', 'SomeController::index');
-        $response = $kernel->handle($request, \Symfony\Component\HttpKernel\HttpKernelInterface::SUB_REQUEST);
-        $kernel->terminate($request, $response);
+        $kernel->handle($request, \Symfony\Component\HttpKernel\HttpKernelInterface::SUB_REQUEST);
         $this->assertSame('GET SomeController::index', $this->storage[0]->getName());
         $this->storage->exchangeArray([]);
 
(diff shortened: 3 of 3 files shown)


bobstrecansky:
@PuvaanRaaj - are you still working on this PR?

PR #694, opened 2026-08-28
fix(doctrine): add re-entrancy guard to prevent duplicate spans in connection wrapper chains

BTC-Tim:
## Problem

`DoctrineInstrumentation` registers hooks on `Doctrine\DBAL\Driver::connect`, `Doctrine\DBAL\Driver\Connection::{query,exec,prepare,beginTransaction,commit,rollBack}` and `Doctrine\DBAL\Driver\Statement::execute`. Those are *interface* methods, so the hooks fire at **every layer of a connection wrapper chain** — a custom `Middleware`, or any `Driver` that delegates to an inner `Driver`.

The result is N spans for one logical database operation, where N is the number of wrapper layers. Traces become noisy and the nesting is misleading.

## Fix

A per-method static re-entrancy depth counter. The pre-hook increments it and skips span creation when `depth > 1`; the post-hook decrements it and only calls `end()` at depth 0. Only the outermost invocation creates and ends a span, so behaviour is unchanged for anyone without a wrapper chain.

The pre-hook closures also get an explicit `: void` return type. They already returned an implicit `null`; the guard's early exit made that inconsistent, which Phan flags as `PhanPluginInconsistentReturnFunction`.

## Test

`test_wrapper_chain_does_not_duplicate_spans` puts a two-layer middleware (driver + connection) in front of the in-memory SQLite driver and asserts exactly one span for connect, exec, prepare, execute, beginTransaction, commit and rollBack.

Without the guard the test fails immediately — `connect` alone emits **3** spans instead of 1.

Local run against `main`, PHP 8.2.30, DBAL 4.4.4:

| Check | Result |
|---|---|
| PHPUnit | OK (10 tests, 56 assertions) |
| PHP-CS-Fixer | 0 of 8 files need fixing |
| Phan | no issues |
| Psalm | no errors found |
| PHPStan | no errors |

## Notes

Fixes open-telemetry/opentelemetry-php#1931, which has a second reporter confirming the same behaviour.

This supersedes opentelemetry-php/contrib-auto-doctrine#3, which I opened in March against the read-only subtree split before realising PRs belong here. I will close that one once this lands.

### DIFF
--- a/src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
+++ b/src/Instrumentation/Doctrine/src/DoctrineInstrumentation.php
@@ -20,6 +20,14 @@ class DoctrineInstrumentation
 {
     public const NAME = 'doctrine';
 
+    /**
+     * Re-entrancy guard: Doctrine's connection wrapper chain fires hooks at
+     * every layer. Only the outermost call creates and ends a span.
+     *
+     * @var array<string, int>
+     */
+    private static array $depth = [];
+
     public static function register(): void
     {
         $instrumentation = new CachedInstrumentation('io.opentelemetry.contrib.php.doctrine');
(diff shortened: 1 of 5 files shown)


BTC-Tim:
Closing this — I'm not able to complete the EasyCLA step, so it can't be merged as-is.

Leaving the work here for anyone who wants to pick it up: the re-entrancy guard, the Phan fix the guard requires, and an integration test that reproduces the bug (a two-layer middleware makes `connect` alone emit 3 spans instead of 1). The branch stays available at `BTC-Tim:fix/doctrine-duplicate-spans-wrapper-chain`.

open-telemetry/opentelemetry-php#1931 remains open and has a second reporter.

PR #706, opened 2026-09-03
Skip shared contrib-ci workflows from being updated by renovate.

ChrisLightfootWild:
Prevent renovate from touching the shared workflow references, instead always pointing to main in this repository.

https://docs.renovatebot.com/configuration-options/#ignorepaths

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -29,5 +29,8 @@
   ],
   "labels": [
     "dependencies"
+  ],
+  "ignorePaths": [
+    "src/**/.github/workflows/contrib-ci.yml"
   ]
 }
--- a/src/Instrumentation/Curl/.github/workflows/contrib-ci.yml
+++ b/src/Instrumentation/Curl/.github/workflows/contrib-ci.yml
@@ -9,6 +9,6 @@ permissions:
 
 jobs:
   CI:
-    uses: open-telemetry/opentelemetry-php-contrib/.github/workflows/instrumentation-curl.yml@4f68ecb9760605760d991360df23c6cdeee7dab4 # main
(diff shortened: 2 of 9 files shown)


ChrisLightfootWild:
Hey @trask - tagging for your input here please, as you added the Zizmor workflow. Apologies if this is not the correct means of raising this :+1:

This repository may feels slightly unusual, in that we perform a gitsplit to allow composer (the PHP package manager) to play nicely with our packages (as they must live within their own repositories). We are essentially wanting a shared workflow to be called ([example](https://github.com/opentelemetry-php/contrib-auto-laravel/blob/main/.github/workflows/contrib-ci.yml) - to be expanded into push/schedule in future).

The problem we face here is that renovate is continuously updating the reference to these workflows. Obviously as we accept those changes, `main` points to a new commit and then renovate decides to update them all again (rinse/repeat).

The proposed options to overcome this are:
 - we create a `contrib-ci-workflow` tag in this repo that we can update when the workflow requires it
 - we move the workflow into the `opentelemetry-php` org and manage it from there
   - not really ideal, as this is not part of the managed `open-telemetry` org
 - renovate ignores the shared workflows from it's maintenance. Zizmor ignore's the unpinned reference
   - that's this PR :see_no_evil: 

Is there any preferred way forward with this? Or another alternative that may not have been considered.

Many thanks in advance!

trask:
hey @ChrisLightfootWild! I think this is probably justifiable from a security perspective, given it is pointing to our own repo rather than an external repo. The workflows are also currently read-only and don't receive secrets.

One alternative would be to create a long-lived branch like `contrib-ci-stable` specifically for these workflows and the callers could remain pinned to that branch's full SHA. Renovate would then update the pins, while regular merges into `main` wouldn't start the loop again.

PR #707, opened 2026-09-04
fix(monolog): replace deprecated ConfigurationResolver with SDK Configuration::getEnum()

intuibase:
Replace deprecated ConfigurationResolver with the EnvComponentLoader/EnvResolver SPI pattern from the API package, avoiding an explicit SDK runtime dependency:

  - Add HandlerConfiguration (InstrumentationConfiguration POPO) to hold mode
  - Add ConfigEnv/HandlerEnvLoader (EnvComponentLoader) — reads OTEL_PHP_MONOLOG_ATTRIB_MODE from env via injected EnvResolver
  - Register HandlerEnvLoader in composer.json extra.spi
  - Handler constructor accepts HandlerConfiguration (defaults to psr3 mode)
  - Remove open-telemetry/sdk from runtime require; move to require-dev
  - Fix $mode from shared static to per-instance property
  - Upgrade vimeo/psalm to ^6.4 for PHP 8.4+ compatibility
  - Add HandlerEnvLoaderTest and update HandlerTest to inject config directly

### DIFF
--- a/src/Logs/Monolog/composer.json
+++ b/src/Logs/Monolog/composer.json
@@ -20,22 +20,29 @@
   },
   "require-dev": {
     "friendsofphp/php-cs-fixer": "^3",
+    "open-telemetry/sdk": ">=1.0",
     "google/protobuf": ">=3.5.0",
     "guzzlehttp/guzzle": "^7.4",
     "nyholm/psr7": "^1.6",
     "open-telemetry/exporter-otlp": ">=1.0",
-    "open-telemetry/sdk": ">=1.0",
     "phan/phan": "^6.0",
     "phpbench/phpbench": "^1.2",
     "phpstan/phpstan": "^1.1",
     "phpstan/phpstan-phpunit": "^1.0",
     "psalm/plugin-phpunit": "^0.19.2",
     "phpunit/phpunit": "^9.5",
-    "vimeo/psalm": "6.4.0"
+    "vimeo/psalm": "^6.4"
   },
   "config": {
     "allow-plugins": {
       "php-http/discovery": false
+    }
+  },
+  "extra": {
+    "spi": {
+      "OpenTelemetry\\API\\Configuration\\ConfigEnv\\EnvComponentLoader": [
+        "OpenTelemetry\\Contrib\\Logs\\Monolog\\ConfigEnv\\HandlerEnvLoader"
+      ]
     }
   }
 }
--- a/src/Logs/Monolog/composer.json
+++ b/src/Logs/Monolog/composer.json
@@ -20,11 +20,11 @@
   },
   "require-dev": {
     "friendsofphp/php-cs-fixer": "^3",
-    "open-telemetry/sdk": ">=1.0",
     "google/protobuf": ">=3.5.0",
     "guzzlehttp/guzzle": "^7.4",
     "nyholm/psr7": "^1.6",
     "open-telemetry/exporter-otlp": ">=1.0",
+    "open-telemetry/sdk": ">=1.0",
     "phan/phan": "^6.0",
     "phpbench/phpbench": "^1.2",
     "phpstan/phpstan": "^1.1",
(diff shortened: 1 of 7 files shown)


ChrisLightfootWild:
I think we should keep away from an explicit dependency on the SDK here.

I queried configuration recently in https://github.com/open-telemetry/opentelemetry-php/issues/1865

`EnvComponentLoader` was the steer, with the suggested changes in https://github.com/ChrisLightfootWild/opentelemetry-php-contrib/pull/1.

intuibase:
> I think we should keep away from an explicit dependency on the SDK here.
> 
> I queried configuration recently in [open-telemetry/opentelemetry-php#1865](https://github.com/open-telemetry/opentelemetry-php/issues/1865)
> 
> `EnvComponentLoader` was the steer, with the suggested changes in [ChrisLightfootWild#1](https://github.com/ChrisLightfootWild/opentelemetry-php-contrib/pull/1).

Thank you @ChrisLightfootWild , I've updated it, PTAL

src/Logs/Monolog/composer.json
```diff
@@ -20,22 +20,29 @@
...
     "allow-plugins": {
       "php-http/discovery": false
     }
+  },
+  "extra": {
+    "spi": {
+      "OpenTelemetry\\API\\Configuration\\ConfigEnv\\EnvComponentLoader": [
+        "OpenTelemetry\\Contrib\\Logs\\Monolog\\ConfigEnv\\HandlerEnvLoader"
+      ]
+    }
```
ChrisLightfootWild, line 46: I'm not sure this does anything for this Handler on the face of it - I need to look into this a little more. Maybe @Nevay can better advise whether there is an expected way of doing this without SPI?

e.g: when *only* the API would be available, is there a single entrypoint for loading config (given `ConfigurationResolver` is deprecated)?

src/Logs/Monolog/composer.json
```diff
@@ -20,22 +20,29 @@
   },
   "require-dev": {
     "friendsofphp/php-cs-fixer": "^3",
+    "open-telemetry/sdk": ">=1.0",
```
ChrisLightfootWild: nit: Can this go back in alphabetical order? Sorry! 🤓

src/Logs/Monolog/src/ConfigEnv/HandlerEnvLoader.php
```diff
@@ -0,0 +1,30 @@
...
+use OpenTelemetry\API\Configuration\Context;
+use OpenTelemetry\Contrib\Logs\Monolog\Handler;
+use OpenTelemetry\Contrib\Logs\Monolog\HandlerConfiguration;
+
+/**
+ * @implements EnvComponentLoader<HandlerConfiguration>
+ */
+final class HandlerEnvLoader implements EnvComponentLoader
+{
+    public function load(EnvResolver $env, EnvComponentLoaderRegistry $registry, Context $context): HandlerConfiguration
```
Nevay:
```suggestion
    public function load(EnvResolver $env, EnvComponentLoaderRegistry $registry, Context $context): InstrumentationConfiguration
```

src/Logs/Monolog/composer.json
Nevay: > I'm not sure this does anything for this Handler on the face of it

Currently, it does nothing. The return type of `HandlerEnvLoader::load()` needs to be changed to `InstrumentationConfiguration` to make the `HandlerConfiguration` available in `ConfigProperties`.

> e.g: when only the API would be available, is there a single entrypoint for loading config

We should add a global [`ConfigProvider`](https://opentelemetry.io/docs/specs/otel/configuration/api/#configprovider). Manually registered instrumentations can then use `Globals::configProvider()->getInstrumentationConfig()` to access the resolved configuration values. Currently, `ConfigProperties` are only accessible in `Instrumentation::register()`.

Tbh, I'm not sure why `$mode` was added as env-based configuration instead of being a constructor parameter. I would expect either both `$level` and `$mode` to be configurable via instrumentation configuration (env-based / file-based) or both to be configured only via code.

PR #709, opened 2026-09-04
feat(laravel): don't create umbrella spans for long-running Artisan commands

AdilAzhariOmsan:
## Summary

`Hooks/Illuminate/Console/Command.php` opens a `Command <name>` span for every Artisan command and only ends it when `Illuminate\Console\Command::execute()` returns; `Hooks/Illuminate/Contracts/Console/Kernel.php` does the same with an `Artisan handler` span (gated behind `OTEL_PHP_TRACE_CLI_ENABLED`).

For a worker/daemon command — `queue:work`, `queue:listen`, `horizon`, `octane:start`, `schedule:work`, a custom consumer loop — that method never returns until the process is stopped. While the span is open it is the ambient `Context::getCurrent()`, so:

- every span created inside the worker loop (Redis `BLPOP`/`eval` from `migrateExpiredJobs`, DB reconnects, cache polls) nests under it;
- every job whose payload carries no valid `traceparent` gets parented onto it, because `Hooks/Illuminate/Queue/Worker.php::setParentContext()` falls through to `$spanBuilder->setParent(Context::getCurrent())`.

The result is one unbounded trace per worker process — collectors reject it (`TRACE_TOO_LARGE`) and memory grows until the worker restarts.

This is a known issue with an agreed-on design that never got finished:

- #1440 names `Console/Command.php` exactly. @ChrisLightfootWild: *"I actually started work on this but never pushed it up."* Auto-closed stale, no fix landed.
- #1411 (same root cause via `CacheWatcher`) — @LauJosefsen proposed a built-in denylist + `#[LongRunningCommand]` attribute for custom commands + an env var to extend it; a maintainer was receptive. Also auto-closed stale.
- A downstream fork (`cego/php-contrib-auto-laravel`) exists solely to patch this.
- `keepsuit/laravel-opentelemetry` shipped the equivalent fix in 1.13.1 ([keepsuit#57](https://github.com/keepsuit/laravel-opentelemetry/issues/57)); the reporter confirmed it "completely resolved the issue".

This PR implements the design from #1411/#1440.

## What changed

- New `Contracts\Console\LongRunningCommand` attribute — mark a custom worker/consumer command with `#[LongRunningCommand]`.
- New `Hooks\Illuminate\Console\LongRunningCommands` helper — decides whether a command is long-running: built-in list (`queue:work`, `queue:listen`, `horizon`, `horizon:work`, `horizon:supervisor`, `schedule:work`, `reverb:start`, `reverb:restart`, `pail`, `octane:start`, `octane:frankenphp`, `octane:swoole`) ∪ `OTEL_PHP_INSTRUMENTATION_LARAVEL_LONG_RUNNING_COMMANDS` (comma separated, `*` wildcards, merged with the built-in list) ∪ the attribute.
- `Console\Command` and `Contracts\Console\Kernel` hooks: skip span creation for a matching command. The `post` hook re-checks the same condition rather than tracking state, so it never detaches a scope it didn't attach (safe with nested command calls, e.g. `optimize:clear`'s sub-commands, which are untouched by this change).
- README: documents the env var and the attribute.
- Tests: unit coverage for the matching logic (built-in names, non-matches, env-var + wildcards, attribute), integration coverage that a long-running command (by attribute, and by env var) produces no span while an ordinary command still does.

## Known trade-off

Removing the umbrella span means spans created between jobs inside the worker loop (e.g. Redis polling) become their own single-span traces instead of nesting under one giant trace — the same behavior change `keepsuit/laravel-opentelemetry`'s fix produced, and what resolved the equivalent report there. A follow-up could suppress instrumentation during the idle poll loop itself; out of scope here.

## Testing

```
vendor/bin/phpunit --testsuite unit         # 35/35 passing locally
vendor/bin/phpstan analyse                  # no new findings (existing findings are pre-existing, unrelated to this change)
vendor/bin/php-cs-fixer fix --dry-run --diff  # no style findings in the changed files
```

The integration suite (`tests/Integration/Console/CommandTest.php`) needs `ext-opentelemetry`, not available in my local environment — CI covers it; the added assertions follow the exact pattern of the existing `test_command_tracing` / `test_failing_command_sets_status_error` tests in that file.

Fixes #1440
Fixes #1411

### DIFF
--- a/src/Instrumentation/Laravel/README.md
+++ b/src/Instrumentation/Laravel/README.md
@@ -22,3 +22,37 @@ The extension can be disabled via [runtime configuration](https://opentelemetry.
 ```shell
 OTEL_PHP_DISABLED_INSTRUMENTATIONS=laravel
 ```
+
+### Long-running commands
+
+No `Command` span is created for worker / daemon commands, because such a command only returns when
+the process is stopped: the span would never end and would become the ambient parent for every job,
+request and query handled during the process lifetime, collapsing everything into a single unbounded
+trace. The work done inside each iteration (e.g. a queued job) is still traced on its own.
+
+Laravel's own long-running commands are recognised out of the box: `queue:work`, `queue:listen`,
+`horizon`, `horizon:work`, `horizon:supervisor`, `schedule:work`, `reverb:start`, `reverb:restart`,
+`pail`, `octane:start`, `octane:frankenphp`, `octane:swoole`.
+
+Add your own with a comma-separated list of names (`*` wildcards allowed), merged with the built-in
+list:
+
+```shell
+OTEL_PHP_INSTRUMENTATION_LARAVEL_LONG_RUNNING_COMMANDS="app:kafka-consume,app:relay-*"
+```
+
+Or mark the command class with an attribute:
+
+```php
+use OpenTelemetry\Contrib\Instrumentation\Laravel\Contracts\Console\LongRunningCommand;
+
+#[LongRunningCommand]
+class KafkaConsumeCommand extends Command
+{
+    // ...
+}
+```
+
+The `Artisan handler` span (only created when `OTEL_PHP_TRACE_CLI_ENABLED=true`) is skipped for the
+same commands. That span sees the command name only, so a custom command marked purely with
+`#[LongRunningCommand]` must also be listed in the environment variable to be skipped there.
--- a/src/Instrumentation/Laravel/src/Contracts/Console/LongRunningCommand.php
+++ b/src/Instrumentation/Laravel/src/Contracts/Console/LongRunningCommand.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Contrib\Instrumentation\Laravel\Contracts\Console;
+
+use Attribute;
+
+/**
+ * Marks an Artisan command as long-running (a worker / daemon / consumer loop).
+ *
+ * The console instrumentation does not open a `Command` span for a command
+ * carrying this attribute, because such a command only returns when the process
+ * stops. A never-ending span becomes the ambient parent for every job, request
+ * or query handled during the process lifetime, producing a single unbounded
+ * trace.
+ *
+ * Laravel's own long-running commands (`queue:work`, `horizon`, `octane:start`,
+ * ...) are recognised by name out of the box; use this attribute for custom
+ * commands, or the
+ * `OTEL_PHP_INSTRUMENTATION_LARAVEL_LONG_RUNNING_COMMANDS` environment variable
+ * for commands you cannot annotate.
+ */
+#[Attribute(Attribute::TARGET_CLASS)]
+final class LongRunningCommand
+{
+}
--- a/src/Instrumentation/Laravel/src/Hooks/Illuminate/Console/Command.php
+++ b/src/Instrumentation/Laravel/src/Hooks/Illuminate/Console/Command.php
@@ -32,6 +32,13 @@ class Command implements LaravelHook
             IlluminateCommand::class,
             'execute',
             pre: function (IlluminateCommand $command, array $params, string $class, string $function, ?string $filename, ?int $lineno) {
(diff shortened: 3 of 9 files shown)


bobstrecansky:
Will review once you've signed the CLA @AdilAzhariOmsan !

andrii-androshchuk:
This is a very common problem, but I don't think we should simply remove the "umbrella spans," as we'll lose the context of what preceded the child spans inside it. This would leave us with a set of seemingly "random" child spans, without any understanding of where they came from or what process triggered them.

AdilAzhariOmsan:
@andrii-androshchuk Yeah, that's the real cost here and I probably should've led with it instead of burying it in "Known trade-off." You're right that once the umbrella span is gone, a poll/job span created inside queue:work's loop shows up as its own little single-span trace with nothing tying it back to "this came from that worker."

Where I landed: the umbrella span isn't really giving us that context safely today either — it never closes for the life of the process, so it either gets rejected by the collector once it's too big, or it silently adopts jobs that have no traceparent of their own, which mixes unrelated invocations into one trace. So it's not "rich context" vs "no context," it's more "misleading context" vs "no context" but I get that "no context" isn't great either.

Is there a middle ground you'd be more comfortable with? A couple of things come to mind but I don't have a strong opinion yet: resource attributes (host/pid/command) on the process so a poll span can still be traced back to its worker without a live parent span, or maybe a bounded span per loop iteration instead of per-process so it actually closes. Curious what you'd want to see here happy to shape this PR (or a follow-up) around whatever solves the "lost context" problem without bringing back the unbounded-span issue.

ChrisLightfootWild:
I'm investigating building this span suppression outside of units of work to prevent the long-running traces.

andrii-androshchuk:
Here is how I think we could solve it, @AdilAzhariOmsan / @ChrisLightfootWild:

The umbrella span shouldn't be scoped to the process lifetime. It should be scoped to the unit of work, same as Laravel Octane already does for HTTP: one persistent worker process, but each request gets an isolated flow, and container/request state reset at the request boundary, not at process start/stop. Apply the same logic here: open/close spans based on inner-process triggers (request/job boundary), instead of the outer process.

AdilAzhariOmsan:
@ChrisLightfootWild glad you're picking this up. Suppressing spans outside units of work is the right direction, and it matches @andrii-androshchuk's idea of scoping spans to the job/request like Octane does. I closed #709 in favour of it. Edge cases I hit:

**Where the boundaries are**
- **`queue:work`:** `Worker::process` already opens a span per job, but `Command queue:work` (and `Artisan handler` with `OTEL_PHP_TRACE_CLI_ENABLED`) stays ambient, so jobs without a `traceparent` get parented to it. Jobs should start from the root context.
- **Bounded commands:** `queue:work --once`/`--max-jobs`/`--stop-when-empty` (which `queue:listen` spawns per job) end on their own. A fixed name list like #709's would wrongly suppress them.
- **No job boundary:** `queue:listen`, `horizon`, `horizon:supervisor`, `pulse:*`, `reverb:start` and custom consumer loops have no job to scope to. They need suppression or an opt-in unit-of-work marker/API. `schedule:work` → per scheduled event; Octane → requests are covered, ticks/task workers aren't.

**Leaks**
- **Empty polls:** `getNextJob` still exports empty `receive` spans (`$span->end()` despite the "discard" comment). They'd become a trace every few seconds per worker.
- **Payload propagation:** `createPayloadArray` injects the current context, so jobs dispatched outside a unit of work (custom consumers, `Looping` listeners) carry the long-running trace into other workers.
- **Watchers:** Cache, Redis, Query and client-request spans need the same suppression, otherwise they become separate root traces.
- **Scopes:** starting from `Context::getRoot()` guards against leaked scopes. Suppressed pre-hooks in `Command`/`Kernel` must skip their post-hooks too.

**Should keep working:** `SyncQueue`, `Artisan::call()` and chained/batched jobs dispatched from a job, and `TracingIsolated`/`Linked`/`Parent`. Also worth handling: flushing in-flight spans on timeout/`SIGTERM`, context safety for Swoole coroutines and ReactPHP loops, and possibly an opt-out for anyone relying on the process span.

Happy to help with tests or any part of this.

PR #710, opened 2026-09-05
fix(laravel): set messaging.operation.type attribute on queue spans

zigzagdev:
### Motivation
Queue/Worker instrumentation embeds the messaging operation (send/create/process/receive) into the span *name* only, e.g. `spanBuilder(vsprintf('%s %s', [MESSAGING_OPERATION_TYPE_VALUE_SEND, ...]))`. 
It was never set as the `messaging.operation.type` span attribute, which the OTel messaging semantic conventions require. 

Therefore, These spans without a queryable/filterable field that every compliant messaging span is expected to carry.

### What I have done
- `Queue::hookBulk` → set `MESSAGING_OPERATION_TYPE` to `send`
- `Queue::hookLater` → set `MESSAGING_OPERATION_TYPE` to `create`
- `Queue::hookPushRaw` → set `MESSAGING_OPERATION_TYPE` to `create`
- `Worker::hookWorkerProcess` → set `MESSAGING_OPERATION_TYPE` to `process`
- `Worker::hookWorkerGetNextJob` → set `MESSAGING_OPERATION_TYPE` to `receive`
- `SyncQueue::hookPush` → set `MESSAGING_OPERATION_TYPE` to `process` (also replaced the hardcoded `'process'` string in the span name with the `MessagingIncubatingAttributes` constant, for consistency with the other hooks touched here)

### Test Plans
- Extended `QueueTest.php` (`test_it_handles_pushing_to_a_queue`, `test_it_can_push_a_message_with_a_delay`, `test_it_can_publish_in_bulk`, `test_it_can_create_with_redis`) and `WorkerTest.php` (`test_job_has_parent_trace_by_default`) with assertions on the new attribute
- Added `test_it_records_operation_type_on_receive_span` to cover the `receive` case

### DIFF
--- a/src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Queue/Queue.php
+++ b/src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Queue/Queue.php
@@ -42,6 +42,7 @@ class Queue implements LaravelHook
                     CodeAttributes::CODE_FUNCTION_NAME => sprintf('%s::%s', $class, $function),
                     CodeAttributes::CODE_FILE_PATH => $filename,
                     CodeAttributes::CODE_LINE_NUMBER => $lineno,
+                    MessagingIncubatingAttributes::MESSAGING_OPERATION_TYPE => MessagingIncubatingAttributes::MESSAGING_OPERATION_TYPE_VALUE_SEND,
                     MessagingIncubatingAttributes::MESSAGING_BATCH_MESSAGE_COUNT => count($params[0] ?? []),
                 ], $this->contextualMessageSystemAttributes($queue, []));
 
@@ -85,6 +86,7 @@ class Queue implements LaravelHook
                     CodeAttributes::CODE_FUNCTION_NAME => sprintf('%s::%s', $class, $function),
                     CodeAttributes::CODE_FILE_PATH => $filename,
                     CodeAttributes::CODE_LINE_NUMBER => $lineno,
+                    MessagingIncubatingAttributes::MESSAGING_OPERATION_TYPE => MessagingIncubatingAttributes::MESSAGING_OPERATION_TYPE_VALUE_CREATE,
                     'messaging.message.delivery_timestamp' => $estimateDeliveryTimestamp,
                 ];
 
@@ -119,6 +121,7 @@ class Queue implements LaravelHook
             pre: function (QueueContract $queue, array $params, string $_class, string $_function, ?string $_filename, ?int $_lineno) {
                 /** @phan-suppress-next-line PhanParamTooFewUnpack */
                 $attributes = $this->buildMessageAttributes($queue, ...$params);
+                $attributes[MessagingIncubatingAttributes::MESSAGING_OPERATION_TYPE] = MessagingIncubatingAttributes::MESSAGING_OPERATION_TYPE_VALUE_CREATE;
 
                 $parent = Context::getCurrent();
                 /** @psalm-suppress ArgumentTypeCoercion */
--- a/src/Instrumentation/Laravel/src/Hooks/Illuminate/Queue/SyncQueue.php
+++ b/src/Instrumentation/Laravel/src/Hooks/Illuminate/Queue/SyncQueue.php
@@ -12,6 +12,7 @@ use OpenTelemetry\Contrib\Instrumentation\Laravel\Hooks\LaravelHookTrait;
 use OpenTelemetry\Contrib\Instrumentation\Laravel\Hooks\PostHookTrait;
 use function OpenTelemetry\Instrumentation\hook;
 use OpenTelemetry\SemConv\Attributes\CodeAttributes;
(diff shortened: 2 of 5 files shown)


ChrisLightfootWild:
>  It was never set as the `messaging.operation.type` span attribute, which the OTel messaging semantic conventions require.
> 
> Therefore, These spans without a queryable/filterable field that every compliant messaging span is expected to carry.

I am not sure that those existed when this was first implemented (I think around `1.25.0` of the semantic conventions).

src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Queue/Queue.php
```diff
@@ -42,6 +42,7 @@ protected function hookBulk(): bool
                     CodeAttributes::CODE_FUNCTION_NAME => sprintf('%s::%s', $class, $function),
                     CodeAttributes::CODE_FILE_PATH => $filename,
                     CodeAttributes::CODE_LINE_NUMBER => $lineno,
+                    MessagingIncubatingAttributes::MESSAGING_OPERATION_TYPE => MessagingIncubatingAttributes::MESSAGING_OPERATION_TYPE_VALUE_SEND,
```
ChrisLightfootWild, line 45: > every compliant messaging span is expected to carry.

It's at the incubating stage, so optional to implement?

zigzagdev:
@ChrisLightfootWild 

> I am not sure that those existed when this was first implemented (I think around 1.25.0 of the semantic conventions).

The messaging semconv is still in Development status.
It hasn't reached a stable release. 
The same applies to messaging.operation.type: it only settled into "conditionally required, if applicable" in `v1.27.0`. 
Before that, there was debate about whether it should be required at all. 
So it seems plausible the spec wasn't settled yet when this code was originally written.

The four operations this PR handles (send, create, receive, process) all match the values the spec expects, so the "required if applicable" condition is satisfied in every case here.

src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Queue/Queue.php
zigzagdev: @ChrisLightfootWild 

> It's at the incubating stage, so optional to implement?

Yes, it's Incubating. 
But this file already sets other attributes from the same `MessagingIncubatingAttributes` class unconditionally, like `MESSAGING_SYSTEM` and `MESSAGING_DESTINATION_NAME`. 
This PR just adds one more attribute from that same family.

Also, this repo has no mechanism anywhere to make Incubating attributes optional. 
Making only this one attribute optional would leave the file inconsistent with itself.

andrii-androshchuk:
This is actually a good question: how should we treat the incubating attributes? Do we want to integrate them from day one, or work only with stable attributes?

I am in favor of using stable attributes only, to avoid unnecessary backward-compatibility issues. Alternatively, if we decide to integrate new (incubating) attributes, we should make sure they can be easily removed if they are eventually rejected.

zigzagdev:
@andrii-androshchuk 

Thanks for  your comment !  🙌

This raises a repo-wide policy question (stable-only vs. incubating attributes) in the review of a single PR that just adds one more attribute from a family (`MessagingIncubatingAttributes`) that this same file already sets unconditionally (`MESSAGING_SYSTEM`, `MESSAGING_DESTINATION_NAME`). 
The decision to adopt incubating semconv attributes here was effectively already made when this file was first written — singling out `messaging.operation.type` for special treatment now would appear to ignore file consistency.

The following 4 reasons suggest we should not use stable-only in this context:

1. Usage is repo-wide, not Laravel-specific: 29 files across `src/` reference an `Incubating\Attributes` class, spanning most instrumentation packages — Guzzle, Symfony, Curl, Psr15, Psr18, Slim, Yii, CodeIgniter, ExtAmqp, ExtRdKafka, MongoDB, and more.
2. The pattern is consistently unconditional: e.g. in Guzzle and Symfony, `HttpIncubatingAttributes::HTTP_REQUEST_BODY_SIZE`/`HTTP_RESPONSE_BODY_SIZE` are set via plain `setAttribute()` chained right alongside stable attributes like `HttpAttributes`/`UrlAttributes`, with no feature flag or conditional gating anywhere.
There's no existing mechanism in this repo to treat incubating attributes as opt-in or easily removable.

3.  There's no documented policy: `CONTRIBUTING.md` doesn't mention "incubating" or "sem-conv"/"semconv" at all.

4. It's already baked into the dependency contract: these packages pin `open-telemetry/sem-conv` with caret ranges (e.g. `^1.38`), where the `Incubating` namespace ships as a normal part of the package, not an opt-in extra.

### Conclusion

Given these considerations, adopting a stable-only policy now would potentially create breaking changes affecting ~29 files across the repo, not something this PR should be gated on. 
I'd like to keep the current implementation as-is. If we want to settle the stable-vs-incubating question as a repo-wide policy, I think that deserves its own issue/discussion rather than blocking this PR.

PR #711, opened 2026-09-05
fix(laravel): queue spans per request signature in `ClientRequestWatcher`

zigzagdev:
### Motivation

`ClientRequestWatcher` tracks in-flight HTTP client spans in a single array slot keyed by `sha1(method|url|body)`. 
When two requests share the same signature and are in flight at the same time — e.g. identical calls sent through `Http::pool()`, or a retried request — the second `recordRequest()` overwrites the first entry in `$this->spans`. 
The overwritten span is never ended (it leaks), and whichever `ResponseReceived`/`ConnectionFailed` event fires next gets attributed to the wrong span (or to none, once both slots have been consumed).

### What I have done
- Changed `$spans` from `array<string, SpanInterface>` to `array<string, list<SpanInterface>>`: one queue per request signature instead of one slot.
- `recordRequest()` now appends the new span to the queue for its signature.
- `recordResponse()`/`recordConnectionFailed()` now pop the oldest span for that signature via a new `shiftSpan()` helper (FIFO), instead of reading/deleting a single keyed entry.
- Concurrent requests with an identical signature are now matched back to their own response/failure in send order, instead of colliding.

### Test Plans
- Added `test_it_matches_spans_to_responses_for_concurrent_identical_requests` to `ClientTest.php`: sends two identical `GET` requests through `Http::pool()`, with `Http::sequence()` returning a distinct status (200, then 500) for each, and asserts each recorded span carries the correct status and error state for its own response.
- Verified this test reproduces the bug: reverting the `ClientRequestWatcher.php` change while keeping the test causes it to fail (only 1 span recorded instead of 2).

### DIFF
--- a/src/Instrumentation/Laravel/src/Watchers/ClientRequestWatcher.php
+++ b/src/Instrumentation/Laravel/src/Watchers/ClientRequestWatcher.php
@@ -23,7 +23,7 @@ use Symfony\Component\HttpFoundation\Response as HttpResponse;
 class ClientRequestWatcher extends Watcher
 {
     /**
-     * @var array<string, SpanInterface>
+     * @var array<string, list<SpanInterface>>
      */
     protected array $spans = [];
 
@@ -67,31 +67,25 @@ class ClientRequestWatcher extends Watcher
                 ServerAttributes::SERVER_PORT => $parsedUrl['port'] ?? '',
             ])
             ->startSpan();
-        $this->spans[$this->createRequestComparisonHash($request->request)] = $span;
+        $this->spans[$this->createRequestComparisonHash($request->request)][] = $span;
     }
 
     /** @psalm-suppress PossiblyUnusedMethod */
     public function recordConnectionFailed(ConnectionFailed $request): void
     {
-        $requestHash = $this->createRequestComparisonHash($request->request);
-
-        $span = $this->spans[$requestHash] ?? null;
+        $span = $this->shiftSpan($this->createRequestComparisonHash($request->request));
         if (null === $span) {
             return;
         }
 
         $span->setStatus(StatusCode::STATUS_ERROR, 'Connection failed');
         $span->end();
-
-        unset($this->spans[$requestHash]);
     }
 
     /** @psalm-suppress PossiblyUnusedMethod */
     public function recordResponse(ResponseReceived $request): void
     {
-        $requestHash = $this->createRequestComparisonHash($request->request);
-
-        $span = $this->spans[$requestHash] ?? null;
+        $span = $this->shiftSpan($this->createRequestComparisonHash($request->request));
         if (null === $span) {
             return;
         }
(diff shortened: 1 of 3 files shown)


ChrisLightfootWild (commented):
There's an assumption that the response return in order, whereas that may not be the case.

I can see there is already a problem currently, but can we think of any other mechanism to help differentiate these requests? 🤔

zigzagdev:
@ChrisLightfootWild 

> I can see there is already a problem currently, but can we think of any other mechanism to help differentiate these requests? 🤔

Yes, there's a better mechanism: correlating requests by object identity instead of content.
So I searched the correct way and [fixed](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/711/commits/0eb3d09819225661b4b25f2739f4cc34c029675f) and [added new test](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/711#:~:text=test(laravel)%3A%20verify%20ClientRequestWatcher%20matches%20spans%20by%20identity) at here.

Firstly, the `FIFO-per-signature` queue still assumed responses resolve in the same order requests were sent. 
That assumption doesn't hold for concurrent requests (e.g. `Http::pool()`). 
If a response resolves out of order, the queue pops the wrong span and misattributes its status and attributes to the wrong request.

I considered a few ways to fix this:

1. Keep content-hash + FIFO, try to fix the ordering somehow
Not really fixable. The content hash (`method|url|body`) carries no information about which in-flight request a given response belongs to. Any queue or ordering-based matching is fundamentally guessing based on send order, and send order isn't guaranteed to match completion order.

2. `SplObjectStorage` keyed by the underlying PSR-7 request object
`RequestSending`, `ResponseReceived`, and `ConnectionFailed` all reference the same PSR-7 request instance for a given attempt ([PendingRequest.php#L238-242](https://github.com/laravel/framework/blob/v10.50.3/src/Illuminate/Http/Client/PendingRequest.php#L238-L242), [#L1397-1416](https://github.com/laravel/framework/blob/v10.50.3/src/Illuminate/Http/Client/PendingRequest.php#L1397)). 
Keying by object identity removes the ordering assumption entirely: two in-flight requests with identical content always get distinct keys, regardless of resolution order. 
`SplObjectStorage` holds a strong reference to that key object, making it the textbook-safe way to do object-identity keying in PHP.

Drawback: it pins the entire PSR-7 request (headers, body, etc.) in memory for as long as the entry sits unresolved.
If a `ResponseReceived/ConnectionFailed` event is ever missed for some reason (a pre-existing risk this watcher already has), that becomes a heavier leak than today's: a whole request object per leaked entry instead of a short hash string.

3. `spl_object_id()` of the underlying PSR-7 request as a plain `int` key (what I went with)
Same identity-based correctness as #2, but it stores only a scalar int per in-flight span instead of holding the request object itself.
Drawback: `spl_object_id()` ids are only unique among currently-living objects. 
Once an object is destroyed, PHP may reuse its id for something else ([php.net](https://www.php.net/manual/en/function.spl-object-id.php): "When an object is destroyed, its id may be reused for other objects."). 
If the PSR-7 request were freed before its matching response/failure event fires, a later, unrelated object could theoretically reuse the same id and get its span misattributed.

### Why did i chose  3
In this specific code path, the request object is kept alive by Guzzle's handler/promise chain for the entire in-flight duration. 
We're not relying on anything outside that, so the reuse window doesn't open before the correlating event fires in practice. 
Given that, I preferred the lighter-weight option over paying #2's memory cost for a guarantee we already get for free here. 
I'm happy to switch to `SplObjectStorage` if you'd rather have the belt-and-suspenders version regardless.

And I also added `ClientRequestWatcherTest::test_it_matches_the_correct_span_when_responses_resolve_out_of_order`, which resolves the later-sent of two identical requests first and verifies (via reflection on the internal span map) that only the resolved request's entry is removed. 
It fails against the old FIFO implementation and passes against this one.

PR #727, opened 2026-09-09
fix(laravel): Prefer toMql() over toSql() in getModels pre-hook for MongoDB queries

### LINKED ISSUE #3578: open-telemetry/opentelemetry-auto-laravel report waring
⚬	Laravel-mongodb Version: 5.7.1
⚬	PHP Version: 8.4.11
⚬	Database Driver & Version: mongodb/laravel-mongodb  5.7.1 & ext-mongodb 2.1.1
Description:
When using OpenTelemetry instrumentation with Laravel and MongoDB, executing Eloquent queries triggers a PHP warning/exception in the OpenTelemetry pre-hook for Illuminate\Database\Eloquent\Builder::getModels(), stating that the method is not supported by MongoDB.
Steps to reproduce
	1.	Install and configure OpenTelemetry PHP auto-instrumentation/extension for Laravel.
	2.	Use mongodb/laravel-mongodb as the database connection.
	3.	Perform a standard Eloquent query (e.g., User::all() or User::where('status', 'active')->get()).
	4.	Check the PHP error logs.
Expected behaviour
Eloquent queries should execute normally and be monitored by OpenTelemetry without throwing unhandled exceptions or warnings regarding method compatibility.
Actual behaviour
A PHP warning is generated in the logs:
PHP Warning: Illuminate\Database\Eloquent\Builder::getModels(): OpenTelemetry: pre hook threw exception, class=Illuminate\Database\Eloquent\Builder function=getModels message=This method is not supported by MongoDB. Try "toMql()" instead. in Unknown on line 0
Logs:
[07-Sep-2026 15:08:42 Asia/Shanghai] PHP Warning: Illuminate\Database\Eloquent\Builder::getModels(): OpenTelemetry: pre hook threw exception, class=Illuminate\Database\Eloquent\Builder function=getModels message=This method is not supported by MongoDB. Try "toMql()" instead. in Unknown on line 0

GromNaN:
The `getModels` pre-hook calls `$builder->getQuery()->toSql()` to set `DB_QUERY_TEXT`, but laravel-mongodb overrides `toSql()` to throw and redirect to `toMql()`, which makes the hook warn on every Eloquent query. Detect `toMql()` on the query builder and prefer `json_encode($query->toMql())` so the hook records a faithful MongoDB query text instead of throwing.

Fix https://github.com/mongodb/laravel-mongodb/issues/3578

### DIFF
--- a/src/Instrumentation/Laravel/src/Hooks/Illuminate/Database/Eloquent/Model.php
+++ b/src/Instrumentation/Laravel/src/Hooks/Illuminate/Database/Eloquent/Model.php
@@ -160,6 +160,12 @@ class Model implements LaravelHook
             'getModels',
             pre: function ($builder, array $params, string $class, string $function, ?string $filename, ?int $lineno) {
                 $model = $builder->getModel();
+                $query = $builder->getQuery();
+                // laravel-mongodb overrides toSql() to throw; use toMql() instead.
+                $queryText = method_exists($query, 'toMql')
+                    ? json_encode($query->toMql())
+                    : $query->toSql();
+
                 $builder = $this->instrumentation
                     ->tracer()
                     ->spanBuilder($model::class . '::get')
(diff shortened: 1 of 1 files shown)


ChrisLightfootWild (commented):
#668 also got raised recently for this.

I think removing the `db.statement` entirely would be my preference here, as it is actually logged by QueryWatcher already as a nested span.

GromNaN:
Closing in favor of #735, which removes the `db.query.text` attribute from the `getModels` span entirely, per the review suggestion.

One nuance worth noting: laravel-mongodb does not dispatch `Illuminate\Database\Events\QueryExecuted` (it runs queries directly through `MongoDB\Collection`, bypassing `BaseConnection::run`), so `QueryWatcher` only produces the nested sql span for SQL drivers. For MongoDB this matches the pre-existing behavior where the attribute was never set because `toSql()` threw, and the warning is now eliminated.

PR #728, opened 2026-09-11
fix(laravel): stop mixing query string into `url.path`

zigzagdev:
### Motivation

`Kernel.php`'s `httpTarget()` appended the query string onto the request path and set the combined result as the `url.path` span attribute. 
OTel semantic conventions define `url.path` as the path component only ([registry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/url/)); the query belongs in the separate `url.query` attribute, which is Conditionally Required (only when a query string was actually received). 
Existing tests asserted the combined `path?query` value as correct, which locked the bug in as expected behaviour.

### What I have done

- Renamed `httpTarget()` to `httpPath()` and removed the query-string concatenation, so it returns the path only.
- `url.path` is now set from `httpPath()`.
- `url.query` is set from `getQueryString()`, guarded behind an explicit `if (null !== $query = ...)` check so the attribute is only added when a query string is actually present. (An earlier version of this fix called `setAttribute` unconditionally and relied on the SDK silently dropping null-valued attributes to keep `url.query` absent otherwise; that's an SDK implementation detail, not a guaranteed contract, so a PR about spec compliance shouldn't lean on it.)
- Updated the existing query-string tests, which previously asserted the buggy `path?query`-in-`url.path` behaviour, to assert the correct split values.
- Added a test covering the no-query-string case, asserting `url.query` is absent via `has()` (not `assertNull()`, which can't distinguish "attribute absent" from "attribute present with a null value").

### Test Plans

- `vendor/bin/phpunit` — 54 tests, all green
- `vendor/bin/phpstan analyse` — no errors
- `vendor/bin/php-cs-fixer fix --dry-run` — clean

### DIFF
--- a/src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Http/Kernel.php
+++ b/src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Http/Kernel.php
@@ -61,7 +61,7 @@ class Kernel implements LaravelHook
                 if ($request) {
                     /** @phan-suppress-next-line PhanAccessMethodInternal */
                     $parent = Globals::propagator()->extract($request, HeadersPropagator::instance());
-                    $span = $builder
+                    $builder = $builder
                         ->setParent($parent)
                         ->setAttribute(UrlAttributes::URL_FULL, $this->httpFullUrl($request))
                         ->setAttribute(HttpAttributes::HTTP_REQUEST_METHOD, $method)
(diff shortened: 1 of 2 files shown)


brettmc (approved):

src/Instrumentation/Laravel/src/Hooks/Illuminate/Contracts/Http/Kernel.php
```diff
@@ -61,21 +61,26 @@ protected function hookHandle(): bool
...
                         ->setAttribute(ServerAttributes::SERVER_ADDRESS, $this->httpHostName($request))
                         ->setAttribute(ServerAttributes::SERVER_PORT, $request->getPort())
                         ->setAttribute(ClientAttributes::CLIENT_PORT, $request->server('REMOTE_PORT'))
                         ->setAttribute(ClientAttributes::CLIENT_ADDRESS, $request->ip())
-                        ->setAttribute(UserAgentAttributes::USER_AGENT_ORIGINAL, $request->userAgent())
-                        ->startSpan();
+                        ->setAttribute(UserAgentAttributes::USER_AGENT_ORIGINAL, $request->userAgent());
+
+                    if (null !== $query = $request->getQueryString()) {
+                        $builder = $builder->setAttribute(UrlAttributes::URL_QUERY, $query);
```
ChrisLightfootWild, line 80: If the value is null, doesn't it get stripped anyway? We might be able to reduce the conditional check if so.

zigzagdev: @ChrisLightfootWild 

Thank you for reviewing !

However, I'd still like to keep this guard for two reasons:

1. Spec alignment — `url.query` is Conditionally Required: omitting it is the actual semconv requirement, not just a side effect of null-dropping.
2. Intent clarity — the `if` signals to readers that this omission is deliberate per spec, not incidental like the other unconditional `setAttribute` calls above it.

ChrisLightfootWild (approved):

ChrisLightfootWild:
Released in https://github.com/opentelemetry-php/contrib-auto-laravel/releases/tag/1.9.1

PR #729, opened 2026-09-11
Add OpenTelemetry auto-instrumentation for Phalcon

anatoliykim:
## Summary

Linked to issue - https://github.com/open-telemetry/opentelemetry-php/issues/2053

- New `open-telemetry/opentelemetry-auto-phalcon` package instrumenting apps built on the [Phalcon](https://phalcon.io/) PHP framework — no existing package covers this today.
- Hooks `Phalcon\Mvc\Application::handle()`, `Phalcon\Mvc\Micro::handle()`, `Phalcon\Cli\Console::handle()`, and `Phalcon\Dispatcher\AbstractDispatcher::dispatch()` (fallback root, only for apps dispatching directly without any of the other three) as root-span sources, sharing one depth-guarded counter so exactly one root span exists regardless of which style the app uses.
- `AbstractDispatcher::callActionMethod()` produces one child span per action actually invoked, including a forwarded or re-entrant action, deduplicated by comparing the hook's declaring-scope parameter against the target's actual polymorphic resolution for `parent::` internally, which would otherwise double-fire the hook for one real action).
- Root span named after the matched route (preferring its name, falling back to its pattern — skipping Phalcon's raw-regex default-route patterns, which aren't human-readable), else `{controller}/{action}` / `{task}/{action}` for CLI. 

## Requirements

- PHP >= 8.2 — Zend Observer API support for internal/extension-class method calls (what every method on a compiled Phalcon class is) was added in 8.2; on 8.1 these hooks register but never fire.
- Phalcon 5.x.

## CI

- Added `.github/workflows/instrumentation-phalcon.yml` (PHP 8.2–8.4), requesting the `phalcon5` extension via `shivammathur/setup-php`.
- Gave `php-contrib.yml` a new `extensions` input, defaulted to the existing hardcoded list, so this is additive — every other package's CI is unaffected.
- Added the `.gitsplit.yml` entry for the eventual `contrib-auto-phalcon` mirror.


.github/workflows/instrumentation-phalcon.yml
```diff
@@ -0,0 +1,39 @@
...
+          - 8.3
+          - 8.4
+
+    uses: ./.github/workflows/php-contrib.yml
+    with:
+      job-name: "PHP ${{ matrix.php-version }}"
+      codecov-flags: 'Instrumentation/Phalcon'
+      package-root: ${{ inputs.package-root || 'src/Instrumentation/Phalcon' }}
+      php-version: ${{ matrix.php-version }}
+      extensions: 'ast, amqp, grpc, opentelemetry, rdkafka, mysqli, pgsql, phalcon5'
```
ChrisLightfootWild, line 39: You probably don't need all of these now.

.github/workflows/php.yml
```diff
@@ -44,6 +44,7 @@ jobs:
           # 'Instrumentation/MySqli', # # Experiment: moved to instrumentation-mysqli.yml
           'Instrumentation/OpenAIPHP',
           'Instrumentation/PDO',
+          # 'Instrumentation/Phalcon', # Needs the phalcon5 extension: moved to instrumentation-phalcon.yml
```
ChrisLightfootWild: Guessing this is AI assisted? 🤖 😆

ChrisLightfootWild: No changes needed here.

src/Instrumentation/Phalcon/README.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+## Installing dependencies and executing tests
+
+From the Phalcon subdirectory:
+
+```bash
+$ composer install
+$ ./vendor/bin/phpunit tests
+```
```
ChrisLightfootWild: There are some makefile targets for this.

.github/workflows/instrumentation-phalcon.yml
anatoliykim: This workflow originally had these. To add a new `phalcon5` I, of course, have to preserve those.

.github/workflows/php.yml
anatoliykim: Yes. Got it, removed now.

src/Instrumentation/Phalcon/.phan/config.php
```diff
@@ -0,0 +1,371 @@
...
+return [
+
+    // Supported values: `'5.6'`, `'7.0'`, `'7.1'`, `'7.2'`, `'7.3'`, `'7.4'`, `null`.
+    // If this is set to `null`,
+    // then Phan assumes the PHP version which is closest to the minor version
+    // of the php executable used to execute Phan.
+    //
+    // Note that the **only** effect of choosing `'5.6'` is to infer that functions removed in php 7.0 exist.
+    // (See `backward_compatibility_checks` for additional options)
+    'target_php_version' => '8.0',
```
ChrisLightfootWild:
```suggestion
    'target_php_version' => '8.2',
```

.github/workflows/instrumentation-phalcon.yml
ChrisLightfootWild: Yes, but this package doesn't need `rdkafka`, `pgsql`, `amqp` etc., as I understand it.

.github/workflows/instrumentation-phalcon.yml
ChrisLightfootWild, line 1: There should also be a `.github/workflows/contrib-ci.yml` which references this.

.github/workflows/instrumentation-phalcon.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  CI:
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
+          - 8.2
+          - 8.3
+          - 8.4
```
brettmc, line 31: should add 8.5, since compose.json allows it

src/Instrumentation/Phalcon/src/PhalconInstrumentation.php
```diff
@@ -0,0 +1,738 @@
...
+                $throwable = $params[0] ?? null;
+                if ($throwable instanceof Throwable) {
+                    Span::getCurrent()->recordException($throwable);
+                }
+            },
+            post: null,
+        );
+    }
+
+    private static function beginRoot(): bool
```
brettmc: I think the function name is confusing, it doesn't actually begin (start) a root span? Same deal for `endRoot()`

.github/workflows/instrumentation-phalcon.yml
anatoliykim: phalcon5 isn't installable on PHP 8.5 via shivammathur/setup-php yet (its extension regex caps phalcon5 at 8.4). Adding 8.5 to this matrix would break CI on the extension-install step. Happy to add it once setup-php supports it

anatoliykim:
Both zizmor findings here are pre-existing across every sibling package (same ./... self-repository syntax, same stale SHA pin pattern in contrib-ci.yml) — not introduced by this PR. Fixing them just for Phalcon would be inconsistent with the rest of the repo, and the $/... self-repository syntax fix in particular may conflict with actionlint (which currently rejects it for reusable-workflow calls).

anatoliykim:
@ChrisLightfootWild:
  - Trimmed CI extensions list (removed rdkafka/pgsql/amqp/mysqli/grpc)
  - Removed the "AI assisted?" comment line from php.yml
  - README now documents the Makefile targets instead of bare composer/phpunit 
  - .phan/config.php target_php_version fixed to 8.2
  - Added the missing contrib-ci.yml

@brettmc brettmc:
  - beginRoot()/endRoot() naming → fixed, renamed to claimRoot()/releaseRoot()

PR #735, opened 2026-09-16
fix(laravel): remove db.query.text from getModels span to avoid toSql() throwing for MongoDB

### LINKED ISSUE #3578: open-telemetry/opentelemetry-auto-laravel report waring

GromNaN:
The `getModels` pre-hook called `$builder->getQuery()->toSql()` to set the `db.query.text` span attribute. Non-SQL drivers such as mongodb/laravel-mongodb override `toSql()` to throw a `BadMethodCallException` ("This method is not supported by MongoDB. Try `toMql()` instead."), because there is no SQL representation. The exception is thrown inside the pre-hook, so OpenTelemetry logs a warning on every Eloquent query (`get()` calls `getModels()` internally) and the attribute is never set.

Rather than special-casing `toMql()`, remove the `db.query.text` attribute from the `getModels` span entirely. For SQL drivers the query text is already recorded by `QueryWatcher` as a nested `sql` span (via `Illuminate\Database\Events\QueryExecuted`), making the parent-span attribute redundant. For MongoDB, laravel-mongodb does not dispatch `QueryExecuted` (it runs queries directly through `MongoDB\Collection`, bypassing `BaseConnection::run`), so no nested span is produced; this matches the pre-existing behavior where the attribute was never set because `toSql()` threw, and the warning is now eliminated.

Fixes mongodb/laravel-mongodb#3578. Supersedes #727.


ChrisLightfootWild:
Thanks for opening this @GromNaN 

I think the caveat makes sense here. It would be nice (down the road) if there was a dedicated MongoDB instrumentation, so if somebody wanted to see that then they could add it as an optional dependency to their application.

I would ultimately like to also explore removing the current QueryWatcher in favour of the PDO instrumentation, such that the query logging is not duplicated cross-package also.

GromNaN:
Note: we are working on a consistent OpenTelemetry implementation across MongoDB drivers in all languages. Tracked in https://jira.mongodb.org/browse/PHPLIB-1610.

ChrisLightfootWild (approved):

ChrisLightfootWild:
Tagged as https://github.com/opentelemetry-php/contrib-auto-laravel/releases/tag/1.9.1

Many thanks for the fix @GromNaN

